PR #4, opened 2016-06-15
Cleanup org.eclipse.xtext.tests

cdietrich:


org.eclipse.xtext.tests/build.gradle
```diff
@@ -1,17 +1,45 @@
+configurations {
+	mwe2 {
+		extendsFrom compile
+	}
```
cdietrich: do you expect to run mwe2 in other modules as well?

spoenemann: Possibly. And it would be good to have the same style of configuration in all repositories. Please have a look at [xtext-web/gradle/java-compiler-settings.gradle](https://github.com/eclipse/xtext-web/blob/master/gradle/java-compiler-settings.gradle) and [xtext-web/org.eclipse.xtext.web.example.statemachine/build.gradle](https://github.com/eclipse/xtext-web/blob/master/org.eclipse.xtext.web.example.statemachine/build.gradle) and check whether that configuration would work here.

spoenemann:
Please move the PR to a branch of the xtext-core repository so we can work together on it.

PR #25, opened 2016-07-01
[web] Fix for broken bracket matching in Orion

AlexTugarev:
In Orion's highlighting configuration all patterns that precede the last block pattern are treated as block patterns. Up to now the position of the `doc_block` pattern in the generated grammar causes all section patterns to become block patterns which breaks the bracket matching.

See also Orion bug #497097

Signed-off-by: Alex Tugarev <alex.tugarev@typefox.io>


spoenemann:
👍

PR #33, opened 2016-07-11
[generator] Use KeywordHelper to obtain all keywords

spoenemann:
When the `ignoreCase` option is active, the set of keywords returned by the KeywordHelper may be smaller than the whole set of keywords. This change fixes a case where a duplicate rule was generated due to the `ignoreCase` option.

Tested in `org.eclipse.xtext.lexer.IgnoreCaseLexerTestLanguage`.

### DIFF
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AbstractAntlrGrammarGenerator.xtend
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AbstractAntlrGrammarGenerator.xtend
@@ -139,7 +139,7 @@ abstract class AbstractAntlrGrammarGenerator {
 	protected def compileTokens(Grammar it, AntlrOptions options) '''
 		«IF options.isBacktrackLexer»
 			tokens {
-				«FOR kw : allKeywords.sort.sortBy[-length]»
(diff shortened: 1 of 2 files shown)


svenefftinge:
Did you regenerate the testlanguages?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AbstractAntlrGrammarGenerator.xtend
```diff
@@ -139,7 +139,7 @@ abstract class AbstractAntlrGrammarGenerator {
 	protected def compileTokens(Grammar it, AntlrOptions options) '''
 		�IF options.isBacktrackLexer�
 			tokens {
-				�FOR kw : allKeywords.sort.sortBy[-length]�
+				�FOR kw : keywordHelper.allKeywords.sort�
```
svenefftinge: The sort seems to be superfluous, as keywordHelper.allKeywords returns a sorted collection.

spoenemann: Right.

spoenemann:
Yes, no changes there. This fix should only make a difference for ignoreCase languages, and who wants them, anyway?

PR #43, opened 2016-07-19
removed src-gen folder as source folder in .ide.tests

LorenzoBettini:
The src-gen folder is not used in this project and should be removed
from the .classpath to avoid build errors in Eclipse.


spoenemann:
👍

Issue #24, opened 2016-07-01
[LanguageServer] Workspace Symbol service not working


cdietrich:
VSCode does not how Workspace Symbols although Xtext Sends them

```
Client Request:
	{"jsonrpc":"2.0","id":4,"method":"workspace/symbol","params":{"query":"S"}}
Server Response:
	{"id":"4","result":[{"name":"S1","kind":0,"location":{"uri":"file:///Users/dietrich/git/xtext-languageserver-example/demo/demo.statemachine","range":{"start":{"line":0,"character":6},"end":{"line":0,"character":8}}}}],"jsonrpc":"2.0"}

```

i dont know if the response is invalid or vscode is buggy

cdietrich:
Document Symbol Service is running fine

cdietrich:
P.S: it seems to work with the Insider build

svenefftinge:
Is this solved now that we return SymbolKind.Property by default?

svenefftinge:
We should file an issue over at vscode

cdietrich:
No it is fixed by the newest vs code version and the enum return. I am not sure how the protocol handles Enums when not specified.

svenefftinge:
So what would be the concrete action for this ticket?

cdietrich:
do we ensure we return a valid enum value? if so we can close.

svenefftinge:
we never send 0 but null in case the kind is not set. Should be handled generically in ls-api.
See https://github.com/TypeFox/ls-api/issues/1

PR #44, opened 2016-07-19
lsapi update

cdietrich:
@spoenemann please review


spoenemann:
Looks good to me, but it would be good if @akosyakov or @svenefftinge could also review this.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.xtend
```diff
@@ -205,17 +206,17 @@ class DocumentSymbolService {
...
 		return description.EClass.symbolKind
 	}
 
 	protected def String getSymbolName(QualifiedName qualifiedName) {
 		return qualifiedName?.toString
 	}
 
-	protected def int getSymbolKind(EClass type) {
-		return 0
-//		return SymbolInformation.KIND_PROPERTY
```
svenefftinge: @akosyakov Do you remember why you used a hard coded 0 here?

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -169,7 +169,7 @@ abstract class AbstractLanguageServerTest implements Consumer<PublishDiagnostics
 
     protected def dispatch String toExpectation(SymbolInformation it) '''
         symbol "�name�" {
-            kind: �kind�
+            kind: �kind.value�
```
svenefftinge: The test expectation would probably be more readable if we were using the symbol names. But we can change that once someone is at it and wants to improve on that.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.xtend
cdietrich: see https://github.com/eclipse/xtext-core/issues/24 as well

akosyakov: I guess because i like an icon for an undefined kind in VSCode more than an icon for a property as a default.

PR #51, opened 2016-07-22
improved uri handling on windows

cdietrich:
This fixes the uri en- and decoding problems on windows.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/IFileSystemScanner.xtend
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/IFileSystemScanner.xtend
@@ -7,10 +7,10 @@
  *******************************************************************************/
 package org.eclipse.xtext.util
 
-import org.eclipse.xtext.util.IAcceptor
-import org.eclipse.emf.common.util.URI
 import com.google.inject.ImplementedBy
 import java.io.File
+import java.nio.file.Paths
+import org.eclipse.emf.common.util.URI
 
 /**
  * @author Sven Efftinge - Initial contribution and API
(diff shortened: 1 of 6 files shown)


cdietrich:
@svenefftinge @spoenemann can you have a look

org.eclipse.xtext.util/src/org/eclipse/xtext/util/IFileSystemScanner.xtend
```diff
@@ -29,13 +29,19 @@ interface IFileSystemScanner {
...
+            acceptor.accept(uri)
             if (file.isDirectory) {
                 for (f : file.listFiles) {
                     scanRec(f, acceptor)
                 }
             }
         }
         
+        def String toPath(java.net.URI uri) {
+            val path = Paths.get(uri)
```
svenefftinge: What is this conversion doing?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/IFileSystemScanner.xtend
```diff
@@ -29,13 +29,19 @@ interface IFileSystemScanner {
...
+            val uri = URI.createURI(file.absoluteFile.toURI.toPath)
+            acceptor.accept(uri)
             if (file.isDirectory) {
                 for (f : file.listFiles) {
                     scanRec(f, acceptor)
                 }
             }
         }
         
+        def String toPath(java.net.URI uri) {
```
svenefftinge: the name 'toPath' is missleading if it returns a string representation of a URI.

cdietrich:
@svenefftinge i added some comments why we need this convertion and inlined the method.
btw the very same toPath naming is inside UriExtension

svenefftinge:
It's a bad name there, too. Maybe we should move a well named implementation to UriUtil?

cdietrich:
which name would you prefer? can you file a separate ticket for that?

svenefftinge:
+1 on merging this

Issue #21, opened 2016-06-30
[LanguageServer] Add support for formatting and rangeFormatting


cdietrich:

cdietrich:
see https://github.com/eclipse/xtext-core/tree/cd_issue21 for a first draft

cdietrich:
resolved with 3d49daa819a55f305809776501ef24f114f2b7e0

PR #54, opened 2016-07-24
[Language Server] Support for Formatting

cdietrich:
Implemented Formatting for LanguageServer #21


cdietrich:
please review

svenefftinge:
Cool! 
I don't like to support two different formatting hooks in future. I think we should deprecate the old one and support only the new one here. @meysholdt what do you think?

cdietrich:
well the old formatter is still the default

svenefftinge:
I think that should be fixed.

cdietrich:
so i will remove the support for the old formatter here and simly return a empty list?

meysholdt:
@svenefftinge yes, it's a good idea to not support the old formatter.
@cdietrich In Xtext languages that use the new generator infrastructure, the new formatter is the default. Only when you use the old generator infrastructure *and* don't have a FormatterFragment in your workflow, then the old formatter is bound in Guice by default. That was for API compatibility purposes, but I don't see any practical use case for it, because the old formatter by default only turned every document into a single line.

svenefftinge:
So we should deprecate the old one and make the new one public API soon, but it is not mandatory to only support the new formatter in LSP. Correct?

cdietrich:
@meysholdt  why is the old formatten then used inside the test language?

cdietrich:
i removed the support for the old formatter now.

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/TestLanguageIdeNewFormatterSetup.xtend
```diff
@@ -0,0 +1,56 @@
...
+import org.eclipse.xtext.ide.tests.testlanguage.TestLanguageStandaloneSetup
+import org.eclipse.xtext.ide.tests.testlanguage.formatting2.TestLanguageFormatter
+import org.eclipse.xtext.ide.tests.testlanguage.testLanguage.TestLanguagePackage
+import org.eclipse.xtext.resource.IResourceFactory
+import org.eclipse.xtext.resource.IResourceServiceProvider
+
+/**
+ * Initialization support for running Xtext languages without Equinox extension registry.
+ */
+class TestLanguageIdeNewFormatterSetup extends TestLanguageStandaloneSetup {
```
meysholdt: could you make the formatter impl part of the TestLanguage? Then this special setup should not be necessary anymore.

cdietrich: so i should not test the case having no formatter?

svenefftinge: Yes. testing that is not worth the extra setup.

cdietrich: ok ill remove that ;)

meysholdt:
thx. please rebase & squash the commits from this PR, then I think it's ready for being merged. 
see https://github.com/eclipse/xtext/blob/master/QUALITY_GUIDELINES.md#commits-and-pull-requests

Issue #67, opened 2016-08-10
[Junit4Fragment2] Exported test package becomes unneccessary public API


kthoms:
The Junit4Fragment2 fragment exports the test base package. The original requirement for this was [bug#378254](https://bugs.eclipse.org/bugs/show_bug.cgi?id=378254) in order to trigger a classpath update.

The exported package won't be used by any downstream projects, but becomes public API. This becomes problematic for projects that have API tooling enabled also for the test projects and don't want to expose that package by default. The package would be exported after deletion again when regenerating the language.

The reason to export that package seems to be just a workaround for a PDE bug. It should be tested if the bug still exists. If not, getExportedPackagesTests() can return null again. 

If the PDE still exists, or the workaround should be kept to be backward compatible, it would also help to mark the exported package as internal. This would also trigger the desired classpath update.

kthoms:
Merged

PR #68, opened 2016-08-10
[#67] Mark exported test packages as internal

kthoms:


svenefftinge:
Not sure why this is important. But if you have tested it than i think it can be merged.

kthoms:
The reason can be read in issue#67. The GEF project does not want to have the test packages exported, so they are deleting the entry from the manifest. After regeneration the package is exported again.
Yes, I tested that, but the main issue to discuss if it is OK to make the API now internal for existing test projects. I think it is reasonable, thus I consider this a valid fix.

kthoms:
Thanks!

PR #86, opened 2016-08-16
[478377] Add null guard

kthoms:
This guard adds more safety, but the actual NPE reason is fixed by
   https://github.com/eclipse/xtext-xtend/pull/26


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=478377

meysholdt:
👍

PR #89, opened 2016-08-24
[500192] IdeContentProposalProvider - Don't call Scope Provider with null model

cdietrich:
[500192] IdeContentProposalProvider - Don't call Scope Provider with null model


cdietrich:
please review

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=500192

PR #90, opened 2016-08-24
[329125] Avoid unnecessary parentheses in generated Antlr grammars

stadlerb:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=329125

This changes the Antlr Grammar Generator to avoid duplicate parentheses as ANTLR3's parser generator seems to have trouble generating correct parsers with backtracking when duplicate parentheses meet alternatives.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=329125

spoenemann:
This is a quite big change. Could someone with some Antlr experience please review it?

JanKoehnlein:
The Antlr grammar generator fragments are used by *all* of our users, and a change could potentially break quite a few of them. 

We should at least re-generate all test grammars and check for the impact from this change. Given that the real problem only appears when ```backtracking``` is enabled, which is really a discouraged setting, I'd rather not risk changing the Antlr generator. 

If we're sure it works, we should always enable it. If we're unsure, we could also consider shipping separate fragments for the backtracking users instead of adding a boolean parameter everywhere.

Opinions?

cdietrich:
i'd prefer a separate fragment. backtracking is discouraged anyway since it leads to unexpected behaviour all over the place e.g. in content assist

spoenemann:
@JanKoehnlein, @cdietrich: sounds sensible. Extracting the backtracking part to a new fragment would require some more work. Should we close this PR and describe what is to be done in a new issue?

JanKoehnlein:
See eclipse/xtext#2589.

PR #96, opened 2016-09-12
[378980] Make line delimiter for Manifest configurable by CodeConfig

kthoms:
Line delimiter is determined by CodeConfig and reached into MergeableManifest.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest.java
@@ -50,38 +50,51 @@ public class MergeableManifest extends Manifest {
 	public static final Attributes.Name BUNDLE_LOCALIZATION = new Attributes.Name("Bundle-Localization");
 	public static final Attributes.Name BUNDLE_ACTIVATOR = new Attributes.Name("Bundle-Activator");
 
-	private static final String LINEBREAK = "\r\n";
+	/**
+	 * @deprecated Is only used in deprecated static methods.
+	 */
+	@Deprecated
+	private static final String LINEBREAK = Strings.newLine();
 
 	/*
 	 * java.util.Manifest throws an exception if line exceeds 512 chars
 	 */
 	/**
+	 * @deprecated Use {@link #make512Safe(StringBuffer, String)} instead
 	 * @since 2.9
 	 */
-    public static String make512Safe(StringBuffer lines) {
-        if (lines.length() > 512) {
-        	StringBuilder result = new StringBuilder(lines.length());
-        	String[] splitted = lines.toString().split("\\r?\\n");
-        	for(String string: splitted) {
-        		if (string.length() > 512) {
-        			int idx = 510;
-        			StringBuilder temp = new StringBuilder(string);
-        			int length = temp.length();
-        			while (idx < length - 2) {
-                        temp.insert(idx, LINEBREAK+" ");
-                        idx += 512;
-                        length += 3;
-                    }
-        			result.append(temp.toString());
-        		} else {
-        			result.append(string);
-        		}
-        		result.append(LINEBREAK);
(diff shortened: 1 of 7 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=378980

meysholdt:
\r\n is hard-coded in java.util.jar.Manifest.write(OutputStream), even though the [spec](http://docs.oracle.com/javase/7/docs/technotes/guides/jar/jar.html) allows for \r\n and \n

kthoms:
Right, I was not aware of that. So does this mean since standard Java Manifest uses \r\n it is also mandatory for MergeableManifest to do the same? Seems so, i.e. I should close the bug as WONTFIX?

meysholdt:
Further investigation shows that `java.util.jar.Manifest.read(InputStream)` supports both `\r\n` and `\n`. Also, the Eclipse Manifest Editor supports both `\r\n` and `\n\`. Furthermore, the `MANIFEST.MF`s in xtext-core only contain `\n`. So I'm fine with making line-edings configurable in `MergeableManifest`. 

Can you confirm that your change doesn't mix line-edings, e.g. because a method from the super-class (`Manifest`) adds `\r\n` but in `MergeableManifest` `\n` is configured?

kthoms:
screenshot 5

This is how it looks like after generating. The example above is without any configuration, it will use Strings.newLine() by default which is "\n". You can see all line endings are such.

The second example shows explicit configuration of "\r\n". Again, all line endings are the same.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest.java
```diff
@@ -50,38 +50,47 @@
 	public static final Attributes.Name BUNDLE_LOCALIZATION = new Attributes.Name("Bundle-Localization");
 	public static final Attributes.Name BUNDLE_ACTIVATOR = new Attributes.Name("Bundle-Activator");
 
-	private static final String LINEBREAK = "\r\n";
+	private static final String LINEBREAK = Strings.newLine();
```
meysholdt: this filed is only used by deprecated methods, right? Then it should be deprecated itself.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/ManifestAccess.xtend
```diff
@@ -38,6 +38,8 @@ class ManifestAccess extends TextFileAccess implements IGuiceAwareGeneratorCompo
 	
 	TypeReference activator
 	
+	String lineDelimiter = "\r\n"
```
meysholdt: I think here the configuration from `CodeConfig` should be considered.

kthoms: Not really. It is just the default value, which is overwritten by setLineDelimiter (available through @Accessors). But the default should be Strings.newLine(). I'll change that.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest.java
kthoms: Correct, it is only used in deprecated contexts. Will fix that.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/ManifestAccess.xtend
meysholdt: ok

meysholdt:
thx!

kthoms:
Thanks for the review :)

PR #104, opened 2016-09-14
[#5] Ignore optional compile problems on *-gen source folders

kthoms:

### DIFF
--- a/org.eclipse.xtext.tests/.classpath
+++ b/org.eclipse.xtext.tests/.classpath
@@ -12,11 +12,13 @@
 	</classpathentry>
 	<classpathentry kind="src" path="src-gen">
 		<attributes>
+			<attribute name="ignore_optional_problems" value="true"/>
 			<attribute name="FROM_GRADLE_MODEL" value="true"/>
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="src" path="emf-gen">
 		<attributes>
+			<attribute name="ignore_optional_problems" value="true"/>
 			<attribute name="FROM_GRADLE_MODEL" value="true"/>
 		</attributes>
 	</classpathentry>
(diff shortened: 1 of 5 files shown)


LorenzoBettini:
Just my thoughts: I usually ignore optional compile problems in src-gen and emf-gen, but never on xtend-gen: Java files generated by Xtend are basically the same as the ones written directly in xtend, so a problem in the generated Java files is likely to be a real problem in the xtend file itself. By the way, there are cases where the Xtend compiler does not detect a compilation error in the xtend file, it generates the Java file and then reports the compilation problem taken from the generated Java file.
What do you think?

kthoms:
It is only regarding *optional warnings*, not errors. This is for example unused imports or variables, raw types etc. Generated Xtend classes have usually `@SuppressWarnings("all")`, so it is usually obsolete to suppress optional compile errors on the xtend-gen folders.

Shall I remove the settings for that?

LorenzoBettini:
You're right. But if I'm not wrong it should now be possible to avoid the suppress warnings in the generated files and in that case, excluding the xtend-gen folder would not show possible warnings. What I'm saying is that, by default, files in xtend-gen folders will not show any warnings, even if not excluded. But if you exclude them, this would prevent warnings from showing up if someone remove the suppress warnings. Or am I missing something?

kthoms:
I will remove the option for xtend-gen folders and update all PRs.

org.eclipse.xtext.tests/.classpath
```diff
@@ -32,6 +34,7 @@
 	</classpathentry>
 	<classpathentry kind="src" path="generator/xtend-gen">
```
LorenzoBettini: you forgot this one

PR #107, opened 2016-09-19
[406914] Explicitly map identifiers for tab, cr and lf

kthoms:
The underlying ICU transliterator maps the special characters to
'Control'<code>, so these characters have to be mapped without using
ICU. Further to keep camel case of the mapped text the text is mapped
directly in toJavaIdentifierSegment.

Signed-off-by: Karsten Thoms karsten.thoms@itemis.de


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=406914

spoenemann:
What about tests for that issue? The original issue mentioned

> This bug occurred in /org.eclipse.xtext.tests/src/org/eclipse/xtext/parser/antlr/Bug301935TestLanguage.xtext

kthoms:
Updated.

JanKoehnlein:
How about \f, \b or unicode chars with \u?
Are ambiguities handled, e.g. for a rule

Foo:
   '\n' | 'newline';

kthoms:
Ambiguities are handled as before, they get a count suffix. See new test language Bug406914TestLanguage.
Handled additionally \b, \f, \\.
For unicode the behavior has not changed and is beyond scope of the referred bug.

kthoms:
Shall I extend the new example by a keyword with unicode?

kthoms:
Is there more action required now? Please review.

spoenemann:
Looks good now.

Issue #108, opened 2016-09-19
[ecore2xtext] Use new formatter fragment


kthoms:
The Ecore2Xtext wizard creates a formatter using the old formatter fragment:

```
fragment = org.eclipse.xtext.generator.adapter.FragmentAdapter {
  fragment = org.eclipse.xtext.generator.ecore2xtext.FormatterFragment {}
}
```

I think Xtext should not encourage this and use the new formatter fragment instead:

```
formatter = { generateStub = true }
```

kthoms:
fixed via #109 
merge commit c93db76702e751b78cef8d1eb8d4b3b2be85becd

PR #109, opened 2016-09-19
[#108] Ecore2Xtext use new formatter fragment

kthoms:


LorenzoBettini:
If I understand correctly, the Formatter2Fragment2 does something like

	protected def doGenerateStubFile() {
		if(!isGenerateStub)
			return;
			
		if(isGenerateXtendStub) {...
		} else {
			LOG.error(this.class.name + " has been configured to generate a Java stub, but that's not yet supported. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=481563")	
		}
	}

Thus, if you specify generateStub=true but no generateXtendStub you'll get an error during the mwe2 run, is that correct?

kthoms:
isGenerateXtendStub returns true without setting the value explicitly. Have a look into AbstractStubGeneratingFragment#isGenerateXtendStub. Since codeConfig has preferXtendStubs=true by default and the property is not explicitly set, the value isGenerateXtendStub defaults to preferXtendStubs.

LorenzoBettini:
OK! I couldn't try that right now and I remember that in the past I had that problem, but something must have changed in the meantime :)

LorenzoBettini:
@kthoms I think this should now be rebased/merged with master due to the other merged PRs.

kthoms:
@LorenzoBettini Rebased. Please review again.

Issue #112, opened 2016-09-20
[generator] Generate launch config


kthoms:
With previous Xtext releases a launch config `Launch Runtime Eclipse.launch` was generated to the `.launch` folder of the runtime project. This feature got lost with the new generator infrastructure.

svenefftinge:
Sorry, I forgot what value these launch configs bring over the default ones eclipse would generate.

kthoms:
The "Launch Runtime Eclipse.launch" is more or less the same as if you would create one with Eclipse. Differences are
1) You don't have to create one, it's provided
2) A default launch config would be named "New_configuration" and the runtime workspace is "runtime-New_configuration". With the provided one the workspace is fixed.
3) The launch config is persisted by default and can be shared, which may not be known to all users how to do this with a launch config

The second launch config is to generate the Xtext infrastructure with. This could be achieved usually with right-click on the grammar.

The original question on slack was if it was a bug or feature that the launch configs are not generated anymore. It was said that I should consider it as a bug, since a feature got lost during the generator migration. Thus I restored the feature with the new generator.

svenefftinge:
I don't recall if we explicitly avoided generation of them, but now that we are thinking about it I'd like to not have them, because the default behavior of Eclipse is good and we don't need to pollute the project tree. In the past the defaults weren't good (e.g. because of a too small permanent generation size) and we were Eclipse-only so it made more sense.
I'm in favor of closing this.

kthoms:
Before actually closing this: I agree that the launch config for the runtime Eclipse is not really necessary, especially since when you create multiple DSL projects in the workspace, there would be multiple of them all with the same name and starting the same environment. One manually created would be enough. And we should assume that developers who want to develop Eclipse plugins should be familiar on how to create such a launch config.

What I would like to discuss the value of the second launch config, which starts the MWE workflow. Usually a user would generate the artifacts by using the context menu actions. Having a dedicated launch config could be still useful when a DSL is part of a workspace materialized with Oomph without checking in generated sources. In this case an Oomph setup could be instrumented to run the launch config, which otherwise would have to be created by the user manually. It would be convenience feature here. The type of launch config would be dependent on the selected build system and project configuration, which is respected by the PR.

I could live if this issue is voted to be a WONTFIX. However, this was a silently dropped feature with 2.9 (and actually using the old generator it would still exist), and it should be documented if a feature is discontinued.

cdietrich:
[...]

PR #113, opened 2016-09-20
[#112] Generate launch configs for Xtext workflow and Runtime Eclipse

kthoms:
- Launch config for Runtime Eclipse is only generated when the runtime
project is a plugin project.
- Launch config for Xtext workflow is dependent on the build system. For
plain Maven projects the launch config uses the m2e launcher, for Gradle
builds buildship is launched.

### DIFF
--- a/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
+++ b/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
@@ -7,11 +7,10 @@
  *******************************************************************************/
 package org.eclipse.xtext.xtext.wizard
 
+import org.eclipse.xtend.lib.annotations.Accessors
 import org.eclipse.xtext.xtext.wizard.ecore2xtext.Ecore2XtextGrammarCreator
 
 import static org.eclipse.xtext.xtext.wizard.ExternalDependency.*
-
-import org.eclipse.xtend.lib.annotations.Accessors
 
 class RuntimeProjectDescriptor extends TestedProjectDescriptor {
 
@@ -113,6 +112,10 @@ class RuntimeProjectDescriptor extends TestedProjectDescriptor {
 		files += super.files
 		files += grammarFile
 		files += file(Outlet.MAIN_JAVA, workflowFilePath, workflow)
+		files += workflowLaunchConfigFile
+		if (config.runtimeProject.isEclipsePluginProject) {
+			files += launchConfigFile
+		}
 		if (isPlainMavenBuild) {
 			files += file(Outlet.ROOT, "jar-with-ecore-model.xml", jarDescriptor)
 		}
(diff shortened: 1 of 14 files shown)


kthoms:
Covered by test `CliWizardIntegrationTest`

LorenzoBettini:
Wouldn't it be better to generate .launch files into a directory NOT hidden by default? For example, "launches" or simply "launch"?

kthoms:
I restored the behavior as it was in Xtext < 2.9, extended by the behavior for different build systems.

See https://github.com/eclipse/xtext/blob/5aac96d0173ae356978e48e23eb59c5c591e2443/plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/DslProjectContributor.xtend , line 29

LorenzoBettini:
Yes, I know that's the old behavior... I was wondering whether it would be better to have the directory for launch configurations visible... just my thoughts; otherwise I'm fine with the old behavior :)

kthoms:
Perhaps we should discuss this in a separate issue.

kthoms:
The Eclipse Launch config is only generated when it was chosen that an Eclipse plugin should be created. When you select plain Gradle or Maven builds the this launch config would not be produced.
But a tool specific launch config for building the DSL is produced. This either is a MWE2 launcher when using plugin builds or m2e for plain Maven or buildship launch for Gradle builds.

Does that fit your needs? At the end this PR restores a behavior that got lost during the generator migration without an issue that it should be removed.

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
```diff
@@ -528,4 +531,130 @@ class RuntimeProjectDescriptor extends TestedProjectDescriptor {
 		#{"**/*.xtend","**/*.mwe2"}
 	}
 	
+	/**
+	 * @since 2.11
+	 */
+	def private getWfLaunchConfigFile() {
```
LorenzoBettini: "wf" might not be trivial to understand immediately... how about getWorkflowLaunchConfigFile? Or getMWE2LaunchConfigFile?

kthoms: I'm fine with the suggestion. The name was chosen since it was the same method name before. But since it is private it is fair enough to make it more readable. I'll change that.

kthoms: Done. Please check again.

LorenzoBettini: For me it's OK to merge. By the way, I cannot check it now: the @since annotations are required by the API tooling for public and protected methods; For private methods they are not required, but do you get any error from the API tooling if you have @since annotations?

kthoms: They are not needed for private methods, I consider them in this case just as annotation that the code was added with 2.11. Usually they are used to specify that *API* was added with a version.
API tooling does not work for me right now. I asked already on Slack how we handle API in Xtend without answer yet.

LorenzoBettini: as far as I know, you have to look at the error markers in the generated Java files :)
anyway, it OK to merge for me

kthoms: then merge it please :)

Issue #118, opened 2016-09-26
[Junit4Fragment2] Improve generated test code


kthoms:
The currently generated test code looks as follows:

```
@RunWith(XtextRunner)
@InjectWith(MyDslInjectorProvider)
class MyDslParsingTest{

    @Inject
    ParseHelper<Model> parseHelper

    @Test 
    def void loadModel() {
        val result = parseHelper.parse('''
            Hello Xtext!
        ''')
        Assert.assertNotNull(result)
    }

}
```

This test would also pass if the parsed model is invalid, e.g. when testing `Hello Xtext` instead of `Hello Xtext!`. Using the ValidationTestHelper would improve that.

Using the with-Operator could make the test more readable. And when Assert is imported as static extension, the code would be even more fluent.

Next, the method name "loadModel" is not appropriate. A model is not loaded from somewhere, it is just parsed from inline text.

Therefore I propose that the generated test code should look like this:

```
@RunWith(XtextRunner)
@InjectWith(MyDslInjectorProvider)
class MyDslParsingTest{

    @Inject extension ParseHelper<Model>
    @Inject extension ValidationTestHelper

    @Test 
    def void testModel() {
        '''
            Hello Xtext!
        '''.parse => [
            assertNotNull
            assertNoErrors
        ]
    }
}
```

LorenzoBettini:
I think this is somehow related to https://bugs.eclipse.org/bugs/show_bug.cgi?id=492034

I think the parsing test should only test parsing and the test should pass also with "Hello world" because the parse can actually create a model out of that, shouldn't it?

kthoms:
I already encountered situations where users where confused by the fact that the test passes also with an invalid model. They were even not aware of the fact that there is the ValidationTestHelper, and usually one would only write tests for models which either are valid or invalid. For the case that someone would like to test an invalid model, this should be also tested with the test helper.

What do you think?

kthoms:
Maybe the test class should be even extended by a negative test?

kthoms:
```
	@Test
	def void testInvalidModel () {
		'''
			Hello Xtext
		'''.parse => [
			assertNotNull
			assertError(it.eClass, Diagnostic.SYNTAX_DIAGNOSTIC, "expecting '!'")
		]
	}
```

LorenzoBettini:
As I said in my original bugzilla bug, I'd like to have more test use cases generated, but I agree with Sven that the generated test class is really a starting point (and only for parsing not for validation).

[...]

PR #119, opened 2016-09-26
[#118] Improve generated test code

kthoms:
Signed-off-by: Karsten Thoms karsten.thoms@itemis.de

### DIFF
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/Junit4Fragment2.xtend
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/Junit4Fragment2.xtend
@@ -82,19 +82,18 @@ class Junit4Fragment2 extends AbstractStubGeneratingFragment {
 		return fileAccessFactory.createXtendFile(exampleRuntimeTest, '''
 			@«runWith»(«xtextRunner»)
 			@«injectWith»(«injectorProvider»)
-			class «exampleRuntimeTest»{
-			
+			class «exampleRuntimeTest» {
 				@«Inject»
 				«parseHelper»<«rootType»> parseHelper
-			
-				@«test» 
+				
+				@«test»
 				def void loadModel() {
 					val result = parseHelper.parse(''«»'
 						Hello Xtext!
 					''«»')
 					«assert».assertNotNull(result)
+					«assert».assertTrue(result.eResource.errors.isEmpty)
 				}
-			
 			}
 		''')
 	}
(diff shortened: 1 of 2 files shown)


LorenzoBettini:
Please have a look at my comment/suggestion on the issue https://github.com/eclipse/xtext-core/issues/118

kthoms:
I still disagree. Right, the test tests that the DSL model is parsed. The test never fails, even if the model is invalid. The error is not detected. This leads to confusion.

See for example:
http://stackoverflow.com/questions/12703318/testing-that-something-fails-to-parse-in-xtext
https://www.eclipse.org/forums/index.php?t=rview&goto=1742997#msg_1742997

cdietrich:
then at least we need a result.eResource.error.isempty

spoenemann:
> then at least we need a result.eResource.error.isempty

Yes, this solution could make it more clear that we're only interested in parser errors, not validation errors.

kthoms:
OK for me, I'll update the PR.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/Junit4Fragment2.xtend
```diff
@@ -82,19 +82,18 @@ class Junit4Fragment2 extends AbstractStubGeneratingFragment {
...
 			@�runWith�(�xtextRunner�)
 			@�injectWith�(�injectorProvider�)
-			class �exampleRuntimeTest�{
-			
+			class �exampleRuntimeTest� {
 				@�Inject�
 				�parseHelper�<�rootType�> parseHelper
-			
-				@�test� 
+
```
cdietrich: @kthoms is this tab here by intention?

kthoms: Yes, it is. It is basically a newline between the injection and the test method. The tab is there in this new line with class-level indentation.

svenefftinge:
+1 for the change. 
Generally I prefer to generate less code, especially generate-once stuff.
The kind of stuff we do here should be something for a wizard not for the generator.
We just put it here, because we didn't want to have compile errors after the wizard did his job.

cdietrich:
@kthoms is this pull request uptodate?

kthoms:
Yes, looks so.

cdietrich:
thx @kthoms

dhuebner:
@cdietrich Could you please verify that core build failure is not caused by the merged change?http://services.typefox.io/open-source/jenkins/job/xtext-core/job/master/220/

cdietrich:
in eclipse the tests are green
ill see what happens in gradle

cdietrich:
gradle tests are green locally as well

cdietrich:
@dhuebner can you try this on your local machine?

dhuebner:
@cdietrich I will check the upstream changes first. xtext-xtend fails too. :(

dhuebner:
@cdietrich Probably caused by the latest xtext-lib change. Thx for verification

cdietrich:
i locally find RequestManagerTest and EncodingTest and UnicodeTest as flaky

Issue #121, opened 2016-09-29
[grammar] prefix overriding parser rules with 'override' keyword


meysholdt:
Java introduced the @Override annotation in Java so that developers get feedback from the compiler when a super class changes a method signature but the sub-class still expects to override it.

We face the same scenario in grammars: When a parser rule is being renamed in a super grammar and there is a parser rule in a sub-grammar overriding that parser rule, currently developers of the sub-grammar get no feedback indicating that they need to rename their parser rule as well. Instead, the parser rule silently turns from being an 'override' into a new declaration.

We could prefix parser rules with the 'override' keyword when they are expected to override a parser rule from a super-grammar. This would give developers an error message if the parser rule from the super grammar is not available anymore. This would also improve readability of a grammar because it would make it explicit where override happens. 

For backward compatibility, overriding rules without using the 'override' keyword would still need to work. However, this should cause a warning. A quick-fix could resolve the warning by adding the 'override' keyword.

Introducing a new keyword to the language may cause conflicts with existing identifiers. In the grammar language, we have identifiers for epackage-aliases, EClasses, EStructuralFeatures, Rules. Only aliases and EStructuralFeatures are candidates for conflicts, as Rule-names and EClass-names start with an uppercase letter by convention. The 'override' keyword could be allowed as an identifier for aliases and EStructuralFeatures.

meysholdt:
@JensN4 describes this al follows in [bug503311](https://bugs.eclipse.org/bugs/show_bug.cgi?id=503311):

	When a grammar extends another one, e.g.
	
	----
	grammar SubGrammar with SuperGrammar
	----
	
	rules can be overridden. This is simply done by using the same rule name.
	
	It would great if an overriding rule could be marked as "override", similar to the Java/Xtend @Override/override feature. Xtext already checks the correct "return type" (the model element created by the rule).
	
	Example (using Xpect notation):
	
	Given 
	----
	grammar SuperGrammar 
	Program: (x+=Item | x+=Element)*;
	Item: "item" prop=IDENTIFIER;
	Element: "element" prop=IDENTIFIER;
	----
	
	and
	----
	grammar SubGrammar with SuperGrammar
	override Program: (x+=Element | x+=More)*;
	
	// XPECT warning --> "rule Element overrides SuperGrammar.rule, should be marked with 'override'"
	Element: "elem" prop=IDENTIFIER;
	
	// XPECT error --> "rule More does not override any super rule"
[...]

PR #126, opened 2016-10-05
[grammar] 'override' keyword

sailingKieler:
Updated!

Introduces `@Override` as desired by #121.
* added a corresponding attribute to `AbstractRule` in `Xtext.ecore`
* added an optional leading keyword '@Override' in the rules `ParserRule`, `TerminalRule`, and `EnumRule`

Comments, suggestions, and approvals are welcome :-)


org.eclipse.xtext/src/org/eclipse/xtext/Xtext.xtext
```diff
@@ -251,6 +253,7 @@ CharacterRange returns AbstractElement:
 ;
 
 EnumRule:
+	(^override?='@Override')?
```
meysholdt: `^` is now not necessary anymore

sailingKieler: probably not

svenefftinge:
Sorry, if my proposal wasn't clear. I meant to introduce the notion of annotations in the grammar.
So a rule would be preceded by any number of annotations.
Something like this:

```
ParserRule :
   (annotations+=Annotation)*
 ....

Annotation : 
   '@' name=ID
;
```

svenefftinge:
Would it be a lot of work to do the bootstrap stuff separately?

sailingKieler:
No I don't think so. I can cherry-pick the grammar change so the bootstrapping stuff can be merged separately.

sailingKieler:
Regarding the explicit annotations:
I got it and agree, but we doubt we can achieve that solution in the currently available time, so we prefer the interim solution.

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
```diff
@@ -1234,5 +1234,32 @@ public void checkOppositeReferenceUsed(Assignment assignment) {
...
+					break;
+					
+				} else {
+					// type compatibility checking is currently done in 'Xtext2EcoreTransformer' being called during linking
+					//  so it's omitted here
+				}
+				
+			} else {
+				if (rule.isOverride()) {
+					error("None of the used grammars contain a rule named " + name, rule,
```
svenefftinge: "used grammars" is not a user-level term. 
How about "The rule 'Foo' must override a rule from a super grammar"?

svenefftinge:
Maybe I'm missing something, but it shouldn't be more work.
You'd basically need to implement `isDeprecated` which internally checks for name=='Override' but not much more.

svenefftinge:
Actually, I would like to see the "SuppressWarnings[code1, code2]" thing Sebastian added one year ago to be replaced with proper annotations, too.
See https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextDiagnosticConverter.java#L94

knutwannheden:
Making `Xtext.xtext` inherit from `XbaseWithAnnotations.xtext` might propel the bootstrapping to the next level of fun! ;-)

sailingKieler:
@knutwannheden for sure! :-)

sailingKieler:
Just pushed my updated draft.
`AbstractRule` now has a list of `Annotation`s, we however permit just one annotation per rule in the concrete syntax for now. A dedicated check ensures that only `@Override` is accepted.

sailingKieler:
Removed the bootstrap project from the changes of this PR.
Added corresponding test in `XtextValidatorTest` testing the four scenarios:
* no rule overriding + no annotation
* rule overriding + no annotation
* no rule overriding + annotation
* rule overriding + annotation

sailingKieler:
From my point of view, the PR can be merged,
and afterwards the branches 'cs/override' on xtext-eclipse + xtext-idea as well.

org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/XtextValidationTest.java
```diff
@@ -102,6 +102,42 @@ private void configureValidator(XtextValidator validator, ValidationMessageAccep
...
+		assertEquals("diag.isError", diag2.getSeverity(), Diagnostic.ERROR);
+		
+		XtextResource resource3 = getResourceFromString(
+				"grammar org.foo.Bar with org.eclipse.xtext.common.Terminals\n" +
+				"@Override\n" +
+				"terminal ID: ('a'..'z'|'A'..'Z'|'_');");
+		List<Issue> issues3 = validator.validate(resource3, CheckMode.FAST_ONLY, CancelIndicator.NullImpl);
+		assertEquals(issues3.toString(), 0, issues3.size());
+	}
+
```
sailingKieler: Sure.

sailingKieler: Done.

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
```diff
@@ -1234,5 +1235,45 @@ public void checkOppositeReferenceUsed(Assignment assignment) {
...
+		
+		final boolean isOverride =
+				rule.getAnnotations().stream().anyMatch(e -> AnnotationNames.OVERRIDE.equals(e.getName()));
+		
+		for (Grammar g : GrammarUtil.getGrammar(rule).getUsedGrammars()) {
+			final AbstractRule r = GrammarUtil.findRuleForName(g, rule.getName());
+			
+			if (r != null) {
+				if (!isOverride) {
+					warning("This rule overrides " + name + " in " + GrammarUtil.getGrammar(r).getName() + ".", rule,
```
meysholdt: * The warning message should also indicate what to do about it:
`This rule overrides Foo in Bar and thus should be annotated with @Override".

* please add an Issue Code, so that later a QuickFix can be provided.

meysholdt:
@svenefftinge could you review this PR?

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
```diff
@@ -1234,5 +1235,45 @@ public void checkOppositeReferenceUsed(Assignment assignment) {
 			}
 		}
 	}
-
+	
+	@Check
+	public void checkAnnotations(AbstractRule rule) {
```
svenefftinge: We should remove this check, because the reason I asked for generic annotations is to allow users to add arbitrary ones without changing our code. E.g. to use them in custom generator fragments.

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
svenefftinge: +1

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
knutwannheden: How about an error validation that can be disabled in the preferences (or even using a system property)? Because some users would certainly be willing to add the declarations once that is required. Then they could at least already benefit from the new syntax. For many use cases (like @SuppressWarnings) that would of course also require some syntax for parameters.

meysholdt: @svenefftinge and I discussed offline that it would be OK for now to allow arbitrary annotations without requiring annotations declarations. In the future, if we choose to introduce annotations declarations, we can offer a smooth migration path by rising a warning (not an error!) for annotations without declarations.

sailingKieler:
Updated the PR wrt. the requested changes.

sailingKieler:
Just fixed the test I forgot to update while fixing the error msg.

Issue #27, opened 2016-07-07
[LanguageServer] Provide DocumentHighlight Service (Mark Occurrences)


cdietrich:

cdietrich:
https://github.com/Microsoft/language-server-protocol/issues/32

cdietrich:
https://github.com/TypeFox/ls-api/issues/40

kittaakos:
I will look into this.

cdietrich:
i hat a look at this and it looks good. is there anything that is still left to do?

spoenemann:
I'd say it's done.

PR #137, opened 2016-10-10
[LanguageServer] Provide DocumentHighlight Service (Mark Occurrences)

kittaakos:
Added default document highlight implementation. #27


spoenemann:
This change depends on ls-api 0.4.0-SNAPSHOT, which includes TypeFox/ls-api#41. We should wait until the next ls-api or lsp4j release before we merge this PR.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/DocumentHighlightTest.xtend
```diff
@@ -0,0 +1,145 @@
...
+ * 
+ * @see DocumentHighlight
+ * @see IDocumentHighlightService
+ */
+public class DocumentHighlightTest extends AbstractTestLangLanguageServerTest {
+
+	@Test
+	def void singleLineNoOccurrences_SelectionBeforeWriteSiteMethodName() {
+		testDocumentHighlight[
+			model = '''type A { op foo() { } }''';
```
spoenemann: You could derive the line / column information directly from the text using a special character, e.g. `'''type A { op# foo() { } }'''` would select offset 11 in the text above (the `#` would be filtered out of the text).

cdietrich:
nice 👍

PR #143, opened 2016-10-13
DefaultIdeModule: Add means for controlling the lifecycle of executors

### LINKED ISSUE #127: DefaultIdeModule creates an Executor that is never shut down
DefaultIdeModule introduced for #11 does
```
binder.bind(ExecutorService).toProvider[Executors.newCachedThreadPool]
```
but the new thread pool is not shut down here. A possible improvement could be to pass a provider to the module like it is done for the DefaultWebModule. Then the creator of the module would be responsible to create and shut down executors. This would also make it quite easy to fix eclipse/xtext-web#6.

spoenemann:
Fixes #127. Also prepared changes required for eclipse/xtext-web#6.


akosyakov:
@spoenemann Where do we user `ExecutorService` by the way right now? only for completions?

spoenemann:
> @spoenemann Where do we user ExecutorService by the way right now? only for completions?

 * Completions in the language server
 * Completions in the web integration
 * Background work scheduling in the web integration (XtextWebDocumentAccess)

PR #168, opened 2016-11-05
[LSP] migrated to lsp4j 0.1.0

svenefftinge:


svenefftinge:
See https://github.com/TypeFox/ls-api/tree/lsp4j

PR #174, opened 2016-11-11
Added semantic highlight service interface as an LSP extension.

kittaakos:
Signed-off-by: Akos Kitta <kittaakos@gmail.com>


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/syntaxColoring/ISemanticHighlightClient.java
```diff
@@ -0,0 +1,34 @@
...
+public interface ISemanticHighlightClient {
+
+	/**
+	 * Pushes the {@link SemanticHighlight semantic highlight} to the client.
+	 * 
+	 * @param highlight
+	 *            the information that should be pushed to the client side. Must
+	 *            not be {@code null}.
+	 */
+	@JsonNotification("lsp4j/pushSemanticHighlight")
```
svenefftinge: Not sure about the 'push' prefix. Maybe 'updateHighlighting' would be better.
Also should we remove the 'Semantic' bit from the protocol, since it is an implementation detail how the Highlighting information was calculated.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/syntaxColoring/SemanticHighlight.xtend
```diff
@@ -0,0 +1,27 @@
...
+import org.eclipse.lsp4j.annotations.LanguageServerAPI
+
+/**
+ * Representation of a computed mapping from ranges to the appropriate
+ * highlighting style.
+ * 
+ * @author akos.kitta - Initial contribution and API
+ */
+@LanguageServerAPI
+class SemanticHighlight {
```
svenefftinge: How about 'updateHighlightingParams' (in symmetry to 'PublishDiagnosticsParams')

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/syntaxColoring/SemanticHighlight.xtend
```diff
@@ -0,0 +1,27 @@
...
+
```
svenefftinge: I think we need a 'uri' here

svenefftinge:
We discussed, that for general extensions, like semantic coloring we don't go through the language specific extension mechanism, but have an experimental extensions interface in LSP4J, that we implement and support in Xtext.

svenefftinge:
I have added the experimental definition https://github.com/eclipse/lsp4j/commit/ce3ff0ddf45011c164fa6c9bfefba886d9262c48

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/coloring/ColoringServiceImpl.xtend
```diff
@@ -0,0 +1,61 @@
...
+					emptyList;
+				}
+			nodes.forEach [
+				val start = offset;
+				val end = offset + length;
+				val range = new Range(document.getPosition(start), document.getPosition(end));
+				builder.add(new ColoringInformation(range, STYLE_IDS));
+			]
+		];
+		return new ColoringParams('''�resource.URI�''', builder.build);
```
svenefftinge: minor: we usually would use `resource.URI.toString`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/coloring/IColoringService.java
```diff
@@ -48,7 +46,7 @@
 
 		@Override
 		public ColoringParams getColoring(final XtextResource resource, final Document document) {
```
svenefftinge: Should the service only return a `List<? extends ColoringInformation>`? I hink the Params and URI should be set outside.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/coloring/ColoringParamsExtensions.xtend
```diff
@@ -0,0 +1,77 @@
...
+import org.eclipse.xtext.resource.XtextResource
+
+/**
+ * Extension for creating empty {@link ColoringParams coloring parameter}
+ * instances or checking whether existing ones are
+ * {@link ColoringParamsExtensions#isEmpty empty} or not.
+ * 
+ * @author akos.kitta - Initial contribution and API
+ */
+abstract class ColoringParamsExtensions {
```
svenefftinge: I made the properties in ColoringParams and ColoringInformation nonNull and initialized the lists with an empty list. With changing the service interface to return a List of ColoringInformation, could we get rid of this extension class?

svenefftinge:
Thank you!

Issue #180, opened 2016-11-21
File system access should return false instead of throwing an exception when the file does not exist.


kittaakos:
This problem occurs when the concrete implementation is the ```URIBasedFileSystemAccess```. Invoking the ```AbstractFileSystemAccess2.isFile(String, String)``` method with a non existing file path throws a ```FileNotFoundException``` instead of returning the expected ```false``` value. See below test case:

```
import com.google.common.base.StandardSystemProperty
import com.google.inject.Inject
import java.nio.file.Files
import java.nio.file.Paths
import java.util.UUID
import org.eclipse.emf.ecore.resource.impl.ExtensibleURIConverterImpl
import org.eclipse.xtext.generator.IFileSystemAccess
import org.eclipse.xtext.generator.IOutputConfigurationProvider
import org.eclipse.xtext.generator.URIBasedFileSystemAccess
import org.eclipse.xtext.testing.InjectWith
import org.eclipse.xtext.testing.XtextRunner
import org.eclipse.xtext.tests.XtextInjectorProvider
import org.junit.Assert
import org.junit.Before
import org.junit.Test
import org.junit.runner.RunWith

/**
 * @author akos.kitta - Initial contribution and API
 */
@RunWith(XtextRunner)
@InjectWith(XtextInjectorProvider)
class URIBasedFileSystemAccessTest {

	static val MISSING_RESOURCE_NAME = 'someMissingResource';
	static val EXISTING_RESOURCE_NAME = 'someExistingResource';

	@Inject
	URIBasedFileSystemAccess fsa;

	@Inject
	IOutputConfigurationProvider configProvider;
	
	@Inject
	ExtensibleURIConverterImpl uriConverter;

	@Before
	def void before() {
		val tmpPath = Paths.get(StandardSystemProperty.JAVA_IO_TMPDIR.value);
		val output = Files.createTempDirectory(tmpPath, '''tempFolder_«UUID.randomUUID»''');
		val resource = Files.createFile(output.resolve(EXISTING_RESOURCE_NAME));
		
		resource.toFile.deleteOnExit;
		output.toFile.deleteOnExit;
		
		val config = configProvider.outputConfigurations.head;
		config.outputDirectory = output.toString;
		fsa.outputConfigurations = #{ IFileSystemAccess.DEFAULT_OUTPUT -> config };
		fsa.converter = uriConverter; 
	}

	@Test
	def void testFalseOnAbsent() {
[...]

PR #181, opened 2016-11-21
Fixed issue in URI based file system access when checking missing files.

kittaakos:
A FNFE was thrown instead of returning with false.

Signed-off-by: Akos Kitta <kittaakos@gmail.com>


kittaakos:
Ticket: #180

PR #186, opened 2016-11-30
[#5] Ignore optional problems on *-gen folder

kthoms:
Modify classpath on Gradle import to set the ignore_optional_problems
flag for source folders ending with *-gen, but not for xtend-gen


cdietrich:
looks good to me.

PR #192, opened 2016-12-05
Some minor refactorings

kthoms:


org.eclipse.xtext/src/org/eclipse/xtext/formatting2/regionaccess/internal/TextSegment.java
```diff
@@ -45,7 +45,7 @@ public ITextRegionAccess getTextRegionAccess() {
 
 	@Override
 	public String toString() {
-		ToStringHelper helper = Objects.toStringHelper(this).add("offset", offset).add("length", length);
+		ToStringHelper helper = MoreObjects.toStringHelper(this).add("offset", offset).add("length", length);
```
svenefftinge: Have you checked that MoreObjects exists in all supported guava versions?

kthoms: The actual string representation is slightly different between `Objects#toStringHelper` and `ToStringBuilder`:
```
Objects#toStringHelper: TextSegment{offset=10, length=0}
ToStringBuilder: TextSegment [offset = 10, length = 0]
```
If we would use ToStringBuilder this would not break any tests in core.

svenefftinge:
looks good!

PR #195, opened 2016-12-08
[508894] RegionTrace StackTraces are filled in in debug mode only

cdietrich:
[508894] RegionTrace StackTraces are filled in in debug mode only

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/RegionTrace.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/RegionTrace.java
@@ -15,6 +15,7 @@ import org.eclipse.xtext.formatting2.regionaccess.ITextSegment;
 public class RegionTrace extends RuntimeException {
 
 	private static final long serialVersionUID = -6420580169021980120L;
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=508894

cdietrich:
@meysholdt please have a look

meysholdt:
I'd prefer a solution without hidden configuration flags and without the need to choose between performance and debug information.

cdietrich:
how do you access the stacktrace when you debug?

svenefftinge:
If we turn it around, i.e. that the slow, verbose mode is default, then there is a possibility to optimize without sacrificing debuggability. Would that be ok @meysholdt or do you have a better idea?

cdietrich:
Updated the PullRequest for a `-Dxtext.formatter.debug.skip=true`

svenefftinge:
+1

meysholdt:
-1

yes, a better way would be to first format without debug tracing and in case an error happens automatically format again with debug tracing enabled.

cdietrich:
@meysholdt thus my question: where is the trace programmatically retrieved?!?

meysholdt:
fixed in #231

PR #199, opened 2016-12-09
Don't add spaces before/after ML comments

tivervac:
Prevents spaces from being added on the newline before or after a ML comment

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/formatting2/internal/CommentFormatterTest.xtend
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/formatting2/internal/CommentFormatterTest.xtend
@@ -8,8 +8,10 @@
 package org.eclipse.xtext.formatting2.internal
 
 import com.google.inject.Inject
+import org.eclipse.xtext.formatting2.IFormattableDocument
 import org.eclipse.xtext.formatting2.internal.formattertestlanguage.IDList
 import org.eclipse.xtext.formatting2.internal.tests.FormatterTestLanguageInjectorProvider
+import org.eclipse.xtext.formatting2.regionaccess.ITextRegionExtensions
 import org.eclipse.xtext.testing.InjectWith
 import org.eclipse.xtext.testing.XtextRunner
 import org.junit.Test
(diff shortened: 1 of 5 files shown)


meysholdt:
Hi Titouan, thank you for the contribution. 

Could you add test cases for the scenarios you have in mind? The test org.eclipse.xtext.formatting2.internal.FormattableDocumentTest.simple() may suit as a role model. it allows to specify custom test text as well as custom formatter impls in the same test methods. Since you're dealing with comments you'll probably not need to extend the test grammar.

tivervac:
@meysholdt I can only create such a test if I change [this line](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/WhitespaceReplacer.java#L92) in `createReplacements` such that it also adds spaces.

If you were to replace `newLines + indentation` with `newLines + indentation + space` the behaviour will be noticeable i.e.:
Formatting:
```
/* comment */
class test { }
```
Will then result in:
```
/* comment */
 class test { }
```

Should I still create a test for this? This will have me extending `FormatterTester` to overwrite `createFormatter`. This method will create a `CustomFormatter` where I overwrote `WhitespaceReplacer` to return a `CustomWhitespaceReplacer` which finally changes the one line mentioned above.

meysholdt:
Thanks for the example, but createReplacements() never invokes the code that you modified in this PR.  So I don't think it can serve as an example on how to accidentally insert bogus whitespace before/after multi-line comments. 

I've added tests for the scenarios that I thought you were talking about in 448b0253952b1dd101d0dcdc4e43754c2f4a5168. But I didn't run into bogus whitespace before/after multi-line comments. 

I'm inclined to close this PR. Did I miss something?

tivervac:
The code from the PR is invoked before `createReplacements` is called. But the changes that were configured in `configureWhitespace` will only be transformed into actual replacements in `createReplacements`. The bogus whitespace will only show up when you're using a custom `WhiteSpaceReplacer`. I will have a look at creating a valid testcase later this week.

tivervac:
@meysholdt I edited the test to reflect my problem.

kthoms:
@meysholdt please review

kthoms:
@meysholdt Ping

cdietrich:
@meysholdt  ping

cdietrich:
@tivervac could you please rebase this cr?

szarnekow:
@meysholdt Could you please take a look at the proposed changes?

kthoms:
@meysholdt @tivervac What's the status for this PR? @tivervac please rebase.

tivervac:
@kthoms Rebased.
I've fixed my ECA but I'll probably have to wait for a bit before it updates in Eclipse's internal system.
Not sure why the build's failing? It appears to be complaining about dependencies but I haven't changed those

cdietrich:
@tivervac https://status.bintray.com/

cdietrich:
@tivervac eca still compains can you squash thr two commits locally?

cdietrich:
@tivervac how shall we proceed here?

tivervac:
I still think my ticket is valid, I guess someone can check for obvious mistakes (NPE, typos,...) and believe that the test coverage is sufficient to make sure we didn't break other parts.

I can do the rebase.

tivervac:
replaced by #1543

PR #201, opened 2016-12-13
Xtext language migration to new generator

sailingKieler:
This PR contains the changes in the generated artifacts of the Xtext language implementation after switching to the new Xtext code generator.

* As part of that this PR contributes a new bundle `org.eclipse.xtext.xtext.ide`.
* Furthermore, semantic and syntactic sequencers have been generated for the `Terminals` grammar replacing the old parseTreeConstructor.


spoenemann:
:+1:

PR #206, opened 2016-12-19
[xtext.wizard] minor improvement of runtime project build file templates

sailingKieler:
This PR adds `org.eclipse.xtext.common.types` to the dependencies of MWE executions in 'RuntimeProjectDescriptor's build file templates.
Currently `org.eclipse.xtext.xbase` is added to the compile dependencies of runtime projects by default, but if people decide not including xbase anymore the workflow execution doesn't work anymore.


kthoms:
Whic options in the new project wizard does the user have to select in order to experience the problem?

sailingKieler:
There isn't such option in the wizard, but users will check the dependencies in the build files generated by the wizard, afterwards.

sailingKieler:
http://services.typefox.io/open-source/jenkins/job/xtext-core/job/cs-wizardImprovement/3/

PR #208, opened 2016-12-19
[xtext.testing + xtext.wizard] fix/improvement of build config of test projects in 'TestProjectDescriptor'

sailingKieler:
* consolidated version of 'org.junit' with our reference in 'gradle/versions.gradle' (currently 4.12)
* added dependency to 'junit:junit:4.12' with scope 'testCompile' to templates of 'pom.xml' + 'build.gradle', is required as `org.eclipse.xtext.testing`s optional dependency to `junit:junit` is not exposed to its `pom.xml` created by gradle

### DIFF
--- a/releng/releng-target/xtext-core.target.target
+++ b/releng/releng-target/xtext-core.target.target
@@ -9,5 +9,9 @@
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="false" type="InstallableUnit">
 			<repository location="http://download.eclipse.org/releases/luna/201502271000/"/>
 		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="http://download.eclipse.org/modeling/tmf/xtext/updates/orbit/"/>
+			<unit id="org.junit" version="0.0.0"/>
+		</location>
 	</locations>
 </target>
--- a/gradle/p2-deployment.gradle
+++ b/gradle/p2-deployment.gradle
@@ -17,4 +17,9 @@ p2gen {
 	dependencies {
 		repositoryUrl 'http://download.eclipse.org/releases/luna/201502271000/'
 	}
+	dependencies {
+		repositoryUrl 'http://download.eclipse.org/modeling/tmf/xtext/updates/orbit/'
+		bundle 'org.junit'
+		includeSource true
+	}
 }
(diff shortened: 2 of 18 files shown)


spoenemann:
👍

sailingKieler:
in order to get the build green (http://services.typefox.io/open-source/jenkins/job/xtext-core/job/cs-wizardImprovement2/4/) I had to add `org.junit:4.12.0` to the target platform of local p2 repo build, see second commit. It's available on the orbit.

Are there any objection against that? I assume we'll have to do that at other places, too

cdietrich:
Which Version is in oomph Target Platform?

sailingKieler:
Who knows anything about that? @kthoms?

kthoms:
The target platform contains 4.11 and 4.12. The 4.11 version is declared explicitly, I think we should change this to 4.12 then.

screenshot 6
screenshot 7

sailingKieler:
And where is version fetched from? An orbit US?

kthoms:
from http://download.eclipse.org/modeling/tmf/xtext/updates/orbit/

sailingKieler:
Is there any way of figuring out what is contained by the offered feature?

kthoms:
Which feature do you mean?

sailingKieler:
I just opened that US in eclipse and saw a feature. Or does that p2 repo contain further bundles beyond those contained in that feature? Is there any list like a category.xml?

kthoms:
Please ask @dhuebner about this repo. The Oomph setup references it for its target platform since quite a while. I was not involved into that.

LorenzoBettini:
@sailingKieler that update site is just a composite update site that refers to the Orbit update site (have a look at http://download.eclipse.org/modeling/tmf/xtext/updates/orbit/compositeContent.xml)

sailingKieler:
Thanks @LorenzoBettini! I tried the same but didn't remember the correct spelling of the xml file

sailingKieler:
OK the build works with the US on http://download.eclipse.org/modeling/tmf/xtext/updates/orbit/
(which in fact redirects to an orbit US) and the single imported unit 'org.junit-0.0.0',
see http://services.typefox.io/open-source/jenkins/job/xtext-core/job/cs-wizardImprovement2/5/

If there're no objections I would like to merge that tomorrow.

releng/releng-target/xtext-core.target.target
```diff
@@ -9,5 +9,9 @@
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="false" type="InstallableUnit">
 			<repository location="http://download.eclipse.org/releases/luna/201502271000/"/>
 		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="false" type="InstallableUnit">
+			<repository location="http://download.eclipse.org/modeling/tmf/xtext/updates/orbit/"/>
+			<unit id="org.junit" version="0.0.0"/>
+		</location>
```
spoenemann: If you edit this file directly, the change will be overwritten the next time the p2 build is generated. You should add the dependency in [p2-deployment.gradle](https://github.com/eclipse/xtext-core/blob/master/gradle/p2-deployment.gradle) instead.

sailingKieler:
@spoenemann feel free to approve.

Let's see how long I'm able to keep this in my mind...

Issue #57, opened 2016-07-27
[LanguageServer] Support the case that rootPath is null


spoenemann:
When VS Code opens a file without opening a directory first, the language server is initialized with rootPath = null. This case should also be supported by the Xtext server (we would have a single resource then).

cdietrich:
what happens if vscode is opened with a file?

svenefftinge:
When rootPath is null, we don't index and start with an empty index and resourceset.
Simply waiting for didOpen notification

spoenemann:
@svenefftinge is this already done?

cdietrich:
imho this is still open

PR #209, opened 2016-12-21
[xtext.ide] added further checks on the language server's rootPath,

sailingKieler:
and  simplified calculation of the rootURI

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
@@ -10,6 +10,8 @@ package org.eclipse.xtext.ide.server
 import com.google.common.collect.LinkedListMultimap
 import com.google.common.collect.Multimap
 import com.google.inject.Inject
+import java.nio.file.Files
+import java.nio.file.Paths
 import java.util.Collections
 import java.util.List
 import java.util.Map
@@ -110,6 +112,13 @@ import org.eclipse.xtext.validation.Issue
 		if (params.rootPath === null) {
 			throw new IllegalArgumentException("Bad initialization request. rootPath must not be null.")
 		}
+		val rootPath = Paths.get(params.rootPath)
+		if (!Files.exists(rootPath) || !Files.isDirectory(rootPath)) {
+			throw new IllegalArgumentException("Bad initialization request. rootPath does not exist or is not a directory.")
+		}
+		if (!Files.isWritable(rootPath)) {
+			throw new IllegalArgumentException("Bad initialization request. rootPath is not writable.")
+		}
 		if (languagesRegistry.extensionToFactoryMap.isEmpty) {
 			throw new IllegalStateException("No Xtext languages have been registered. Please make sure you have added the languages's setup class in '/META-INF/services/org.eclipse.xtext.ISetup'")
 		}
@@ -134,7 +143,7 @@ import org.eclipse.xtext.validation.Issue
 		]
 
 		requestManager.runWrite [ cancelIndicator |
-			val rootURI = URI.createFileURI(params.rootPath).toPath.toUri
+			val rootURI = URI.createURI(rootPath.toUri.toString)
 			workspaceManager.initialize(rootURI, [this.publishDiagnostics($0, $1)], cancelIndicator)
 			return null
 		]
--- a/org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/server/LanguageServerImpl.java
+++ b/org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/server/LanguageServerImpl.java
@@ -12,6 +12,10 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Multimap;
 import com.google.inject.Inject;
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -23,7 +27,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import org.apache.log4j.Logger;
-import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.lsp4j.CodeActionParams;
 import org.eclipse.lsp4j.CodeLens;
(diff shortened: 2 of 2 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -134,7 +143,7 @@ import org.eclipse.xtext.validation.Issue
 		]
 
 		requestManager.runWrite [ cancelIndicator |
-			val rootURI = URI.createFileURI(params.rootPath).toPath.toUri
+			val rootURI = URI.createURI(rootPath.toUri.toString)
```
akosyakov: `rootURI` should be `'file://' + rootPath` where `rootPart` is an absolute path, if it so then looks good to me

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -110,6 +112,13 @@ import org.eclipse.xtext.validation.Issue
 		if (params.rootPath === null) {
 			throw new IllegalArgumentException("Bad initialization request. rootPath must not be null.")
 		}
+		val rootPath = Paths.get(params.rootPath)
+		if (!Files.exists(rootPath) || !Files.isDirectory(rootPath)) {
+			throw new IllegalArgumentException("Bad initialization request. rootPath does not exist or is not a directory.")
+		}
+		if (!Files.isWritable(rootPath)) {
```
akosyakov: not sure that it is necessary, what if one just opens the folder for reading?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -110,6 +112,13 @@ import org.eclipse.xtext.validation.Issue
 		if (params.rootPath === null) {
```
akosyakov: Do we have an issue to support null as `rootPath`? It is a valid argument, the server should operate without the root folder then.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
sailingKieler: You're right - good point.
Maybe I had a slightly wrong idea of the language server.
Is the LS allowed or in charge of creating files? Or is that dann by another component?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
sailingKieler: IMO it does so - I just condensed the repeated 'toPath()' & 'toUri()' calls. I'm convinced that 'Paths.get().toUri()' does the correct thing. We just need to be aware that 'Path.get()' may behave slightly different on different platforms (I remember a situation where my code worked well on my Mac but didn't on a Windows system)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
akosyakov: It can change, e.g. generate files, but not necessary, depends on the language.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
akosyakov: It would be nice to have tests for it if there are not yet. Could you try to break it and see whether some tests break :)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
sailingKieler: @svenefftinge what do think about that?

svenefftinge: Yes, for sure. The LS doesn't write to the source folder anyway.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
kittaakos: #57

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -110,6 +112,13 @@ import org.eclipse.xtext.validation.Issue
...
 			throw new IllegalArgumentException("Bad initialization request. rootPath must not be null.")
 		}
+		val rootPath = Paths.get(params.rootPath)
+		if (!Files.exists(rootPath) || !Files.isDirectory(rootPath)) {
+			throw new IllegalArgumentException("Bad initialization request. rootPath does not exist or is not a directory.")
```
kittaakos: Could we make two separate checks instead? I think it would be nice to see what the exact problem was.
So something like this:
```
if (!Files.exists(rootPath)) {
    throw new IAE('does not exist');
}
if (!Files.isDirectory(rootPath)) {
    throw new IAE('not a directory');
}
```

sailingKieler: Sure. Would you take care on that?
Meanwhile I'm too far away from that topic. You can just force push the branch.

kittaakos: I am happy to take care of it. Please keep in mind; I should do it on a fork.

kthoms:
@kittaakos when do you plan to work on the issue, or can we close the PR for now? If you would fork the branch, we could close the branch here also.

kittaakos:
@kthoms, sorry for my late response, I was completely off. Well, I can make those minor changes if that matters. Otherwise, it is your call whether you close this PR or not. Just in case you wait for my changes, let me know, I can apply them this week. Cheers!

kthoms:
It is just that this PR is open for quite a while and you wanted to finish it. I would like to get this into 2.14.

kittaakos:
With the above changes, I have 66 test failures. For instance in `ReferenceTest`. The failing tests neither assume nor require existing path based on `InitializeParams#getRootUri`. Does this PR still valid at all?

kthoms:
In two days this PR is celebrating anniversary. 
@sailingKieler @kittaakos Close?

kittaakos:
Well, I have already added my remarks. Besides not adding much value, this PR breaks the tests. The ball is in your court.

kthoms:
Right. But tests have to be fixed before this can be further reviewed. That's not my part.

cdietrich:
@sailingKieler @kittaakos any news on this?

kittaakos:
Not from my side; I have added my initial comments. Then I added a few more remarks based on the failing tests. Please close this if you want.

svenefftinge:
I think we can close it.

PR #231, opened 2017-01-10
[formatter/performance] Don't create stack traces

meysholdt:
This change:
- introduces a flag that by default disables creation of stack traces
- lets the formatter do a second pass on conflict with that flag enabled

This combines the benefits of both worlds:
- performance without the need for tweaking
- debug information without the need to enable some magic config

see 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=508894
https://bugs.eclipse.org/bugs/show_bug.cgi?id=507174
https://github.com/eclipse/xtext-core/pull/195


org.eclipse.xtext/src/org/eclipse/xtext/formatting2/AbstractFormatter2.java
```diff
@@ -275,6 +280,17 @@ public ITextReplacer createWhitespaceReplacer(ITextSegment hiddens, IHiddenRegio
 		}
 	}
 
+	protected IFormattableDocument handleTraceMissing(IFormattableDocument problematic, RegionTraceMissingException e) {
```
cdietrich: No @since ?

PR #232, opened 2017-01-10
230: Junit4Fragment2 added dependencies for .tests project

### LINKED ISSUE #230: Junit4Fragment2 added dependencies for .tests project
The Xtext project wizard in Xtext 2.11, when it creates a new .tests project automatically adds these deps

```
 org.eclipse.xtext.testing,
 org.eclipse.xtext.xbase.testing
```

The Junit4Fragment2 only makes sure to add `org.eclipse.xtext.testing` but not the other one, while it should also add `org.eclipse.xtext.xbase.testing` so that, when run on a project created with a previous version of Xtext, it will make sure the .tests project have up to date dependencies

LorenzoBettini:
Task-Url: https://github.com/eclipse/xtext-core/issues/230

Closes #230

### DIFF
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/Junit4Fragment2.xtend
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/Junit4Fragment2.xtend
@@ -27,11 +27,19 @@ class Junit4Fragment2 extends AbstractStubGeneratingFragment {
 			"org.eclipse.xtext.testing"
 	}
 	
+	def protected getXbaseTestingPackage() {
+		if (useDeprecatedClasses)
+			"org.eclipse.xtext.xbase.junit"
+		else
+			"org.eclipse.xtext.xbase.testing"
+	}
+	
 	override generate() {
 		if (projectConfig.runtimeTest.manifest !== null) {
 			projectConfig.runtimeTest.manifest => [
 				requiredBundles.addAll(
 					testingPackage,
+					xbaseTestingPackage,
 					"org.eclipse.xtext.xbase.lib"
 				)
 				exportedPackages.add(grammar.runtimeTestBasePackage+";x-internal=true")
@@ -41,6 +49,7 @@ class Junit4Fragment2 extends AbstractStubGeneratingFragment {
 			projectConfig.eclipsePluginTest.manifest => [
 				requiredBundles.addAll(
 					testingPackage,
+					xbaseTestingPackage,
 					"org.eclipse.core.runtime",
 					"org.eclipse.ui.workbench;resolution:=optional"
 				)
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/Junit4Fragment2.xtend
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/Junit4Fragment2.xtend
@@ -20,6 +20,9 @@ class Junit4Fragment2 extends AbstractStubGeneratingFragment {
 	@Accessors(PUBLIC_SETTER)
 	boolean useDeprecatedClasses
 	
+	@Accessors(PUBLIC_SETTER)
+	boolean skipXbaseTestingPackage
+	
 	def protected getTestingPackage() {
 		if (useDeprecatedClasses)
 			"org.eclipse.xtext.junit4"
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/Junit4Fragment2.xtend
```diff
@@ -27,11 +27,19 @@ class Junit4Fragment2 extends AbstractStubGeneratingFragment {
...
+		else
+			"org.eclipse.xtext.xbase.testing"
+	}
+	
 	override generate() {
 		if (projectConfig.runtimeTest.manifest !== null) {
 			projectConfig.runtimeTest.manifest => [
 				requiredBundles.addAll(
 					testingPackage,
+					xbaseTestingPackage,
```
spoenemann: Shouldn't we check `XbaseUsageDetector.inheritsXbase` before adding the dependency? If someone wants it anyway, she can add it manually.

LorenzoBettini: Actually this dependency is added by the Xtext wizard by default in tests projects when it creates the projects initially, that's why I was adding that independently.

spoenemann: Yes, but the difference is that this code will add the dependency even if the user has removed it before, so there is no way to use the fragment without having that dependency. The project generator of the wizard, on the other hand, is a one-time generator.

LorenzoBettini: I see. Actually some classes in this bundle, mainly CompilationTestHelper, are valuable also in projects that do not use Xbase, just for testing the code generator (at least, that's the path I follow also in my book, for testing generated code of a DSL not based on Xbase).

@cdietrich was proposing to add a flag to skip the generation of this dependency, https://github.com/eclipse/xtext-core/issues/230#issuecomment-271631841, and, at least for me, I would prefer this flag in the fragment. Something like skipXbaseTestingPackage (default to false).

Summarizing, I'd like to have this dependency by default :) but then I leave the final decision to you.

spoenemann: What's the point in having an extra flag just for adding a dependency? You could as well simply add the dependency yourself, that's more transparent.

LorenzoBettini: The flag would allow you to SKIP such dep, not to add it.

The dep would be added by default, mimicking the default deps of .tests project created by the Xtext wizard. If the mwe2 adds the dep by default, as I was proposing, then mwe2 will bring an existing .tests project (created with Xtext 2.10) to the same situation of .tests project freshly created by the Xtext wizard.

All in all, my proposal to add this dep by default when running mwe2 aims at easier porting existing projects (Xtext 2.10, not necessarily based on Xbase) to Xtext 2.11.

spoenemann: Ok.

LorenzoBettini: @spoenemann OK as it is? Or should I add the above mentioned flag to skip the generation of dependency?

spoenemann: Ok to have a flag to skip it.

LorenzoBettini: @spoenemann I added the flag and tested that manually

spoenemann: So it's no problem to add an empty string to the requiredBundles? If that's ok, 👍

LorenzoBettini: No, that's not a problem.

PR #235, opened 2017-01-11
479196: AntlrToolFacade always downloads antlr-generator in Maven build

LorenzoBettini:
Task-Url: https://bugs.eclipse.org/bugs/show_bug.cgi?id=479196


LorenzoBettini:
@dhuebner please review

spoenemann:
It would be good to have the same dependency in the Gradle build: `RuntimeProjectDescriptor.buildGradle()` &rarr; `dependencies`

LorenzoBettini:
@spoenemann Since I'm not very familiar with gradle, this is the right way to do that right?

```gradle
mwe2 "org.eclipse.xtext:xtext-antlr-generator:[2.1.1, 3.0)"
```

spoenemann:
Yes, but please use `[2.1.1,3)` as version. Having `3.0` or `3.0.0` as upper bound might yield unexpected results, like accepting `3.0.0-SNAPSHOT`. I think the logic behind it is `3` < `3.0.0-SNAPSHOT` < `3.0.0`

LorenzoBettini:
@spoenemann Done! I guess that is valid also for Maven version range so I updated them both.

spoenemann:
👍

PR #237, opened 2017-01-11
[language server / incrmental builder] allow reusing resource set

svenefftinge:


svenefftinge:
This is for post 2.11

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ProjectManager.xtend
```diff
@@ -105,17 +105,26 @@ class ProjectManager {
...
+        } else {
+            val resDescs = ChunkedResourceDescriptions.findInEmfObject(this.resourceSet);
+            resDescs.setContainer(projectDescription.name, newIndex)   
+        }
+        return this.resourceSet;
     }
     
     def Resource getResource(URI uri) {
-        resourceSet.getResource(uri, true)
+        val resource = resourceSet.getResource(uri, true)
```
akosyakov: maybe we should pass `false` here and check whether the result is null or not?

svenefftinge: Loading it should be ok. We just need to make sure that it is not done in parallel.
I.e. we must not allow parallel read access if we are not sure that everything has been fully initialized.

spoenemann:
There are failing tests:
http://services.typefox.io/open-source/jenkins/job/xtext-core/job/master/320/

svenefftinge:
I am looking into it

Issue #241, opened 2017-01-12
[LSP] ServerLauncher validates messages


cdietrich:
Messages from VSCode are validated in ServerLauncher.

The messages LSP4J Expects and the ones VSCode dont fit.
See https://github.com/eclipse/lsp4j/issues/50

maybe we can disable validation locally

or introduce a param to the main method so that i can decide from outside

cdietrich:
workaround for user: create own serverlauncher class that has validation false.
maybe we can make this parametrizable via arg to main method

cdietrich:
the bug in LSP4J is fixed. so what is left to decide if we want to introduce a arg for that. might be decided after 2.11

svenefftinge:
We should have 0.1.1 release of LSP4J also I'm fine with additional optional arguments for the main method

spoenemann:
I'm closing this as it's not Xtext related. We'll include the LSP4J fix in the Xtext 2.11.0 release.

PR #243, opened 2017-01-16
Switched to LSP4J 0.1.0 release

spoenemann:
The Xtext 2.11.0 release needs a fixed LSP4J version to depend on.

### DIFF
--- a/gradle/upstream-repositories.gradle
+++ b/gradle/upstream-repositories.gradle
@@ -12,7 +12,6 @@ def jenkinsPipelineRepo = { jobName -> "http://services.typefox.io/open-source/j
 repositories {
 	jcenter()
 	if (findProperty('useJenkinsSnapshots') == 'true') {
-		maven { url jenkinsPipelineRepo('lsp4j') }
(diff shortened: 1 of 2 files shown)


cdietrich:
is that release already built or is it still wip (see e.g. the fixes done today)?

spoenemann:
The release build is already available since a week or so.

cdietrich:
but that means we will deliver these bugs with the release so that xtext 2.11 is unusable with vscode?!?

cdietrich:
see https://github.com/eclipse/xtext-core/issues/241

spoenemann:
We'll do LSP4J 0.1.1

svenefftinge:
@cdietrich could you be specific about which of the bugs are really a show stopper? It is strange as it seem to work for the JDT language server and also our applications work nicely (they are using monaco and not vscode directly, though)

svenefftinge:
I just checked. We are not using the message validator. Maybe JDT LS doesn't do that as well.

svenefftinge:
so let's do a 0.1.1 release of LSP4J this week

PR #248, opened 2017-01-16
Fixed #238:PolymorphicDispatcher#compare violates contr. of Comparators

meysholdt:
This change implements the same strategy for PolymorphicDispatch which
is already in place for Xtend's 'dispatch' methods:

see org.eclipse.xtend.core.jvmmodel.DispatchHelper.compare(...)
see https://github.com/eclipse/xtext-core/issues/238

Signed-off-by: Moritz Eysholdt <moritz.eysholdt@typefox.io>

PR #249, opened 2017-01-16
Fixed #238:PolymorphicDispatcher#compare violates contr. of Comparators

### LINKED ISSUE #238: PolymorphicDispatcher#compare violates the contract for Java Comparators
Sometimes, we get the following exception when our Xsemantics-generated type system is initialized:

```
1) Error injecting constructor, java.lang.IllegalArgumentException: Comparison method violates its general contract!
  at eu.numberfour.n4js.typesystem.CustomInternalTypeSystem.<init>(Unknown Source)
  at eu.numberfour.n4js.typesystem.CustomInternalTypeSystem.class(Unknown Source)
  while locating eu.numberfour.n4js.typesystem.CustomInternalTypeSystem
  while locating eu.numberfour.n4js.xsemantics.InternalTypeSystem
    for field at eu.numberfour.n4js.typesystem.N4JSTypeSystem.ts_internal(Unknown Source)
  at org.eclipse.xtext.service.MethodBasedModule.configure(MethodBasedModule.java:57)
  while locating eu.numberfour.n4js.typesystem.N4JSTypeSystem
    for field at eu.numberfour.n4js.validation.AbstractN4JSDeclarativeValidator.ts(Unknown Source)
  at org.eclipse.xtext.service.MethodBasedModule.configure(MethodBasedModule.java:57)
  while locating eu.numberfour.n4js.validation.validators.N4JSProjectSetupValidator
Caused by: java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.util.TimSort.mergeHi(TimSort.java:899)
	at java.util.TimSort.mergeAt(TimSort.java:516)
	at java.util.TimSort.mergeForceCollapse(TimSort.java:457)
	at java.util.TimSort.sort(TimSort.java:254)
	at java.util.Arrays.sort(Arrays.java:1512)
	at java.util.ArrayList.sort(ArrayList.java:1454)
	at java.util.Collections.sort(Collections.java:175)
	at org.eclipse.xtext.util.PolymorphicDispatcher.getDeclaredMethodsOrderedBySpecificParameterType(PolymorphicDispatcher.java:353)
	at org.eclipse.xtext.util.PolymorphicDispatcher.<init>(PolymorphicDispatcher.java:167)
	at org.eclipse.xtext.util.PolymorphicDispatcher.<init>(PolymorphicDispatcher.java:160)
	at it.xsemantics.runtime.XsemanticsRuntimeSystem$2.<init>(XsemanticsRuntimeSystem.java:86)
	at it.xsemantics.runtime.XsemanticsRuntimeSystem.buildPolymorphicDispatcher1(XsemanticsRuntimeSystem.java:86)
	at eu.numberfour.n4js.xsemantics.InternalTypeSystem.init(InternalTypeSystem.java:535)
	at eu.numberfour.n4js.xsemantics.InternalTypeSystem.<init>(InternalTypeSystem.java:531)
	at eu.numberfour.n4js.typesystem.CustomInternalTypeSystem.<init>(CustomInternalTypeSystem.java:39)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at com.google.inject.internal.DefaultConstructionProxyFactory$1.newInstance(DefaultConstructionProxyFactory.java:85)
	at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:85)
	at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:254)
	at com.google.inject.internal.ProviderToInternalFactoryAdapter$1.call(ProviderToInternalFactoryAdapter.java:46)
	at com.google.inject.internal.InjectorImpl.callInContext(InjectorImpl.java:1031)
	at com.google.inject.internal.ProviderToInternalFactoryAdapter.get(ProviderToInternalFactoryAdapter.java:40)
	at com.google.inject.Scopes$1$1.get(Scopes.java:65)
	at com.google.inject.internal.InternalFactoryToProviderAdapter.get(InternalFactoryToProviderAdapter.java:40)
	at com.google.inject.internal.FactoryProxy.get(FactoryProxy.java:54)
	at com.google.inject.internal.SingleFieldInjector.inject(SingleFieldInjector.java:53)
	at com.google.inject.internal.MembersInjectorImpl.injectMembers(MembersInjectorImpl.java:110)
	at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:94)
	at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:254)
	at com.google.inject.internal.ProviderToInternalFactoryAdapter$1.call(ProviderToInternalFactoryAdapter.java:46)
	at com.google.inject.internal.InjectorImpl.callInContext(InjectorImpl.java:1031)
	at com.google.inject.internal.ProviderToInternalFactoryAdapter.get(ProviderToInternalFactoryAdapter.java:40)
	at com.google.inject.Scopes$1$1.get(Scopes.java:65)
	at com.google.inject.internal.InternalFactoryToProviderAdapter.get(InternalFactoryToProviderAdapter.java:40)
	at com.google.inject.internal.SingleFieldInjector.inject(SingleFieldInjector.java:53)
	at com.google.inject.internal.MembersInjectorImpl.injectMembers(MembersInjectorImpl.java:110)
	at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:94)
	at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:254)
	at com.google.inject.internal.ProviderToInternalFactoryAdapter$1.call(ProviderToInternalFactoryAdapter.java:46)
	at com.google.inject.internal.InjectorImpl.callInContext(InjectorImpl.java:1031)
	at com.google.inject.internal.ProviderToInternalFactoryAdapter.get(ProviderToInternalFactoryAdapter.java:40)
	at com.google.inject.Scopes$1$1.get(Scopes.java:65)
	at com.google.inject.internal.InternalFactoryToProviderAdapter.get(InternalFactoryToProviderAdapter.java:40)
	at com.google.inject.internal.InternalInjectorCreator$1.call(InternalInjectorCreator.java:204)
	at com.google.inject.internal.InternalInjectorCreator$1.call(InternalInjectorCreator.java:198)
	at com.google.inject.internal.InjectorImpl.callInContext(InjectorImpl.java:1024)
	at com.google.inject.internal.InternalInjectorCreator.loadEagerSingletons(InternalInjectorCreator.java:198)
	at com.google.inject.internal.InternalInjectorCreator.injectDynamically(InternalInjectorCreator.java:179)
	at com.google.inject.internal.InternalInjectorCreator.build(InternalInjectorCreator.java:109)
	at com.google.inject.Guice.createInjector(Guice.java:95)
	at com.google.inject.Guice.createInjector(Guice.java:72)
	at com.google.inject.Guice.createInjector(Guice.java:62)
	at eu.numberfour.n4js.N4JSInjectorProvider$1.createInjector(N4JSInjectorProvider.java:89)
	at eu.numberfour.n4js.N4JSStandaloneSetup.createInjectorAndDoEMFRegistration(N4JSStandaloneSetup.java:80)
	at eu.numberfour.n4js.N4JSInjectorProvider.internalCreateInjector(N4JSInjectorProvider.java:91)
	at eu.numberfour.n4js.N4JSInjectorProvider.getInjector(N4JSInjectorProvider.java:76)
	at eu.numberfour.n4js.N4JSInjectorProvider.setupRegistry(N4JSInjectorProvider.java:101)
	at org.eclipse.xtext.junit4.XtextRunner.methodBlock(XtextRunner.java:43)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.createRequestAndRun(JUnitCoreWrapper.java:113)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.executeEager(JUnitCoreWrapper.java:85)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.execute(JUnitCoreWrapper.java:54)
	at org.apache.maven.surefire.junitcore.JUnitCoreProvider.invoke(JUnitCoreProvider.java:134)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:200)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:153)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:103)
```

Upon further investigation, it turns out that Xtext's `PolymorphicDispatcher`, upon initialization, tries to sort all dispatch methods using a compare method that violates the general contract for Java Comparators. That contract specifies that a compare method should define a total order, i.e., it should be antisymmetric, transitive, and total.

However, there are pairs of methods `(a,b)` where the last requirement is violated. Such methods are incomparable to each other, that is, neither is `a <= b` nor `a <= b`. The compare method returns `0` for such cases, thereby indicating that `a == b`, which is obviously wrong. See the following example:

```
public class PolymorphicDispatcherTest extends PolymorphicDispatcher<Object> {

	public PolymorphicDispatcherTest() {
		super(Collections.singletonList(new Object()), method -> false);
	}

	private Method findMethod(Class<?> current, String name, Class<?>... parameterTypes) {
		if (current == Object.class)
			throw new IllegalArgumentException("Cannot find method");
		try {
			return current.getDeclaredMethod(name, parameterTypes);
		} catch (java.lang.NoSuchMethodException | SecurityException e) {
			return findMethod(current.getSuperclass(), name, parameterTypes);
		}
	}

        private static class S {}

	private static class A extends C {
	}

	private static class B extends S {
	}

	private static class C extends S {
	}

	public void test(A a) {
	}

	public void test(B b) {
	}

	public void test(C c) {
	}

	public void showTransitivityViolation() {
		MethodDesc m1 = createMethodDesc(this, findMethod(getClass(), "test", A.class));
		MethodDesc m2 = createMethodDesc(this, findMethod(getClass(), "test", B.class));
		MethodDesc m3 = createMethodDesc(this, findMethod(getClass(), "test", C.class));

		int m1m2 = PolymorphicDispatcherTest.this.compare(m1, m2); // 0
		int m2m3 = PolymorphicDispatcherTest.this.compare(m2, m3); // 0
		int m1m3 = PolymorphicDispatcherTest.this.compare(m1, m3); // 1

		if (m1m2 <= 0 && m2m3 <= 0 && !(m1m3 <= 0))
			throw new IllegalStateException("Transitivity violation");
	}

	public static void main(String[] args) {
		PolymorphicDispatcherTest test = new PolymorphicDispatcherTest();
		test.showTransitivityViolation();
	}
}
```

In this example, the method returns `0` when comparing method pairs `m1,m2` and `m2,m3`, even though in both cases, the methods are not identical! Comparing `m1` and `m3` then yields `1`, which violates transitivity.

We think the problem is due to the fact that the subtype relation, on which the compare method is based, only defines a partial order. The compare method attempts to extend this to a total order in an incorrect way.

This exception is thrown only sometimes, and it has popped up only recently (to our knowledge). We think that this is due to the fact that the initial order of the methods to be sorted depends on the initial order of the methods returned by `Class#getDeclaredMethods`, which is undefined by Java. Furthermore, Java uses different sorting algorithms depending on the number of elements to be sorted. In our particular cases, we are trying to sort 59 methods.

meysholdt:
based on @LorenzoBettini suggestions in
https://github.com/eclipse/xtext-core/pull/248#issuecomment-272873308

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/PolymorphicDispatcher.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/PolymorphicDispatcher.java
@@ -353,7 +353,23 @@ public class PolymorphicDispatcher<RT> {
 		Collections.sort(cachedDescriptors, new Comparator<MethodDesc>() {
 			@Override
 			public int compare(MethodDesc o1, MethodDesc o2) {
-				return PolymorphicDispatcher.this.compare(o1, o2);
+				int compare = PolymorphicDispatcher.this.compare(o1, o2);
+				if (compare != 0) {
+					return compare;
+				}
+				Class<?>[] p1 = o1.getParameterTypes();
+				Class<?>[] p2 = o2.getParameterTypes();
+				int to = Math.min(p1.length, p2.length);
+
+				for (int i = 0; i < to; i++) {
+					final String n1 = p1[i].getName();
+					final String n2 = p2[i].getName();
+					compare = n1.compareTo(n2);
+					if (compare != 0) {
+						return compare;
+					}
+				}
+				return compare;
 			}
 		});
 		return cachedDescriptors;


org.eclipse.xtext.util/src/org/eclipse/xtext/util/PolymorphicDispatcher.java
```diff
@@ -353,7 +353,23 @@ protected RT handleAmbigousMethods(List<MethodDesc> result, Object... params) {
...
+
+				for (int i = 0; i < to; i++) {
+					final String n1 = p1[i].getName();
+					final String n2 = p2[i].getName();
+					compare = n1.compareTo(n2);
+					if (compare != 0) {
+						return compare;
+					}
+				}
+				return compare;
```
LorenzoBettini: It looks fine.
But I was wondering whether this is better/more efficient than simply comparing the method desc string representation.
Just a curiosity

meysholdt: :)
I see some problems with using the output of `toString()` as a basis for comparisons
* If `toString()` contains the output of `System#identityHashCode()`, then it's not deterministic
* `toString()` computes the full string, even though only the first `char`s may be needed for the comparison
* `toString()` involves string concatenation, which usually is not very fast, and which is not needed at all to make a comparison

LorenzoBettini: OK, thanks for the clarification!

LorenzoBettini: We should merge shouldn't we?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/PolymorphicDispatcher.java
```diff
@@ -353,7 +353,23 @@ protected RT handleAmbigousMethods(List<MethodDesc> result, Object... params) {
...
 			@Override
 			public int compare(MethodDesc o1, MethodDesc o2) {
-				return PolymorphicDispatcher.this.compare(o1, o2);
+				int compare = PolymorphicDispatcher.this.compare(o1, o2);
+				if (compare != 0) {
+					return compare;
+				}
+				Class<?>[] p1 = o1.getParameterTypes();
+				Class<?>[] p2 = o2.getParameterTypes();
+				int to = Math.min(p1.length, p2.length);
```
kduske-n4: You're probably just playing it safe, but doesn't `compare == 0` imply `p1.length == p2.length`?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/PolymorphicDispatcher.java
```diff
@@ -353,7 +353,23 @@ protected RT handleAmbigousMethods(List<MethodDesc> result, Object... params) {
...
+
+				for (int i = 0; i < to; i++) {
+					final String n1 = p1[i].getName();
+					final String n2 = p2[i].getName();
+					compare = n1.compareTo(n2);
```
kduske-n4: Comparing the reversed string might perform better because FQNs are more likely to have a common prefix (package name), esp. in the case of dispatch methods on class hierarchies. To avoid actually creating reversed strings, you'd need a custom String comparison method, but that's trivial.

That's in the case that performance actually plays a role here, which it probably doesn't.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/PolymorphicDispatcher.java
LorenzoBettini: I think you're right!

org.eclipse.xtext.util/src/org/eclipse/xtext/util/PolymorphicDispatcher.java
kduske-n4: This solution looks good to me. Will it be merged in time for the next release?

spoenemann: > Will it be merged in time for the next release?

Not for 2.11.0, we're already in the testing phase for release candidates there. But probably 2.11.1 won't be much later.

LorenzoBettini:
Unfortunately, this implementation does not solve the issue reported by @kduske-n4 : I provided a patched polymorphic dispatcher according to @meysholdt changes in Xsemantics (https://github.com/LorenzoBettini/xsemantics/blob/f447456929654bdcd3db8ef1b697578dc04f2b1b/plugins/it.xsemantics.runtime/src/it/xsemantics/runtime/internal/PatchedPolymorphicDispatcher.java) but @kduske-n4  reported that their problem is still there. Indeed, we (@kduske-n4 and I) think the patched comparator strategy used while sorting still does not guarantee a total ordering; it avoids returning 0 when the two elements are not equal, but I don't think it guarantees transitivity when two method descriptions are not related. In the end, I don't think we can make the new sorting algorithm happy since when you have method descriptions and Java types you can't come up with a total ordering anyway, since some types are not related.

I then tried with another patched implementation still in the context of Xsemantics https://github.com/LorenzoBettini/xsemantics/blob/3c64109f1795cdb6c2b599e62d416109ef89f5c1/plugins/it.xsemantics.runtime/src/it/xsemantics/runtime/internal/PatchedPolymorphicDispatcher.java. This alternative implementation still uses @meysholdt compare method, but it uses a TreeSet for storing method descriptions (important part is here: https://github.com/LorenzoBettini/xsemantics/blob/3c64109f1795cdb6c2b599e62d416109ef89f5c1/plugins/it.xsemantics.runtime/src/it/xsemantics/runtime/internal/PatchedPolymorphicDispatcher.java#L169). A TreeSet does not have strong assumptions on the ordering as long as it is consistent with equals, and the new @meysholdt compare strategy that was used for sorting is consistent with equals. This version of the polymorphic dispatcher in Xsemantics is currently making @kduske-n4 jobs green again.

This should not break the behavior of the polymorphic dispatcher since the ordering is kept as it used to be for method descriptions that are actually related (some further tests also in this project https://github.com/LorenzoBettini/polymorphic.dispatcher.experiments).

kduske-n4:
Hi everyone, [here](https://gist.github.com/kduske-n4/4abfeee5e25412fb77da8bf2e99892ab) is a test program that I wrote to diagnose the issue. Unfortunately it requires types from N4JS, but @meysholdt should be able to run it easily. It contains 58 methods with signatures taken from our type system, and attempts to sort these methods using different techniques and comparators. 

Since the problem depends on the initial order of the sequence to be sorted, it creates randomly sorted sequences in an endless loop until the problem occurs. For the first three methods, this happens within a second or so. For the fourth method, which is @LorenzoBettini's TreeSet plus his new compare method, it appears to work.

meysholdt:
fixed in https://github.com/eclipse/xtext-core/commit/923422737ed632756d4670dcfb7b079ecc5b5252

PR #258, opened 2017-01-25
Change import of AssumptionViolatedException

kthoms:
With JUnit 4.12 the class was made public API, the old class is
deprecated.

### DIFF
--- a/org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/smoketest/internal/AbstractParallelScenarioRunner.java
+++ b/org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/smoketest/internal/AbstractParallelScenarioRunner.java
@@ -21,7 +21,7 @@ import java.util.concurrent.TimeUnit;
 import org.eclipse.xtext.testing.IInjectorProvider;
 import org.eclipse.xtext.testing.IRegistryConfigurator;
 import org.eclipse.xtext.testing.smoketest.ScenarioProcessor;
-import org.junit.internal.AssumptionViolatedException;
+import org.junit.AssumptionViolatedException;
 import org.junit.runner.Description;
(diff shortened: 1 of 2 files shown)


LorenzoBettini:
JUnit 4.12 is now the minimal version required, isn't it?  Everywhere I mean.

kthoms:
Actually searching for "junit" in MANIFEST.MF files reveals that we are not consistent here. Many package imports have version 4.5.0 or 4.7.0 as lower bound.

Bundle `org.eclipse.xtext.testing`, however, defines 4.12 as lower bound:
https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext.testing/META-INF/MANIFEST.MF
Since the changed classes are from that bundle I think we are safe here. But it might be a further discussion which version of JUnit we require in general.

LorenzoBettini:
Yes, I think we're safe, but I agree that we should be consistent

kthoms:
@LorenzoBettini Ok to merge?

LorenzoBettini:
I'm worried about possible "import packages" as opposed to "required bundles"... what happens for "import packages"? @dhuebner ?

dhuebner:
@LorenzoBettini  Equinox will pick some available version if there are any version constraints defined.
FYI Luna only contains junit 4.11

kthoms:
As said above, the bundle already defines a lower bound of 4.12. Thus it won't be usable with Luna, at least without Orbit. Is it a bug that 4.12 is minimum requirement for xtext.testing? If not, this patch is IMHO valid.

cdietrich:
i assume that 4.12 is there on purpose @spoenemann do you know?

spoenemann:
I thought it's a good idea to start with the latest junit version if we create a new testing bundle.

cdietrich:
so if testing uses this version anyway there should be no problem i think

PR #278, opened 2017-02-14
Introduce Nullable Annotation

cdietrich:
Introduce Nullable Annotation to allow Null-Injections with google guice. see https://github.com/eclipse/xtext-eclipse/issues/153 for details


cdietrich:
@spoenemann @kthoms is the javadoc ok?

Issue #280, opened 2017-02-15
UnicodeTest fails in System.encoding != UTF-8


cdietrich:
UnicodeTest fails in System.encoding != UTF-8

i propose to run the tests with UTF-8 encoding

```
test {
	exclude 'org.eclipse.xtext.serializer.contextFinderTest.NestedTypeRecursiveTest.class'
	exclude '**/*Suite.class'
	systemProperty "file.encoding", "UTF-8"
}
``` 

@all what do you think?

PR #281, opened 2017-02-16
run test with UTF-8 always #280

cdietrich:
run test with UTF-8 always #280

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/parser/terminalrules/UnicodeTest.java
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/parser/terminalrules/UnicodeTest.java
@@ -8,14 +8,24 @@
 package org.eclipse.xtext.parser.terminalrules;
 
 import java.io.ByteArrayOutputStream;
+import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


cdietrich:
please review

spoenemann:
I'd prefer to fix the single failing test instead of manipulating a system property. Why does the test depend on that property?

cdietrich:
you can see the test fail when starting the test with
`-Dfile.encoding=ANSI_X3.4-1968`
which is the default encoding on the workstations i try to build xtext on

spoenemann:
`AbstractXtextTests.doGetResource(InputStream, URI)` might also be a problem. The default implementation of `IEncodingProvider.Runtime` uses `Charset.defaultCharset()`, so the test might have to set a different default encoding there.

cdietrich:
ill have a look

cdietrich:
save is affected as well

spoenemann:
Another thought: if it's too hard to fix the tests directly, we could add something like this to `java-compiler-settings.gradle`:
```
tasks.withType(Test) {
	systemProperty 'file.encoding', 'ISO-8859-1'
}
```
But a more direct fix is certainly favorable.

cdietrich:
fixed problems locally

cdietrich:
@spoenemann is it ok to merge now?

Issue #284, opened 2017-02-22
[Wizard] Include .xtext file in jar created by gradle


cdietrich:
[Wizard] Include .xtext file in jar created by gradle

to be able to reference one grammar from another one we need to package the grammar

cdietrich:
current generation

```
sourceSets.all {
    resources.exclude '**/*.g', '**/*.xtext', '**/*.mwe2', '**/*.xtend', '**/*._trace'
}
```

cdietrich:
same problem with non plain layout

cdietrich:
fixed in 2.11.1

PR #285, opened 2017-02-22
[Wizard] Include .xtext file in jar created by gradle #284

cdietrich:
[Wizard] Include .xtext file in jar created by gradle #284


cdietrich:
please review

PR #286, opened 2017-02-22
[Wizard] Include .xtext file in jar created by gradle #284

cdietrich:

### DIFF
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.gradle/org.xtext.example.gradle.parent/gradle/source-layout.gradle
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.gradle/org.xtext.example.gradle.parent/gradle/source-layout.gradle
@@ -15,6 +15,9 @@ jar {
 	from('model') {
 		into('model')
 	}
+	from(sourceSets.main.allSource) {
+		include '**/*.xtext'
+	}
 	manifest {
 		attributes 'Bundle-SymbolicName': project.name
 	}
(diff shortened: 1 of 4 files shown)


cdietrich:
please review

org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.gradle/org.xtext.example.gradle.parent/gradle/source-layout.gradle
```diff
@@ -15,6 +15,9 @@ jar {
 	from('model') {
 		into('model')
 	}
+	from(sourceSets.main.allSource) {
+		include '**/*.xtext'
+	}
```
spoenemann: Why is this necessary? If you include the xtext files in the resources, they'll be added to the jar automatically.

cdietrich: yes but in maven layout we have src/main/java with xtext+mwe2 and src/main/resources with nothing

spoenemann: Then it would be more correct to move the xtext file to src/main/resources. But ok, let's keep your solution so we don't change the current behavior too much.

cdietrich: yes i consider this as intended (xtext in stc main java)

cdietrich: is it ok to merge?

spoenemann: 👍

Issue #287, opened 2017-02-24
[generator] Support Tracing in Code Generators


svenefftinge:
We have a bit of infrastructure for tracing in core, but some important pieces are still burried in xbase. I'd like to add support for tracing for any language (not only xbase), by adding the missing pieces to core.

cdietrich:
related to https://bugs.eclipse.org/bugs/show_bug.cgi?id=474469 https://bugs.eclipse.org/bugs/show_bug.cgi?id=471147 https://bugs.eclipse.org/bugs/show_bug.cgi?id=448853

PR #288, opened 2017-02-25
[code gen] Added support for tracing in code generation (#287)

svenefftinge:


svenefftinge:
WIP

svenefftinge:
Here is some rough documentation about the approach: https://github.com/TypeFox/tracing_codegen/blob/master/README.md

cdietrich:
will this give general support for import managers as well?

svenefftinge:
@cdietrich yes, you could for instance introduce a special "ImportSectionNode" and add it to the place you want to have it. Then while creating the generator node tree, you would pass along an import manager that adds qualified names to the "ImportSectionNode".

svenefftinge:
Since one creates a tree first and then produce the actual string in a second pass (GeneratorNodeProcessor) allows for any number of passes over the tree before further processing. Also one can subclass GeneratorNodeProcessor and add handling for custom nodes.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingSugar.xtend
```diff
@@ -0,0 +1,79 @@
...
+	def void generateTracedFile(IFileSystemAccess2 fsa, String path, EObject rootTrace, StringConcatenationClient code) {
+		val node = trace(rootTrace, code)
+		generateTracedFile(fsa, path, node);
+	}
+	
+	/**
+	 * Use to generate a file based on generator node.
+	 */
+	def void generateTracedFile(IFileSystemAccess2 fsa, String path, CompositeGeneratorNode rootNode) {
+		val proc = new GeneratorNodeProcessor()
```
meysholdt: use DI instead of `new`

org.eclipse.xtext/xtend-gen/org/eclipse/xtext/generator/trace/node/WhiteSpaceConfig.java
```diff
@@ -0,0 +1,22 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.xtext.generator.trace.node;
+
+/**
+ * @author Sven Efftinge - Initial contribution and API
+ */
+@SuppressWarnings("all")
+public class WhiteSpaceConfig {
```
meysholdt: Let's use a name that includes to word `Generator` or similar to avoid confusion with `org.eclipse.xtext.formatting.IWhitespaceInformationProvider`
`org.eclipse.xtext.formatting.IIndentationInformation` and 
`org.eclipse.xtext.formatting.ILineSeparatorInformation`

meysholdt: e.g. `GeneratorWhitespaceConfig` or `OutputWhitespaceConfig`

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingExtensions.xtend
```diff
@@ -0,0 +1,96 @@
...
+		val annotationType = findTypeGlobally(TracingExtensions)
+		val factories = annotatedClass.findAnnotation(annotationType)?.getClassArrayValue("value")
+		if (factories === null) {
+			return;
+		}
+		for (f : factories.map[type].filter(InterfaceDeclaration)) {
+			for (t: f.declaredMethods.filter[simpleName.startsWith('create') && parameters.empty].map[returnType]) {
+				for (getter : t.allResolvedMethods.filter[ isGetter]) {
+					val rt = getter.resolvedReturnType
+					if (allowedLowerCaseTypeNames.contains(rt.type.simpleName.toLowerCase)) {
```
meysholdt: If I understand it correctly, `allowedLowerCaseTypeNames` is expected to match types with values that can be used in the generated output without additional conversion. So I assume that `float` is not contained because when turing a float into text it's necessary to decide about it's format.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingExtensions.xtend
```diff
@@ -0,0 +1,96 @@
...
+							addParameter('target', t)
+							body = '''
+								�EStructuralFeature� feature = target.eClass().getEStructuralFeature("�getter.featureName�");
+								�ILocationData� location = this.location(target, feature, -1);
+								�CompositeGeneratorNode� trace = this.trace(location);
+								this.append(trace, target.�getter.declaration.simpleName�());
+								return trace;
+							'''
+						]
+					} else if (eobjectType.isAssignableFrom(rt)) {
```
meysholdt: I think it would be nice to have the `getFoo(target, stringProvider)` for any kind of EStructuralFeature-based Getter, even for Strings and floats. They may require some formatting, e.g. turning a string into a getter name or deciding for a float about the number digits after the dot.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingExtensions.xtend
```diff
@@ -0,0 +1,96 @@
...
+		val eobjectType = EObject.newTypeReference()
+		annotatedClass.extendedClass = TracingSugar.newTypeReference()
+		val annotationType = findTypeGlobally(TracingExtensions)
+		val factories = annotatedClass.findAnnotation(annotationType)?.getClassArrayValue("value")
+		if (factories === null) {
+			return;
+		}
+		for (f : factories.map[type].filter(InterfaceDeclaration)) {
+			for (t: f.declaredMethods.filter[simpleName.startsWith('create') && parameters.empty].map[returnType]) {
+				for (getter : t.allResolvedMethods.filter[ isGetter]) {
```
meysholdt: There can be Getters that don't have EStructuralFeatures, e.g. from EOperations. To identify them exactly, we'd need to load the Ecore (better: GenModel). Not sure if that's possible here.

org.eclipse.xtext/META-INF/MANIFEST.MF
```diff
@@ -34,6 +34,7 @@ Export-Package: org.eclipse.xtext,
  org.eclipse.xtext.generator,
  org.eclipse.xtext.generator.trace;x-friends:="org.eclipse.xtext.xbase.ui,org.eclipse.xtend.ide",
  org.eclipse.xtext.generator.trace.internal;x-friends:="org.eclipse.xtext.ui,org.eclipse.xtext.xbase.ui",
+ org.eclipse.xtext.generator.trace.node,
```
meysholdt: This should be exported with `;x-internal:=true,` for now.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/GeneratorNodeExtensions.xtend
```diff
@@ -0,0 +1,117 @@
...
+		parent.children += new NewLineNode(wsConfig.lineDelimiter, true)
+		return parent
+	}
+	/**
+	 * Creates a text node containing the toString() representation of the given object and
+	 * appends it to the given parent node.
+	 * 
+	 * @return the given parent node 
+	 */
+	def CompositeGeneratorNode append(CompositeGeneratorNode parent, Object object) {
```
meysholdt: I assume this where it would be nice to hook in external stateful object-to-string converters, for example import manages.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingExtensions.xtend
```diff
@@ -0,0 +1,96 @@
...
+import org.eclipse.xtend.lib.macro.declaration.InterfaceDeclaration
+import org.eclipse.xtend.lib.macro.declaration.MutableClassDeclaration
+import org.eclipse.xtend.lib.macro.declaration.ResolvedMethod
+import org.eclipse.xtext.generator.trace.ILocationData
+
+/**
+ * @author Sven Efftinge - Initial contribution and API
+ */
+@Active(TracingExtensionsProcessor)
+annotation TracingExtensions {
```
meysholdt: Maybe call this `TracedModelAccess`. I'm not in favor of calling this class `*Extension` as it actually is an active annotation. So technically `*ExtensionsProvider` would be correct ;)

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingExtensions.xtend
svenefftinge: It's not. It will bail at runtime, though.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/GeneratorNodeExtensions.xtend
svenefftinge: Yes, I suggest to do that through overriding.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingExtensions.xtend
svenefftinge: ok

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingSugar.xtend
svenefftinge: ok

org.eclipse.xtext/xtend-gen/org/eclipse/xtext/generator/trace/node/WhiteSpaceConfig.java
svenefftinge: ok

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingExtensions.xtend
svenefftinge: I did not contain it, because it is never used in meta models.
But I will change it such, that for all other types we have the variant with the string provider.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/GeneratorNodeExtensions.xtend
meysholdt: This leaves open the question where the import manager's state should live. `GeneratorNodeExtensions` is stateless and `CompositeGeneratorNode` doesn't know its parent so it's not possible to navigate up the GeneratorNodeTree until some ImportSectionNode is found.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracingExtensions.xtend
svenefftinge: I called it `TracedAccessors`.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/GeneratorNodeExtensions.xtend
svenefftinge: You could either pass down an ImportManager explicitly, or create ImportNodes during code gen and replace them with the simple names and import section in a second pass.

PR #296, opened 2017-03-03
Copied Flaky into .xtext.testing

LorenzoBettini:
This annotation was still present in .xtext.junit4, so why not copying
that into the new .xtext.testing and then later deprecate the one in
.xtext.junit4?


LorenzoBettini:
please review

spoenemann:
> This annotation was still present in .xtext.junit4, so why not copying that into the new .xtext.testing and then later deprecate the one in .xtext.junit4?

Because Flaky is kind of a workaround for problematic tests. I did not include it in the initial version of xtext.testing because we don't want to encourage users to use workarounds...

PR #299, opened 2017-03-07
Use getValuesProvider function in IssueSeveritiesProvider.

gallandarakhneorg:
The function `getValueProvider` is defined in `IssueSeveritiesProvider` for replying the value provider instance. 
This function seems to be defined as protected in order to be overriden. But is it not called from the rest of the code of `IssueSeveritiesProvider`. This PR fixes this issue.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/validation/IssueSeveritiesProvider.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/validation/IssueSeveritiesProvider.java
@@ -15,6 +15,7 @@ import com.google.inject.Inject;
 
 /**
  * @author Sven Efftinge - Initial contribution and API
+ * @author Stephane Galland - add calls to getValuesProvider()
  * @since 2.4
  */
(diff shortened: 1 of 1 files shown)


cdietrich:
The getter is there so that you can access it from subclasses it is not there to overrideit in subclasses

cdietrich:
Thanks

LorenzoBettini:
@cdietrich this might be an academic discussion, but if it's not meant to be overwritten then it should be final, shouldn't it? Otherwise, it makes sense to call it explicitly

LorenzoBettini:
oh, but I see you merged it now :)

cdietrich:
Yes it is ;)

cdietrich:
And yes it is academic

gallandarakhneorg:
Just for clarify my usage for further discussions. I have defined a specific IssueSeveritiesProvider, which enables to override the severities programmatically without changing the preferences or any storage behind the provider. For doing, I provides a specific value provider in my subclass.

Thanks for the merging :-)

PR #305, opened 2017-03-22
[380449] Enhancements to Task Finder

cdietrich:
[380449] Enhancements to Task Finder

- Introduced a hook to allow stripping of the node text before calling the task parser
- Stripped trailing comment in tasks

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/tasks/DefaultTaskFinderTest.xtend
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/tasks/DefaultTaskFinderTest.xtend
@@ -47,6 +47,7 @@ class DefaultTaskFinderTest extends AbstractXtextTests {
 				 * Fixme no match
 				 * FOO also no match
 				 */
+				/* TODO Get rid of this */
 				Hello notATODO!
 			''')
 		)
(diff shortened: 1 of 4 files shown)


cdietrich:
please review

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=380449

svenefftinge:
It would be better to reuse something like 'org.eclipse.xtext.documentation.impl.AbstractMultiLineCommentProvider.getTextFromMultilineComment(String)' to identify any terminals to strip off.

cdietrich:
well that one is non api so i cannot use it

svenefftinge:
I think adding 
```
        public String getDocumetation(String rawToken) {
		return getTextFromMultilineComment(rawToken);
	}
```
would be good.

cdietrich:
hmmmm, and directly inject the  provider? at other places IEObjectDocumentationProvider(Extension) is used

svenefftinge:
No that doesn't sound good. 
Let's instead than just reuse `endTag` from `AbstractMultiLineCommentProvider.MultiLineCommentProviderProperties` with the same default as in `AbstractMultiLineCommentProvider`

cdietrich:
wont work, MultiLineCommentProviderProperties is private

cdietrich:
@svenefftinge can you please have a look again?

PR #306, opened 2017-03-22
[493990] Fixed Duplicate Problem in TraceRegionMerger

cdietrich:

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/TraceRegion.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/TraceRegion.java
@@ -7,6 +7,7 @@
  *******************************************************************************/
 package org.eclipse.xtext.generator.trace;
 
+import java.util.Collection;
 import java.util.List;
 
 import org.eclipse.xtext.util.TextRegionWithLineInformation;
@@ -45,7 +46,15 @@ public class TraceRegion extends AbstractStatefulTraceRegion {
  		}
 	}
 	
+	/**
+	 * @deprecated use {link {@link #TraceRegion(int, int, int, int, boolean, Collection, AbstractTraceRegion)} instead.
+	 */
+	@Deprecated
 	public TraceRegion(int myOffset, int myLength, int myLineNumber, int myEndLineNumber, boolean useForDebugging, List<ILocationData> allLocationData, AbstractTraceRegion parent) {
+		this(myOffset, myLength, myLineNumber, myEndLineNumber, useForDebugging, (Collection<ILocationData>) allLocationData, parent);
+	}
+	
+	public TraceRegion(int myOffset, int myLength, int myLineNumber, int myEndLineNumber, boolean useForDebugging, Collection<ILocationData> allLocationData, AbstractTraceRegion parent) {
 		super(new TextRegionWithLineInformation(myOffset, myLength, myLineNumber, myEndLineNumber), useForDebugging, Lists.newArrayList(allLocationData), parent);
 		if (parent == null) {
 			for(ILocationData locationData: allLocationData) {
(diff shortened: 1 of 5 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=493990

cdietrich:
please review

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/TraceRegionMerger.java
```diff
@@ -177,7 +179,7 @@ private AbstractTraceRegion replaceTruncated(int index, AbstractTraceRegion regi
 	}
 
 	private AbstractTraceRegion replaceMerged(int index, AbstractTraceRegion region, List<ILocationData> locationData) {
-		List<ILocationData> mergedLocations = newArrayList(region.getAssociatedLocations());
+		Set<ILocationData> mergedLocations = newLinkedHashSet(region.getAssociatedLocations());
 		mergedLocations.addAll(locationData);
 		AbstractTraceRegion merged = new TraceRegion(
 				region.getMyOffset(), region.getMyLength(),
```
svenefftinge: Could you turn it into a list here already? Then we don't need to change the signature of TraceRegion.

cdietrich: ok ill change that.

cdietrich: if it is ok that we then do the conversion twice - i dont know what is better)

cdietrich: if i understand the javadoc at TraceRegion correct it in internal use only so changing it is no problem

svenefftinge: True, but it has been around a bit and I know projects that use it. So I'd like to avoid breaking it if there is no good reason. OTOH copying twice is indeed not a good idea (I am unsure if the copying makes sense at all). If you prefer your proposed solution, could you add an overloaded constructor with the old signature and deprecate it?

cdietrich: yes i will

cdietrich: done

cdietrich: @svenefftinge  can you please have a look again

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/TraceRegion.java
```diff
@@ -45,7 +46,15 @@ public TraceRegion(int myOffset, int myLength, int myLineNumber, int myEndLineNu
...
 	
+	/**
+	 * @deprecated use {link {@link #TraceRegion(int, int, int, int, boolean, Collection, AbstractTraceRegion)} instead.
+	 */
+	@Deprecated
 	public TraceRegion(int myOffset, int myLength, int myLineNumber, int myEndLineNumber, boolean useForDebugging, List<ILocationData> allLocationData, AbstractTraceRegion parent) {
+		this(myOffset, myLength, myLineNumber, myEndLineNumber, useForDebugging, (Collection<ILocationData>) allLocationData, parent);
+	}
+	
+	public TraceRegion(int myOffset, int myLength, int myLineNumber, int myEndLineNumber, boolean useForDebugging, Collection<ILocationData> allLocationData, AbstractTraceRegion parent) {
```
svenefftinge: The collection type is overly constrained. We only need the upperbound, which will make List<ILocationData> directly assignable. So all the casts are not neccessry (except. the one above because of overload resolution).
I.e. `Collection<? extends ILocationData>`

Issue #219, opened 2017-01-09
NPE in PartialParsingHelper


dhakehurst:
I'm experiencing a NPE in class PartialParsingHelper#87:
```
ReplaceRegion replaceRegion = tokenRegionProvider.getTokenReplaceRegion(insertChangeIntoReplaceRegion(oldRootNode, changedRegion), changedRegion);
```
The field tokenRegionProvider is injected optional, and it is null in my scenarion. There has to be a null guard.

cdietrich:
@dhakehurst how did you manage it to provide null as TokenRegionProvider

cdietrich:
fixed in Xtext 2.12

PR #319, opened 2017-04-12
NPE in PartialParsingHelper #219

cdietrich:
NPE in PartialParsingHelper #219


cdietrich:
please review.
i am still wondering how to test this with a normal language/injection.
since xtext ususally uses JIT bindings there is always a TokenRegionProvider available and injecting null is not allowed since the member is not @Nullable

PR #326, opened 2017-04-26
Allow postprocessing on generator result tree

spoenemann:
It can be very useful to post-process the resulting tree of the generator. This change makes all generator node types mutable.

Also changed `equals()`, `hashCode()`, and `toString()` to be consistent for all node types.


spoenemann:
Removed.

PR #328, opened 2017-04-28
Fixed inconsistencies between TemplateNode and Xtend templates

spoenemann:


org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/GeneratorNodeProcessor.xtend
```diff
@@ -66,59 +66,60 @@ class GeneratorNodeProcessor {
...
-		}
-		try {
-			ctx.currentIndents.push(node.indentationString)
-			ctx.pendingIndent = true
-			doProcessChildren(node, ctx)
-		} finally {
-			ctx.currentIndents.pop
+		if (node.indentImmediately && !ctx.pendingIndent) {
+			ctx.currentLine.append(node.indentationString)
+		}
```
svenefftinge: so it appends an indent even if there are no children?

spoenemann: I did that to avoid too many special cases. Probably it makes sense to check whether there is at least one leaf node. I'll change this.

PR #329, opened 2017-05-01
handle null argument for projectConfig

kthoms:
refactored method to return null when precondition is not satisfied to remove one indentation layer
removed null guard for projectConfig later in the method


cdietrich:
are all the other methods null safe regarding project config?

kthoms:
All methods calling getTraceToTarget are safe to handle null as return value. asFile is safe, since it is not called with the result of getProjectConfig(sourceFile).

cdietrich:
i still have the feeling we fix downstream errors that are caused somewhat upstream.
thus id be interested how the user gets into this situation

kthoms:
The effect is just that no trace will be available for some situation we cannot reproduce. I don't know much enough about the tracing stuff.

cdietrich:
me neither. thus i am unsure if we should go this way

svenefftinge:
I agree with @cdietrich, that we should rather handle null pointers on the call site than making the parameters nullable.

kthoms:
OK, will raise another PR to also assure that the method is not called with null parameter.

PR #330, opened 2017-05-02
[515846] Don't call getTraceToTarget when src project config unavailable

kthoms:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=515846

cdietrich:
@kthoms is this pr obsolte by your other changes?

kthoms:
Hmm, getProjectConfig might return null, and getTraceToTarget handles the null argument. Seems that this is not really neccessary, maybe just more explicit. What do you think? Change or drop?

cdietrich:
you introduced another pr for the null check. please check the history

cdietrich:
i consider it as obsolete and thus will close

Issue #251, opened 2017-01-16
Formatter erroneously applies default formatting


meysholdt:
Since 6bbad4f14219993811310efebbb828321953217d the formatter filters out identity edits. 
This, however, causes org.eclipse.xtext.formatting2.AbstractFormatter2.postProcess()
to misbehave: It can't distinguish between filtered identity-edits and missing edits and therefore it applies default formatting for all of them. Default formatting is to apply a single whitespace.

meysholdt:
This only occurs for regions for which org.eclipse.xtext.formatting2.regionaccess.IHiddenRegion.isUndefined() is true, i.e. during serialization.

PR #331, opened 2017-05-04
Do not allow identity edits, even for undefined regions

tivervac:
Tell me if I'm missing something, but I do not think we're doing any harm by disallowing identity edits (if the option was checked) in the general case, even for undefined regions.

Signed-off-by: Titouan Vervack <titouan.vervack@sigasi.com>

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/TextReplacerContext.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/TextReplacerContext.java
@@ -228,16 +228,10 @@ public class TextReplacerContext implements ITextReplacerContext {
 			request.getExceptionHandler().accept(exception);
 			return;
 		}
-		if (!isInRequestedRange(replacement)) {
+		if (!isInRequestedRange(replacement) ||
+			(!request.allowIdentityEdits() && isIdentityEdit(replacement)) ||
+			(!isInUndefinedRegion(replacement) && request.isFormatUndefinedHiddenRegionsOnly())) {
 			return;
-		}
-		if (!isInUndefinedRegion(replacement)) {
-			if (request.isFormatUndefinedHiddenRegionsOnly()) {
-				return;
-			}
-			if (!request.allowIdentityEdits() && isIdentityEdit(replacement)) {
-				return;
-			}
 		}
 		try {
 			replacements.add(replacement);


cdietrich:
@tivervac this makes the following test fail

```
org.eclipse.xtext.formatting2.internal.FormatterReplacementsTest > testUndefinedIdentityEditsAreNotFiltered FAILED
    org.junit.ComparisonFailure at FormatterReplacementsTest.java:177
```

tivervac:
@cdietrich I'll have a look at it. Is there an easy way to run all the tests?

cdietrich:
simply calling ./gradlew build on the command line should do the tricj

tivervac:
The test fails because it tests that undefined regions are not filtered, even with the flag enabled. This is exactly what I'm trying to fix with this commit.

Since this functionality is specifically tested, it would be nice to get some feedback as to why it is necessary.

cdietrich:
well it looks like @meysholdt has currently zero time to look at anything. sorry about that. i dont know enhough to comment. @svenefftinge do you know?

meysholdt:
@tivervac could you tell on why you want to filter the identity edits?

Historically, we had this behavior after https://github.com/eclipse/xtext-core/commit/6bbad4f14219993811310efebbb828321953217d and I removed it in https://github.com/eclipse/xtext-core/commit/2012ce725640c0b2e54c995faac54e53a4fe0891

see https://github.com/eclipse/xtext-core/issues/251#issue-201058036 for why this was a problem.

meysholdt:
Conceptually, I find it problematic to allow *identity edits* inside *undefined regions* since *identity edits* replace *old text* with identical *new text* and in *undefined regions* it is undefined what the *old text* actually is.

tivervac:
@meysholdt 
I have some undefined identity edits that are produced as a result of [this](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/HiddenRegionReplacer.java#L115) code.
When we finally reach `postProcess()` I start manipulating the list of replacements directly. To be sure all my replacements stay ordered, I use a TreeSet.

During one of these manipulations, another replacement is added at the same offset as the undefined identity edit. However, a TreeSet doesn't allow two elements with the same key. As a result, my second edit, the only real edit, is ignored and the undefined identity edit is kept.

That being said, there are other ways to fix my problem, but the only way a conflict can occur is if there are identity edits being allowed through the filtering. For this reason and because of the conceptual problem, I came up with this current solution.

meysholdt:
Hi @tivervac,

I'm sorry for the late response. 

It sounds like you might want to use `org.eclipse.xtext.formatting2.internal.TextSegmentSet<T>` instead of `TreeSet`. There you can register a merge strategy to handle conflicts. 

The problem that I have with disallowing identity edits on undefined regions is that those regions are "undefined": It literally is not defined what chars are in this region. They are being defined by having a TextReplacement filling them with text.

Thus, closing this.

PR #336, opened 2017-05-16
server: file log name has whitespace

vladdu:
The whitespaces make it not work on Windows. 

I'm not sure if the suggested solution is the best, so take it just as a suggestion.


kthoms:
could you be so kind to check out your branch, and amend a signed-off commit? the ip-validation check does not work currently when using the GitHub web editor. Sorry for the inconvenience.

vladdu:
Sorry, I didn't think about that.

kthoms:
Thanks for updating the PR.

PR #344, opened 2017-05-19
Fix assertEquals in GrammarConstraintProviderFeatureTest

DelRed2:


spoenemann:
Thanks!

meysholdt:
good catch. Thank you.

Issue #343, opened 2017-05-19
Tests fail on Windows due to windows line separators


DelRed2:
TracingSugarTest::testCodeGeneration
CliWizardIntegrationTest::testProjectCreation

cdietrich:
did a pull request for the wizard part. for the `TracingSugarTest ` part i need some guidance from @svenefftinge : do i see it correct that the trace files generally use unix line endings?

cdietrich:
if yes we could try something like

```
val root = parseHelper.parse(LineDelimiters.toUnix('''
			type String {}
			type Foo {
				String name;
			}
		'''))
		val fsa = new InMemoryFileSystemAccess
		fsa.generateTracedFile('foo/bar.txt', root, '''
			«FOR t : root.types»
				«t._generateType»
			«ENDFOR»
		''')
		val generated = fsa.textFiles.get(IFileSystemAccess.DEFAULT_OUTPUT + 'foo/bar.txt')
		
		// check the generated string is as expected
		Assert.assertEquals(LineDelimiters.toUnix('''
			«FOR t : root.types»
				«t.generateType»
			«ENDFOR»
		'''), generated.toString)
		
		val trace = (generated as ITraceRegionProvider).traceRegion
		Assert.assertEquals(LineDelimiters.toUnix('''
			CompletableTraceRegion [myOffset=0, myLength=55] associations={
			  LocationData [TextRegionWithLineInformation [0:41][lineNumber=0, endLineNumber=3]][path=__synthetic0.lazylinkingtestlanguage]
			} nestedRegions={
			  CompletableTraceRegion [myOffset=0, myLength=17] associations={
			    LocationData [TextRegionWithLineInformation [0:14][lineNumber=0, endLineNumber=0]][path=__synthetic0.lazylinkingtestlanguage]
			  } nestedRegions={
			    CompletableTraceRegion [myOffset=6, myLength=6] associations={
			      LocationData [TextRegionWithLineInformation [5:6][lineNumber=0, endLineNumber=0]][path=__synthetic0.lazylinkingtestlanguage]
			    }
			  }
			  CompletableTraceRegion [myOffset=17, myLength=38] associations={
			    LocationData [TextRegionWithLineInformation [15:26][lineNumber=1, endLineNumber=3]][path=__synthetic0.lazylinkingtestlanguage]
			  } nestedRegions={
			    CompletableTraceRegion [myOffset=23, myLength=3] associations={
[...]

PR #346, opened 2017-05-20
Fixed CliWizardIntegrationTest on windows. #343

cdietrich:
Fixed CliWizardIntegrationTest on windows. #343


cdietrich:
please review: i think forcibly configure the workflow to use unix line endings on both unix like and windows os is the easiest solution for that

cdietrich:
windows is such a pile of crap. test that is green in eclipse is not in cmd line.

PR #350, opened 2017-05-24
dont generate traced accessor for java.lang.Object#getClass

cdietrich:
dont generate traced accessor for java.lang.Object#getClass


org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracedAccessors.xtend
```diff
@@ -89,6 +89,8 @@ class TracedAccessorsProcessor extends AbstractClassProcessor {
 		if (declaration.static)
 			return false
 		val n = it.declaration.simpleName
+		if (n == 'getClass')
+			return false
```
spoenemann: I think it would be better to check whether the declaring type of the method is `Object`.

cdietrich:
do we have an api for that or do i need to check the name?

cdietrich:
@spoenemann updated

PR #358, opened 2017-06-01
[LSP] made aquisition of read lock synchronous to ensure right order of read and write requests. Only reads should go parallel and can be executed in any order.

### LINKED ISSUE #332: RequestManagerTest is flaky
I have the phenomenon that RequestManagerTest failing on different machines,
depending on how executing the test (Eclipse/Gradle/Suite/....) on different machines.

the problem is what the write closure is actually executed before the read closure.

The Systems i see this have in common that they are linux machines and thus use openjdk

svenefftinge:
Fixes #332


cdietrich:
looks good on my machine

svenefftinge:
Anton reminded me, that the reason we did it like before, is that within a write transaction there was client code that would trigger a read request and hence block itself. Need to be investigated further.

PR #361, opened 2017-06-02
better message in parsing test

cdietrich:


cdietrich:
what do you think?

meysholdt:
👍 since checking for syntax error is the natural thing to do for a parser test

PR #363, opened 2017-06-05
Various LS fixes and improvements

akosyakov:


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ServerLauncher.xtend
```diff
@@ -7,61 +7,135 @@
...
+	}
+
+	def static void redirectStandardStreams(OutputStream out, OutputStream err) {
+		redirectStandardStreams(ServerLauncher.silentIn, out, err)
+	}
+
+	def static void redirectStandardStreams(InputStream in, OutputStream out, OutputStream err) {
+		System.setIn(in)
+		System.setOut(new PrintStream(out))
+		System.setErr(new PrintStream(err))
```
svenefftinge: I know it was like that before, but we should not touch System.err as it is not used at all.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
```diff
@@ -163,13 +163,16 @@ import org.eclipse.xtext.workspace.IWorkspaceConfig
...
+        getProjectManager(uri)?.resourceSet.URIConverter.exists(uri, null)
     }
     
     def <T> T doRead(URI uri, (Document, XtextResource)=>T work) {
     	val resourceURI = uri.trimFragment
     	val projectMnr = getProjectManager(resourceURI)
-        val resource = projectMnr.getResource(resourceURI) as XtextResource
+        val resource = projectMnr?.getResource(resourceURI) as XtextResource
+        if (resource === null) {
+        	return work.apply(null, null)
```
svenefftinge: Why do you want to execute the callback, if no resource was found?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/findReferences/WorkspaceResourceAccess.xtend
```diff
@@ -25,7 +25,11 @@ class WorkspaceResourceAccess implements IResourceAccess {
 
 	override <R> readOnly(URI targetURI, IUnitOfWork<R, ResourceSet> work) {
 		return workspaceManager.doRead(targetURI) [ document, resource |
-			return work.exec(resource.resourceSet)
+			val resourceSet = resource?.resourceSet
+			if (resourceSet === null) {
+				return null
```
svenefftinge: If this really happens, it deserves some error logging

svenefftinge: Also what you really seem to check here is `resource === null`. Checking for the resourceSet was a bit missleading to me. Are you really considering a case, where there is a resource but not resourceSet?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ServerLauncher.xtend
akosyakov: I see that uncaught exceptions end up in `System.err`. If we don't redirect it, I think, it will break communication.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
akosyakov: In our project resources can link against a static library that is not described as a project and it is fine as a user should not be able to access resources from the library. But this code fails with an NPE for a project, e.g. during the computation of locations from language services. Instead of changing all services, I prefer to fix it here. I can override and copy this method to our project. But will not it be useful for other projects?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/findReferences/WorkspaceResourceAccess.xtend
akosyakov: We return `null` in all other implementation if `resourceSet` cannot be found, that's why I've put a comment about it the javadoc for `ResourceAccess` and used such guard, e.g. https://github.com/eclipse/xtext-eclipse/blob/242b5fcc83f4697f6381f9264f710773e8435ce8/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/LoadingResourceAccess.java#L46. 

But I think you are right to return `null` only when `resource === null` here and fail in other cases.

akosyakov:
@svenefftinge could you answer please that i know how to proceed

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
svenefftinge: It's ok for me, as it is now. If the resource is null, the document is, too. The situation also is not swallowed.

akosyakov:
@svenefftinge I've addressed your comments, good to merge?

Issue #161, opened 2016-10-21
TypeReference(packageName, className) ERRROR


ELHAMLAOUI:
Hi ,
When I try to launch the generation of Xtext artifacts I get the following error:
`[ERROR]: GeneratorException: (Element: -UNKNOWN-; Reported by: XtextGenerator)
     java.lang.IllegalArgumentException: Cannot determine the package name of 'GS.GSPackage'. Please use the TypeReference(packageName, className) constructor`

cdietrich:
see https://www.eclipse.org/forums/index.php/t/1082021/

svenefftinge:
We should add a proper validation for the disallowed naming.

cdietrich:
i had a look and i do not really see a clean solution for that.
what might work is something like

```
public interface IValidationJobCallback {
	
	public boolean isReadOnly();

}
```

in XtextDocumentProvider

```
protected void registerAnnotationInfoProcessor(ElementInfo info) {
		XtextDocument doc = (XtextDocument) info.fDocument;
		if(info.fModel != null) {
			AnnotationIssueProcessor annotationIssueProcessor = new AnnotationIssueProcessor(doc, info.fModel,
				issueResolutionProvider);
			ValidationJob job = new ValidationJob(resourceValidator, doc, annotationIssueProcessor, CheckMode.FAST_ONLY);
			job.setValidationJobCallback(new IValidationJobCallback() {
				
				@Override
				public boolean isReadOnly() {
					return info.fElement != null && XtextDocumentProvider.this.isReadOnly(info.fElement);
					
				}
			});
			doc.setValidationJob(job);
		}
	}
```

```
	@Override
	protected void disposeElementInfo(Object element, ElementInfo info) {
		if (info.fDocument instanceof XtextDocument) {
			XtextDocument document = (XtextDocument) info.fDocument;
			ValidationJob job = (ValidationJob) document.getValidationJob();
			if (job != null) {
				job.cancel();
				job.setValidationJobCallback(null);
			}
			document.disposeInput();
		}
		super.disposeElementInfo(element, info);
	}
```

in ValidationJob

```
	@Override
	protected IStatus run(final IProgressMonitor monitor) {
		if (monitor.isCanceled())
			return Status.CANCEL_STATUS;
		if (callback !=null && callback.isReadOnly()) 
			return Status.OK_STATUS;
                 .....
```

```
	private IValidationJobCallback callback;
        public void setValidationJobCallback(IValidationJobCallback callback) {
[...]

PR #369, opened 2017-06-29
fix problem with bad naming convention epackages

cdietrich:
fix problem with bad naming convention epackages #161 


cdietrich:
please review

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/TypeReference.xtend
```diff
@@ -177,4 +190,12 @@ class TypeReference {
 		path + ".xtend"
 	}
 	
+	@FinalFieldsConstructor
+	static class QualifiedClazzName {
```
spoenemann: Why `QualifiedClazzName` instead of `QualifiedClassName`?

cdietrich: changed

PR #372, opened 2017-06-30
Create an in-memory workspace configuration when no root URI is given

### LINKED ISSUE #57: [LanguageServer] Support the case that rootPath is null

spoenemann:
Fixes eclipse/xtext-core#57. WorkspaceConfigFactory switches to the new InMemoryProjectConfig when the root URI is null. This allows to work with files without a real workspace, e.g. when using the standalone Monaco web editor.

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/InMemoryWorkspaceTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/InMemoryWorkspaceTest.xtend
@@ -1,0 +1,52 @@
+/*******************************************************************************
+ * Copyright (c) 2017 TypeFox GmbH (http://www.typefox.io) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.server
+
+import org.eclipse.lsp4j.DidOpenTextDocumentParams
+import org.eclipse.lsp4j.Position
+import org.eclipse.lsp4j.TextDocumentIdentifier
+import org.eclipse.lsp4j.TextDocumentItem
+import org.eclipse.lsp4j.TextDocumentPositionParams
+import org.junit.Test
+
+class InMemoryWorkspaceTest extends AbstractTestLangLanguageServerTest {
+	
+	@Test
+	def void testCompletion() {
+		initialize[rootUri = null]
+		val inmemoryUri = 'inmemory:/mydoc.testlang'
+		languageServer.didOpen(new DidOpenTextDocumentParams => [
+			textDocument = new TextDocumentItem => [
+				uri = inmemoryUri
+				text = '''
+					type Foo {}
+				'''
+			]
+		])
+		
+		val completionItems = languageServer.completion(new TextDocumentPositionParams => [
+			textDocument = new TextDocumentIdentifier(inmemoryUri)
+			position = new Position(0, 10)
+		])
+		val result = completionItems.get
+		val items = if (result.isLeft) result.getLeft else result.getRight.items 
+		val actualCompletionItems = items.toExpectation
+		val expectedCompletionItems = '''
+			Foo (TypeDeclaration) -> Foo [[0, 10] .. [0, 10]]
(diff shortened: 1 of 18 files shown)


spoenemann:
Note that I had to add a workaround for eclipse/xtext#2598 in TestLanguageGenerator.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/InMemoryWorkspaceTest.xtend
```diff
@@ -0,0 +1,52 @@
...
+import org.eclipse.lsp4j.TextDocumentItem
+import org.eclipse.lsp4j.TextDocumentPositionParams
+import org.junit.Test
+
+class InMemoryWorkspaceTest extends AbstractTestLangLanguageServerTest {
+	
+	@Test
+	def void testCompletion() {
+		initialize[rootUri = null]
+		val inmemoryUri = 'inmemory:/mydoc.testlang'
```
akosyakov: Is it MUST to use `inmemory` scheme if a root URI is not provided? or it is is just an example

akosyakov: I am asking because VS code send `didOpen` with a file URI if only a single file is opened without the folder

spoenemann: I'll add a test with another scheme.

spoenemann:
@akosyakov I verified that it works regardless of which URI scheme you use in `didOpen`.

akosyakov (commented):
LGTM

akosyakov (approved):

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/MultiProjectWorkspaceConfigFactory.xtend
```diff
@@ -61,9 +66,11 @@ class MultiProjectWorkspaceConfig implements IWorkspaceConfig {
     }
     
     override findProjectContaining(URI member) {
-        name2config.values.filter[
+        val candidates = name2config.values.filter[
             findSourceFolderContaining(member) !== null
-        ].maxBy[path.segmentCount]
+        ]
+        if (!candidates.empty)
+        	return candidates.maxBy[path.segmentCount]
```
svenefftinge: the null case should be explicit: `else return null`

org.eclipse.xtext/src/org/eclipse/xtext/workspace/InMemoryWorkspaceConfig.xtend
```diff
@@ -0,0 +1,66 @@
...
+import static extension org.eclipse.xtext.util.UriUtil.*
+
+@FinalFieldsConstructor
+class InMemoryWorkspaceConfig implements IWorkspaceConfig {
+	
+	val IProjectConfig projectConfig
+	
+	override findProjectByName(String name) {
+		if (projectConfig.name == name)
+			return projectConfig
```
svenefftinge: same here, implicit null returning are a frequent source of bugs

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/generator/TestLanguageGenerator.xtend
```diff
@@ -24,6 +24,10 @@ import org.eclipse.xtext.ide.tests.testlanguage.testLanguage.Property
 class TestLanguageGenerator extends AbstractGenerator {
 
 	override void doGenerate(Resource resource, IFileSystemAccess2 fsa, IGeneratorContext context) {
+		if (fsa.getURI('').scheme == 'inmemory') {
```
svenefftinge: I think we should disable code generation in language server mode in general using a NEVER impl of `org.eclipse.xtext.generator.IShouldGenerate`.

spoenemann: Should be handled in #373.

PR #376, opened 2017-07-04
[ls] cancellable builds

akosyakov:
I've split `didOpen`, `didChange`, `didClose` and `didChangeWatchedFiles` to 2 write requests:
- applying document changes - performed synchronously, cannot be canceled
- building - performed asynchronously, can be canceled

In that way, we preserve the order of document changes and don't lose them, and at the same time are able to cancel builds.

Read requests are handled as before, so they can be canceled only by the client but not by another request. I've tried to cancel them with write requests but it leads to unexpectedly canceled requests on the client side.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/BuildManager.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/BuildManager.xtend
@@ -36,11 +36,15 @@ class BuildManager {
 
     val dirtyFiles = <URI>newLinkedHashSet
     val deletedFiles = <URI>newLinkedHashSet
+    
+    def Buildable submit(List<URI> dirtyFiles, List<URI> deletedFiles) {
+    	queue(this.dirtyFiles, deletedFiles, dirtyFiles)
+        queue(this.deletedFiles, dirtyFiles, deletedFiles)
+        return [cancelIndicator|internalBuild(cancelIndicator)]
+    }
 
     def List<IResourceDescription.Delta> doBuild(List<URI> dirtyFiles, List<URI> deletedFiles, CancelIndicator cancelIndicator) {
-        queue(this.dirtyFiles, deletedFiles, dirtyFiles)
-        queue(this.deletedFiles, dirtyFiles, deletedFiles)
-        return internalBuild(cancelIndicator)
+    	return submit(dirtyFiles, deletedFiles).build(cancelIndicator)
     }
     
     protected def void queue(Set<URI> files, Collection<URI> toRemove, Collection<URI> toAdd) {
@@ -99,5 +103,9 @@ class BuildManager {
     protected static class ProjectBuildData {
         List<URI> dirtyFiles
         List<URI> deletedFiles
+    }
+    
+    static interface Buildable {
+    	def List<IResourceDescription.Delta> build(CancelIndicator cancelIndicator)
     }
 }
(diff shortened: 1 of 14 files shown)


akosyakov:
@svenefftinge please review
@spoenemann could you try with your Monaco example? do you still have issues?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
```diff
@@ -43,73 +65,147 @@ class RequestManager {
...
+	 * </p>
+	 * <p>
+	 * A synchronous read request:
+	 * <ul>
+	 * 	<li>cannot be canceled by another request;</li>
+	 *  <li>does not cancel any request.</li>
+	 * </ul>
+	 * </p>
+	 */
+	def <V> V runReadSync(()=>V readRequest) {
```
svenefftinge: what's the purpose of this? why synchronous?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
```diff
@@ -43,73 +65,147 @@ class RequestManager {
...
-			if (isCancelException(t)) {
-				LOGGER.info("request cancelled.")
-				throw new CancellationException()
-			}
-			throw t
-		} finally {
-			semaphore.release(MAX_PERMITS)
-		}
+		cancel
+		return runAsync(WRITE_PERMITS) [
```
svenefftinge: async locking for write request doesn't work, since it will jeopardize the order of changes.
They are coming in ordered, and we need to keep the. E.g. a read request coming directly after a write request, might be executed before it, running on outdated state.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/BuildManager.xtend
```diff
@@ -36,11 +36,15 @@ class BuildManager {
 
     val dirtyFiles = <URI>newLinkedHashSet
     val deletedFiles = <URI>newLinkedHashSet
+    
+    def Buildable submit(List<URI> dirtyFiles, List<URI> deletedFiles) {
+    	queue(this.dirtyFiles, deletedFiles, dirtyFiles)
+        queue(this.deletedFiles, dirtyFiles, deletedFiles)
+        return [cancelIndicator|internalBuild(cancelIndicator)]
```
svenefftinge: what happens with the changes if the build got canceled? they need to be queued again.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
akosyakov: I use synchronous for buffering document changes and async only for builds

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
akosyakov: just to be symmetric with `runWriteSync`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/BuildManager.xtend
akosyakov: it happens in synchronous and not cancelable block

spoenemann:
@akosyakov I tried with this branch, but that doesn't solve my issue reported at TypeFox/monaco-languageclient#3. Maybe that's because the build is so fast that it's not canceled (only one very small file).

PR #377, opened 2017-07-04
safe retrieval of bundle in EEF

kthoms:
The EEF used to produce the following code:
```
public class MyDslExecutableExtensionFactory extends AbstractGuiceAwareExecutableExtensionFactory {

	@Override
	protected Bundle getBundle() {
		return MydslActivator.getInstance().getBundle();
	}
	
	@Override
	protected Injector getInjector() {
		return MydslActivator.getInstance().getInjector(MydslActivator.ORG_XTEXT_EXAMPLE_MYDSL_MYDSL);
	}
	
}
```

This raises a NPE when the activator was not started and hence getInstance() returns null. This can happen due to configuration issues and happens quite frequently.

This change will produce the following code:
```
	@Override
	protected Bundle getBundle() {
		return Platform.getBundle(MydslActivator.PLUGIN_ID);
	}
	
	@Override
	protected Injector getInjector() {
		MydslActivator activator = MydslActivator.getInstance();
		return activator != null ? activator.getInjector(MydslActivator.ORG_XTEXT_EXAMPLE_MYDSL_MYDSL) : null;
	}
```

Additionally the PLUGIN_ID constant is generated to the Activator class.

The proper handling of the null conditions will be done in a separate PR for AbstractGuiceAwareExecutableExtensionFactory.

### DIFF
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGenerator.xtend
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGenerator.xtend
@@ -307,7 +307,7 @@ class XtextGenerator extends AbstractWorkflowComponent2 {
 	
 	protected def void generateActivator() {
 		if (projectConfig.eclipsePlugin.srcGen !== null && !languageConfigs.empty)
-			templates.createEclipsePluginActivator(languageConfigs).writeTo(projectConfig.eclipsePlugin.srcGen)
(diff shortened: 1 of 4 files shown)


cdietrich:
i am not sure if this a good idea. did you test what happens if you start and xtext plugin with a null injector?

kthoms:
With eclipse/xtext-eclipse#295 this would be handled properly. Anyway, currently there would be a NPE, and without eclipse/xtext-eclipse#295 also. I see both PRs connected.

PR #379, opened 2017-07-05
Throw a response error when workspace is not initialized

### LINKED ISSUE #378: NPE thrown when an LSP extension tries to read a document before initialize
The following exception occurs when an LSP extension calls `ILanguageServerAccess.doRead(..)` before the `initialize` request has arrived. The WorkspaceManager should throw an IllegalStateException instead.
```
java.util.concurrent.CompletionException: java.lang.NullPointerException
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:604)
	at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:577)
	at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:442)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
	at org.eclipse.xtext.ide.server.WorkspaceManager.getProjectManager(WorkspaceManager.java:177)
	at org.eclipse.xtext.ide.server.WorkspaceManager.doRead(WorkspaceManager.java:243)
	at org.eclipse.xtext.ide.server.LanguageServerImpl$1.lambda$doRead$1(LanguageServerImpl.java:760)
	at org.eclipse.xtext.ide.server.concurrent.RequestManager.lambda$runRead$2(RequestManager.java:116)
	at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:602)
	... 5 more
```

spoenemann:
Fixes #378.

Also added missing `void` return types to some WorkspaceManager methods. This changes their signature, since previously `didOpen` and `didClose` returned the build result. However, `didChange` was already `void` before, so I assume the previous implicit return types were not on purpose.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
@@ -16,6 +16,9 @@ import java.util.Map
 import java.util.Set
 import org.eclipse.emf.common.util.URI
 import org.eclipse.lsp4j.TextEdit
+import org.eclipse.lsp4j.jsonrpc.ResponseErrorException
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseError
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode
 import org.eclipse.xtext.ide.server.BuildManager.Buildable
 import org.eclipse.xtext.ide.server.ILanguageServerAccess.IBuildListener
 import org.eclipse.xtext.resource.IExternalContentSupport.IExternalContentProvider
@@ -106,6 +109,13 @@ import org.eclipse.xtext.workspace.IWorkspaceConfig
         val result = buildManager.doInitialBuild(newProjects, cancelIndicator)
         afterBuild(result)
     }
+    
+    protected def checkInitialized() {
+    	if (workspaceConfig === null) {
+    		val error = new ResponseError(ResponseErrorCode.serverNotInitialized, "Workspace has not been initialized yet.", null)
+    		throw new ResponseErrorException(error)
+		}
+    }
 	
 	protected def afterBuild(List<Delta> deltas) {
 		for (listener : buildListeners) {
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
```diff
@@ -103,6 +106,13 @@ import org.eclipse.xtext.workspace.IWorkspaceConfig
         val result = buildManager.doInitialBuild(newProjects, cancelIndicator)
         afterBuild(result)
     }
+    
+    protected def checkInitialized() {
```
akosyakov: maybe better to have a getter that does a check, instead of checking each time

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
```diff
@@ -103,6 +106,13 @@ import org.eclipse.xtext.workspace.IWorkspaceConfig
...
+    	if (workspaceConfig === null) {
```
akosyakov: Should not we set `workspaceConfig` to `null` on shutdown?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
```diff
@@ -104,8 +107,16 @@ import org.eclipse.xtext.ide.server.BuildManager.Buildable
         val result = buildManager.doInitialBuild(newProjects, cancelIndicator)
         afterBuild(result)
     }
+    
+    protected def IWorkspaceConfig getWorkspaceConfig() {
+    	if (workspaceConfig === null) {
```
svenefftinge: Should we rename the field to e.g. `_workspaceConfig`, to make sure it is always accessed through the getter?

PR #390, opened 2017-07-13
[eclipse/xtext-eclipse#299] Added support for Language Server Jar

kthoms:
The IDE project's build can be extended to create an executable Language
Server jar. For Maven based builds the maven-shade-plugin is used and
for Gradle based builds the ShadowJar plugin is used.
Due to a limitation of the maven-shade-plugin in a Maven/Tycho
environment the build must be tweaked a bit to materialize dependencies
with the maven-dependencies-plugin and attached for inclusion with the
addjars-maven-plugin.

Extended the test suite for the different configurations with language
server support:
- plain Maven
- Maven/Tycho
- Gradle


svenefftinge:
uberjars are a bit hacky and problematic, as they don't allow duplicate resources.
https://docs.gradle.org/3.3/userguide/application_plugin.html seems to be a more sustainable approach

kthoms:
Which duplicate resources are you referring to? The ShadowJar plugin is supposed to complement the application plugin:
http://imperceptiblethoughts.com/shadow/#integrating_with_application_plugin

The configuration will use the application and shadowJar plugin.

There is only one duplicate resource to consider I am aware of: plugin.properties. The values of plugin.properties is appended to one file.

Is there something in the resulting uber Jar that does not work as expected?

cdietrich:
the main problem is plugin.properties
and emf then complaining about properties not beeing there

LorenzoBettini:
When I need to create a fatjar with Eclipse bundles with Maven/Tycho I create a jar with jars in it, and use a jar-in-jar loader (the one you get when you export an executable jar from Eclipse). I still haven't documented that in a blog post but you can find an example here: https://github.com/LorenzoBettini/eclipse-fatjar-example/tree/master/org.example.standalone (a similar technique is used in a few examples of the 2nd edition of the Xtext book, but, again, I did not document that in the book). This technique works also with the mentioned problems of EMF (indeed the above example uses EMF just for a demonstration).

I don't know about gradle though, but I think the same technique could be reused.

spoenemann:
I would prefer not to encourage users to create fat jars. For Gradle the application plugin works perfectly. There must be something similar for Maven.

kthoms:
When deploying a language server it is preferrable to have a single jar for it. Users usually don't want to care about the internal dependencies nor build specific stuff. They'll get a single jar that they could embed into their different LS extensions without the need to consider build system specific stuff or manage complex classpaths.

Note that this is opt-in. Users are free to choose or leave it. If you want a different way you are free to add support for an exploded format and let the user choose either one. But as for now I think this is a valuable extension how it is.

spoenemann:
> When deploying a language server it is preferrable to have a single jar for it. Users usually don't want to care about the internal dependencies nor build specific stuff. They'll get a single jar that they could embed into their different LS extensions without the need to consider build system specific stuff or manage complex classpaths.

That's exactly what the Gradle application is for. It creates a start script that takes care of the classpath, so users don't have to find the Java executable and set up the classpath themselves.

kthoms:
I'll give http://www.mojohaus.org/appassembler/appassembler-maven-plugin/plugin-info.html a try, seems to do something similar.

Maybe I could extend it to let the user choose the distribution format. Some might prefer the uberjar, but we could make this the 2nd option.

cdietrich:
LGTM

PR #391, opened 2017-07-14
refreshWorkspaceConfig - Fixed initialization of remainingProjectNames

kthoms:
It is wrong to initialize the remainingProjectNames with the newHashSet
function since it has the side effect that it will put the keySet as
whole as single entry to the new map instead of putting its values in
the map.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
@@ -10,10 +10,13 @@ package org.eclipse.xtext.ide.server
 import com.google.inject.Inject
 import com.google.inject.Provider
 import java.util.ArrayList
+import java.util.HashSet
 import java.util.List
 import java.util.Map
+import java.util.Set
 import org.eclipse.emf.common.util.URI
 import org.eclipse.lsp4j.TextEdit
+import org.eclipse.xtext.ide.server.BuildManager.Buildable
 import org.eclipse.xtext.ide.server.ILanguageServerAccess.IBuildListener
 import org.eclipse.xtext.resource.IExternalContentSupport.IExternalContentProvider
 import org.eclipse.xtext.resource.IResourceDescription
@@ -26,7 +29,6 @@ import org.eclipse.xtext.util.CancelIndicator
 import org.eclipse.xtext.util.internal.Log
 import org.eclipse.xtext.validation.Issue
 import org.eclipse.xtext.workspace.IWorkspaceConfig
-import org.eclipse.xtext.ide.server.BuildManager.Buildable
 
 /**
  * @author Sven Efftinge - Initial contribution and API
@@ -85,7 +87,7 @@ import org.eclipse.xtext.ide.server.BuildManager.Buildable
     protected def refreshWorkspaceConfig(CancelIndicator cancelIndicator) {
         workspaceConfig = workspaceConfigFactory.getWorkspaceConfig(baseDir)
         val newProjects = newArrayList
-        val remainingProjectNames = newHashSet(projectName2ProjectManager.keySet)
+        val Set<String> remainingProjectNames = new HashSet(projectName2ProjectManager.keySet)
         workspaceConfig.projects.forEach [ projectConfig | 
             if(projectName2ProjectManager.containsKey(projectConfig.name)) {
                 remainingProjectNames.remove(projectConfig.name)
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
```diff
@@ -85,7 +86,7 @@ import org.eclipse.xtext.ide.server.BuildManager.Buildable
     protected def refreshWorkspaceConfig(CancelIndicator cancelIndicator) {
         workspaceConfig = workspaceConfigFactory.getWorkspaceConfig(baseDir)
         val newProjects = newArrayList
-        val remainingProjectNames = newHashSet(projectName2ProjectManager.keySet)
+        val remainingProjectNames = new HashSet(projectName2ProjectManager.keySet)
```
cdietrich: maybe <String>newHashSet

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
```diff
@@ -183,7 +183,7 @@ import java.util.HashSet
     }
     
     protected def boolean exists(URI uri) {
-        getProjectManager(uri)?.resourceSet.URIConverter.exists(uri, null)
+        getProjectManager(uri)?.resourceSet?.URIConverter.exists(uri, null)
```
cdietrich: how comes the resourceset beeing null

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
szarnekow: Please make the type of the local var explicit so that subsequent changes do not revert the fix by accident.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
kthoms: when getProjectManager(uri) returns null, resourceSet is null and getURIConverter is invoked on it. See the generated Java code for the potential null access.

kittaakos: Just asking; then the `URIConverter` can be `null` as well, right?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
kthoms: updated

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
kthoms: No, for XtextResourceSet and ResourceSetImpl it is guaranteed to return not null.

org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/server/WorkspaceManager.java
```diff
@@ -247,7 +249,11 @@ protected boolean exists(final URI uri) {
     if (_projectManager!=null) {
       _resourceSet=_projectManager.getResourceSet();
     }
-    return _resourceSet.getURIConverter().exists(uri, null);
+    URIConverter _uRIConverter = null;
+    if (_resourceSet!=null) {
+      _uRIConverter=_resourceSet.getURIConverter();
+    }
+    return _uRIConverter.exists(uri, null);
```
akosyakov: could be an NPE here

kthoms: right. that'll be a problem. update follows.

kthoms:
updated, please review

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
```diff
@@ -182,7 +184,10 @@ import org.eclipse.xtext.ide.server.BuildManager.Buildable
     }
     
     protected def boolean exists(URI uri) {
-        getProjectManager(uri)?.resourceSet.URIConverter.exists(uri, null)
+        val rs = getProjectManager(uri)?.resourceSet
```
cdietrich: simpy put ?. on all members on the chain does not work?

kthoms: would also, but the generated code is smaller this way

PR #392, opened 2017-07-14
Added null guard for result of listFiles

kthoms:
According to contract of File.listFiles() the result can be null if an
I/O error occurred.


szarnekow:
👍

PR #394, opened 2017-07-17
Throw an IllegalStateException if a language server extension tries to read a file before the server is initialized

spoenemann:
The fix for #378 leads to a ResponseErrorException when the workspace is accessed before it is initialized. This is ok for client requests, but for local accesses an IllegalStateException is more appropriate. This PR translates the exception accordingly.


akosyakov:
Here the access is used for `resolveCodeLens`: https://github.com/eclipse/xtext-core/pull/395/files#diff-0fc0eb4aab841adf6e1c198a00c6e95dR455.

spoenemann:
Hmm, you're right. I changed this because I ran into a situation where a request made by an LSP extension could arrive before the initialize request. In this particular case I need to catch the ResponseErrorException, which is a bit strange, so I thought catching an IllegalStateException seems more natural.

PR #395, opened 2017-07-17
[lsp] added hooks for code lenses and language specific capabalities

svenefftinge:


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codelens/ICodeLensResolver.xtend
```diff
@@ -0,0 +1,24 @@
...
+import org.eclipse.xtext.ide.server.ILanguageServerAccess
+import org.eclipse.xtext.util.CancelIndicator
+
+/**
+ * @author Sven Efftinge - Initial contribution and API
+ */
+interface ICodeLensResolver {
+	
+	/**
+	 * Resolve the given code lens, only if it was created by this code lens service.
```
akosyakov: the doc is confusing

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -411,11 +433,33 @@ import org.eclipse.xtext.validation.Issue
...
 	}
 
 	override resolveCodeLens(CodeLens unresolved) {
-		return CompletableFuture.completedFuture(unresolved)
+		return CompletableFutures.computeAsync[ cancelChecker |
+			var CodeLens current = unresolved
+			for (language : this.allLanguages) {
+				val codeLensService = language.get(ICodeLensResolver)
+				if (codeLensService !== null) {
+					current = codeLensService.resolveCodeLens(this.access, current, [
```
akosyakov: Should not we pass a special access that cancels all read requests if the user cancels this future? It looks like the user will end up with 2 cancel indicators at some point and should check both all the time, otherwise, he will miss one or another cancellation.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -411,11 +433,33 @@ import org.eclipse.xtext.validation.Issue
...
+
+			return workspaceManager.doRead(uri) [ document, resource |
+				codeLensService.computeCodeLenses(document, resource, params, cancelIndicator)
+			]
+		]
 	}
 
 	override resolveCodeLens(CodeLens unresolved) {
-		return CompletableFuture.completedFuture(unresolved)
+		return CompletableFutures.computeAsync[ cancelChecker |
```
akosyakov: Is not `doRead` here more appropriate, otherwise, this future won't be canceled by a write request.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
svenefftinge: Not sure I understand. Why would the use get two cancelIndicators? Where does the second come from?
Shouldn't it be fine if I change the locking to `doRead`, as you suggested above?

akosyakov: - the first indicator from `resolveCodeLens` to cancel JSON-RPC request that is passed as a 3rd argument
- the second from `access.doRead` to cancel the read request

With `doRead` it would be the same, it is because `access.doRead` starts a new read request. But with `doRead` there is no need to use `access` one can directly read documents.

spoenemann:
👍

svenefftinge:
I decided to install a URI into the code lens, to call the right implementation with the right resource again on resolve. 

Also, note, that there are no semantic changes in the WorkspaceManager, I just formatted it.

akosyakov:
LGTM

PR #400, opened 2017-07-21
[LSP] Added language hook for ‘codeAction’.

svenefftinge:
Signed-off-by: Sven Efftinge <sven.efftinge@typefox.io>


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
```diff
@@ -65,6 +66,12 @@ import org.eclipse.xtend.lib.annotations.Data
         }
         return new Position(line, column)
     }
+    
+    def String getSubstring(Range range) {
```
akosyakov: maybe just `substring` or `getValue`?

PR #403, opened 2017-07-21
Decode file URL

xcariba:
Decodes URL-encoded file paths.
Original implementation fails on Linux when LSP request contains Cyrillic characters in file name.

Signed-off-by: Alexander Kozinko <xcariba@gmail.com>

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
@@ -7,43 +7,51 @@
  *******************************************************************************/
 package org.eclipse.xtext.ide.tests.server
 
+import javax.inject.Inject
+import org.eclipse.emf.common.util.URI
+import org.eclipse.xtext.ide.server.UriExtensions
+import org.eclipse.xtext.ide.tests.testlanguage.TestLanguageIdeInjectorProvider
+import org.eclipse.xtext.testing.InjectWith
+import org.eclipse.xtext.testing.XtextRunner
 import org.junit.Test
+import org.junit.runner.RunWith
 
 import static org.junit.Assert.*
+import java.io.File
 
-class UriExtensionsTest extends AbstractTestLangLanguageServerTest {
+@RunWith(XtextRunner)
+@InjectWith(TestLanguageIdeInjectorProvider)
+class UriExtensionsTest {
+
+	@Inject
+	extension UriExtensions;
+
 	@Test
-	def void testFilesWithSpaces() {
-		initialize
-
-		val fileURI = 'Foo Bar.testlang'.virtualFile
-		fileURI.open('''
-			type FooBar {
-			}
-		''')
-
-		diagnostics.get(fileURI).forEach [
-			println("err: " + it.message)
-		]
-
-		assertTrue(diagnostics.get(fileURI).empty)
+	def void testConversion() {
+		assertEquals("file:///dir/name.ext", "file:///dir/name.ext".toUri.toPath)
 	}
 
 	@Test
-	def void testFilesWithCyrillic() {
-		initialize
+	def void testFileUriConversion() {
+		assertEquals("file:///dir/name.ext", URI.createFileURI("/dir/name.ext").toPath)
+	}
 
-		val fileURI = "\u0424\u0443 \u0411\u0430\u0440.testlang".virtualFile
-		fileURI.open('''
-			type FooBar {
-			}
-		''')
+	@Test
+	def void testFilesWithSpaces() {
+		assertEquals("file:///dir/Foo Bar.testlang", "file:///dir/Foo Bar.testlang".toUri.toPath)
+	}
 
-		diagnostics.get(fileURI).forEach [
-			println("err: " + it.message)
-		]
+	@Test
+	def void testFilesWithCyrillicSymbols() {
+		assertEquals("file:///dir/\u0424\u0443 \u0411\u0430\u0440.testlang",
+			"file:///dir/\u0424\u0443 \u0411\u0430\u0440.testlang".toUri.toPath)
+	}
 
-		assertTrue(diagnostics.get(fileURI).empty)
+	@Test
+	def void testFolderIsPrefix() {
+		var directory = new File("./test-data/test-project")
+		var uri = URI.createFileURI(directory.absolutePath).toPath.toUri
+		assertTrue(uri.isPrefix)
 	}
 }
 
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
@@ -1,0 +1,49 @@
+/*******************************************************************************
+ * Copyright (c) 2017 TypeFox GmbH (http://www.typefox.io) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.server
+
+import org.junit.Test
+
+import static org.junit.Assert.*
+
+class UriExtensionsTest extends AbstractTestLangLanguageServerTest {
+	@Test
+	def void testFilesWithSpaces() {
+		initialize
+
+		val fileURI = 'Foo Bar.testlang'.virtualFile
+		fileURI.open('''
+			type FooBar {
+			}
+		''')
+
+		diagnostics.get(fileURI).forEach [
+			println("err: " + it.message)
+		]
+
+		assertTrue(diagnostics.get(fileURI).empty)
+	}
+
+	@Test
+	def void testFilesWithCyrillic() {
+		initialize
+
+		val fileURI = "\u0424\u0443 \u0411\u0430\u0440.testlang".virtualFile
+		fileURI.open('''
+			type FooBar {
+			}
+		''')
+
+		diagnostics.get(fileURI).forEach [
+			println("err: " + it.message)
+		]
+
+		assertTrue(diagnostics.get(fileURI).empty)
+	}
+}
+
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
@@ -28,7 +28,7 @@ class UriExtensions {
 	}
 
 	def String toPath(URI uri) {
-		return uri.toString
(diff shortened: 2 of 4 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
```diff
@@ -31,8 +33,8 @@ class UriExtensions {
 
 	def String toPath(java.net.URI uri) {
 		try {
-			val path = Paths.get(uri)
-			return path.toUri.toString
+			val path = Paths.get(uri.path)
+			return URLDecoder.decode(path.toUri.toString, "UTF-8");
```
kittaakos: It is just a minor thing but what about using the `StandardCharsets.UTF_8.name` instead of the `"UTF-8"` string?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
```diff
@@ -20,21 +22,31 @@ import org.eclipse.emf.common.util.URI
...
 	def URI toUri(String pathWithScheme) {
-		val javaNetUri = java.net.URI.create(pathWithScheme);
-		val path = javaNetUri.toPath
-		return URI.createURI(path)
+		// URI is used to get path of the uri without scheme
+		var path = URI.createURI(pathWithScheme).path
+		return URI.createURI(path.toPath)
 	}
 
 	def String toPath(URI uri) {
```
akosyakov: This method should ensure that the returned value starts with `file:///` if the scheme is `file`. It should work as for URIs created via `UriExtensions.toUri` as well as for URIs created with `URI.createURI` and `URI.createFileURI`. Unfortunately, there are not tests for it.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
```diff
@@ -0,0 +1,49 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.server
+
+import org.junit.Test
+
+import static org.junit.Assert.*
+
+class UriExtensionsTest extends AbstractTestLangLanguageServerTest {
```
akosyakov: It would be to test different  `URI` <-> `string` conversions with an instance of `UriExtension`.

```xtend
@RunWith(XtextRunner)
@InjectWith(TestLanguageIdeInjectorProvider)
class UriExtensionsTest extends Assert {

	@Inject
	extension UriExtensions;
	
	@Test
	def void testConversion_01() {
		assertEquals("file:///dir/name.ext", "file:///dir/name.ext".toUri.toPath)
	}
	
	@Test
	def void testConversion_02() {
		assertEquals("file:///dir/name.ext", URI.createFileURI("/dir/name.ext").toPath)
	}

}
```

and so on

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
```diff
@@ -0,0 +1,49 @@
...
+		initialize
+
+		val fileURI = 'Foo Bar.testlang'.virtualFile
+		fileURI.open('''
+			type FooBar {
+			}
+		''')
+
+		diagnostics.get(fileURI).forEach [
+			println("err: " + it.message)
```
akosyakov: better to put assertions then printing if you expect something

akosyakov:
@xcariba please take care of comments

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	@Test
+	def void testFileUriConversion() {
+		assertEquals("file:///dir/name.ext", URI.createFileURI("/dir/name.ext").toPath)
+	}
+
+	@Test
+	def void testFilesWithSpaces() {
+		assertEquals("file:///dir/Foo Bar.testlang", "/file://dir/Foo Bar.testlang".toUri.toPath)
```
szarnekow: Why is there a leading slash / in the pathWithScheme?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
```diff
@@ -20,21 +22,31 @@ import org.eclipse.emf.common.util.URI
...
+		return URI.createURI(uri.path.toPath).toString
 	}
 
 	def String toPath(java.net.URI uri) {
+		return uri.path.toPath
+	}
+
+	/**
+	 * We need to check if current path represents directory in file system
+	 * and need to add trailing slash if path represents directory.
```
szarnekow: Is the comment up to date? If yes, could you please add a testcase? Or did I just miss something?

xcariba: There is a test for that, UriExtensionsTest.testFolderIsPrefix.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
xcariba: My bad, It's a typo

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
szarnekow: That relation was not obvious to me. Sorry for the unnecessary noise.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
xcariba: Fixed and squashed commits

kittaakos:
It seems this change breaks all LSP tests where we use `inmemory:/` scheme for the URIs. I will double check this and if that is the case I will create a sperate ticket with failing tests.

kittaakos:
A straightforward test case that shows the regression:
```java
@Test
def void testInMemory_GH_403() {
  initialize;
  val uri = 'inmemory:/foo/bar.testlang';
  open(uri, '''
    type Foo { }
  ''');
  assertNotNull('''Diagnostics could be empty but never null.''', diagnostics.get(uri));
}
```

And the problem is the following:
Previous behavior before (GH-403):
 - The input `pathWithScheme` argument was `inmemory:/foo/bar.testlang`.
 - The converted `javaNetUri` URI bacome `inmemory:/foo/bar.testlang`.
 - The file was not found: hence it returned with the `toString` of the URI: `inmemory:/foo/bar.testlang`
 - From this string we have created a EMF (hierarchical) URI from the `inmemory:/foo/bar.testlang`.

Current behavior:
 - The input `pathWithScheme` argument was `inmemory:/foo/bar.testlang`.
 - After creating the EMF (hierarchical) URI with the argument we get the following URI: `inmemory:/foo/bar.testlang`.
 - The path of the URI is `/foo/bar.testlang`.
 - Inside the `toPath` method the `path.toUri().toString()` will construct the following URI: `file:///foo/bar.testlang` which is a regression comapred to the previous behavior.

akosyakov:
>Inside the toPath method the path.toUri().toString() will construct the following URI: file:///foo/bar.testlang which is a regression comapred to the previous behavior.

it is bogus and should be fixed

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
kittaakos: Could you please clarify that, why this is required? Assuming a very simple test case, the `isPrefix` returns with a `false` for the `URI.createFileURI`. So basically, when I have an EMF URI and I convert it to string and back to URI then I would get a different URI than I had initially. (Or I have just overlooked something which could easily happen).

Test case:
```java
	@Test
 	def void testFolderIsPrefix_why() {
 		var directory = new File("./test-data/test-project")
 		assertTrue(URI.createFileURI(directory.absolutePath).isPrefix)
 	}
```

(Sorry for the formatting.)

xcariba: Because `UriUtil.isPrefixOf` checks if `URI` has trailing path separator. It is used in `FileProjectConfig`. If we do not add trailing separator `FileProjectConfig` will not find any projects associated with given file. That is why I convert input URI to string path and create a new hierarchical one. I did not realize that it removes `inmemory:/` scheme, since it passed tests with this scheme.

kittaakos: Cool, thanks for the update! One more question, if I have this:

```
var directory = new File("./test-data/test-project")
URI.createFileURI(directory.absolutePath).isPrefix //false
```

Why would you expect to have `isPrefix` `true` after the string then to URI conversion. Besides that, this is a behavior change if I understood you correctly, and it is unrelated to the URL encoding/decoding issue at all, right?

PS: and do not feel bad because of the `inmemory:/` problem, that's why we need to add test and do the review correctly.

kittaakos: > and it is unrelated to the URL encoding/decoding

If that is true, can we handle this in another PR/issue?

xcariba: `File("./test-data/test-project")` is an actual directory, located in `org.eclipse.xtext.ide.tests` during tests. It calls `toPath(String)` method, that uses `Paths.get()` to form a path and converts it to URI. And `java.nio.file.Path.toUri()` states that
> ... if file is a directory, then the resulting URI will end with a slash

So it should return true it that case. Maybe it is not the best idea, since it depends on other tests.
Original Cyrillic letter decoding is fixed by `URLDecoder`, but while debugging I found out that spaces are not supported either, because `java.net.uri` does not support space symbol. I`m not sure if it is related to URI decoding in general, but yes, it is not related to Cyrillic letters decoding.

svenefftinge: This conversion shouldn't be done in general, but rather where needed. Also the code you mentioned seems to take care of it already : https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext/src/org/eclipse/xtext/workspace/FileProjectConfig.xtend#L98

svenefftinge: Could you tell us the concrete scenario in which the `isPrefixOf` didn't return the desired result?

PR #405, opened 2017-07-24
[LSP] Fixed concurrency issues in RequestManager

svenefftinge:
Signed-off-by: Sven Efftinge <sven.efftinge@typefox.io>

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
@@ -7,16 +7,13 @@
  *******************************************************************************/
 package org.eclipse.xtext.ide.server.concurrent
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder
 import com.google.inject.Inject
-import java.util.concurrent.CancellationException
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionException
 import java.util.concurrent.CopyOnWriteArrayList
 import java.util.concurrent.ExecutorService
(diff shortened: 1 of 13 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
```diff
@@ -38,96 +35,79 @@ import org.eclipse.xtext.util.CancelIndicator
...
+	var CompletableFuture<Void> writePending = CompletableFuture.completedFuture(null)
+
+	def <U, V> CompletableFuture<V> runWrite(()=>U nonCancellable, (CancelIndicator, U)=>V request) {
+		val result = new CompletableFuture<V>()
+		writePending = new CompletableFuture<Void>()
+		parallel.submit[
+			try {
+				cancel()
+				w.lock
+				writePending.complete(null)
```
spoenemann: This reference to `writePending` could be different from the one created in line 79. In this case there can be readers waiting forever. Make sure that the CompletableFuture that is completed here is really the one that was created for this write request.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
```diff
@@ -38,96 +35,79 @@ import org.eclipse.xtext.util.CancelIndicator
...
+				// run the non-cancelable work
+				val intermediateResult = nonCancellable.apply()
+				
+				// run the cancelable work
+				val cancelIndicator = new RequestCancelIndicator(result)
+				try {
```
spoenemann: The two `try` blocks could be unified.

Issue #408, opened 2017-07-28
[textRegionAccess] Allow to create semantic regions for hidden nodes


AlexTugarev:
`NodeModelBasedRegionAccessBuilder` lacks protected getters/setter for state, which would make it easier in a subclass to create semantic regions for hidden nodes.

kthoms:
Merged.

AlexTugarev:
PR https://github.com/eclipse/xtext-core/pull/410 to fix missing import.

PR #409, opened 2017-07-28
[textRegionAccess] Make state of `NodeModelBasedRegionAccessBuilder` accessible in subclasses

AlexTugarev:
Cf. #408


kthoms:
LGTM. Thanks for the patch!

PR #414, opened 2017-08-08
Unifies workspace configs

akosyakov:
It allows working with documents, not under the root URI, regardless whether they are in memory or not. Plus removing duplicated functionality and cleaning up APIs.


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/IWorkspaceConfigFactory.xtend
```diff
@@ -30,16 +33,92 @@ interface IWorkspaceConfigFactory {
...
+
+	new() {
+		addProject(unknown)
+	}
+
+	override getProjects() {
+		projects.values.toSet
+	}
+
+	def FileProjectConfig addProject(URI uri, String name) {
```
svenefftinge: These factory methods should go to `ProjectWorkspaceConfigFactory` so it can be overridden through DI.
Also, this is not a heavily used API so we don't need all the convenient overrides.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/IWorkspaceConfigFactory.xtend
```diff
@@ -30,16 +33,92 @@ interface IWorkspaceConfigFactory {
...
+		return workspaceConfig
+	}
+
+	def void findProjects(MultiProjectWorkspaceConfig workspaceConfig, URI uri) {
+		workspaceConfig.addProject(uri)
+	}
+
+}
+
+class MultiProjectWorkspaceConfig implements IWorkspaceConfig {
```
svenefftinge: Should be `DefaultWorkspaceConfig` or `WorkspaceConfig`

org.eclipse.xtext/src/org/eclipse/xtext/workspace/FileWorkspaceConfig.xtend
```diff
@@ -10,32 +10,35 @@ package org.eclipse.xtext.workspace
 import java.io.File
```
svenefftinge: file should be renamed

Issue #419, opened 2017-08-10
MWE Workflow generates an invalid method in the content assist parser for large grammars 


SailorJoe6:
As described [in the Xtext community forums here](https://www.eclipse.org/forums/index.php/m/1770381/#msg_1770381), when a grammar is large enough, the MWE Workflow generates classes and methods larger than the maximum allowed size for java (>64kb).  

The recommended technique for solving this problem is to enable class splitting in the Antler parser generator, by using **XtextGeneratorLanguage** (vice StandardLanguage) in the workflow and adding options to the antlr parser generator such as the following: 
```
fragment = parser.antlr.XtextAntlrGeneratorFragment2 {
    options = {
        classSplitting = true
        casesPerSpecialStateSwitch= "200"
        fieldsPerClass = "200"
        methodsPerClass = "200"
    }
}
```

While this method ensures that classes don't have too many fields or methods in them, it does **not** address the issue of methods with too many lines of code.  As a result, when the grammar is large enough, the Parser file generated in the ide project will refuse to compile with the following error message:
```
The code of constructor x() is exceeding the 65535 bytes limit. 
```
Assuming a dsl named **MyDsl**, this error is found in file mydsl.ide > src-gen > mydsl > ide > contentassist > antlr > MydslParser.java, in the function getRuleName() in the constructor to the nameMappings field. 

The  function looks as follows:
```
@Override 
protected String getRuleName(AbstractElement element) {
  if (nameMappings == null) {
    nameMappings = new HashMap<AbstractElement, String>() {
      private static final long serialVersionUID = 1L;
      {
        put(grammarAccess.getGreetingAccess().getGroup(), "rule__Greeting__Group__0");
        put(grammarAccess.getModelAccess().getGreetingsAssignment(), "rule__Model__GreetingsAssignment");
        put(grammarAccess.getGreetingAccess().getName1Assignment_1(), "rule__Greeting__Name1Assignment_1");
        put(grammarAccess.getGreetingAccess().getName2Assignment_2(), "rule__Greeting__Name2Assignment_2");
        put(grammarAccess.getGreetingAccess().getName3Assignment_3(), "rule__Greeting__Name3Assignment_3");
        put(grammarAccess.getGreetingAccess().getName4Assignment_4(), "rule__Greeting__Name4Assignment_4");
        put(grammarAccess.getGreetingAccess().getName5Assignment_5(), "rule__Greeting__Name5Assignment_5");
[...]

PR #420, opened 2017-08-10
generate initialization of nameMapping with chunked method calls #419

cdietrich:
generate initialization of nameMapping with chunked method calls #419


cdietrich:
@szarnekow do you know a better solution for this problem?

cdietrich:
or should i introduce something like
```
			parserGenerator = {
				splitNameMapping = true
			}
```

PR #428, opened 2017-08-17
[refactoring] extracted LTK independent APIs to ide

JanKoehnlein:


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/refactoring/RefactoringIssueAcceptor.xtend
```diff
@@ -0,0 +1,38 @@
...
+import org.eclipse.xtext.util.ITextRegion
+
+/**
+ * @author koehnlein - Initial contribution and API
+ * @since 2.13
+ */
+interface RefactoringIssueAcceptor {
+	
+	enum Severity {
+		FATAL, ERROR, WARNING, INFO, OK
```
cdietrich: is there a reason to have the inverse order compared to `RefactoringStatus`

JanKoehnlein: I made the order symmetric to `xtext.validation.Severity`, which could not be reused because of the missing `FATAL` literal.
The order should only matter if you compare two Severities with `<`or similar. This would already behave strange in `xtext.validation.Severity`. I'd not recommend to compare literals with this way. Am I missing something else?

cdietrich: i am more thinking of the "adapter" usecase RefactoringIssueAcceptor.Severity <=> RefactoringStatus

JanKoehnlein: It's a switch-case where the order should not matter
https://github.com/eclipse/xtext-eclipse/commit/241c1f73014df6af5d4dae84db745a9d6c91acab#diff-52aef468dd0def3666d5565c8b9f4df7

cdietrich: +1

cdietrich: so as long as nobody looks at the backing int values of the enum everything is fine

PR #440, opened 2017-08-22
Handle the case that dynamicRegistration is null

spoenemann:
According to the LSP, `dynamicRegistration` can be undefined in the JSON message.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/commands/ExecutableCommandRegistry.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/commands/ExecutableCommandRegistry.xtend
@@ -35,10 +35,10 @@ import org.eclipse.xtext.util.internal.Log
 	
 	LanguageClient client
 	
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/commands/ExecutableCommandRegistry.xtend
```diff
@@ -38,7 +38,7 @@ import org.eclipse.xtext.util.internal.Log
 	def void initialize(Iterable<? extends IResourceServiceProvider> allLanguages, ClientCapabilities capabilites, LanguageClient client) {
 		this.client = client
 		registeredCommands = HashMultimap.create
-		val hasDynamicRegistration = capabilites.workspace.executeCommand.dynamicRegistration
+		val boolean hasDynamicRegistration = capabilites.workspace.executeCommand.dynamicRegistration ?: false
```
svenefftinge: `executeCommand` is optional, too.

Issue #446, opened 2017-08-29
o.e.xtext.ide depends on Equinox


sjka:
I'm currently trying to get LSP support into Eclipse SmartHome (eclipse/smarthome#4148). 
However, as it turned out, the `org.eclipse.xtext.ide` has a "require bundle" dependency to `org.eclipse.core.runtime` which in turn drags the whole equinox enchilada in. As it claims to provide "Platform-independent language support.", this is somewhat surprising.

There was a related issue for the xtext core in bugzilla (https://bugs.eclipse.org/bugs/show_bug.cgi?id=434490). Would it be possible to factor out the eclipse specific stuff into an optional eclipse-related bundle? Or at least make this stuff optional?

cdietrich:
it uses org.eclipse.core.runtime.IProgressMonitor /NullProgressMonitor from equinox common but this change wouldnt help you neither correct?

and org.eclipse.xtext has the same dependency (although its optional)

sjka:
> from equinox common 

looking at its dependencies: nope 😄 

> and org.eclipse.xtext has the same dependency (although its optional)

An optional dependency is perfectly fine - as long as the stuff really doesn't fail without it. Our use-case is to run an embedded language server, so no Eclipse extension points are registered. Therefore I assume we should be reasonably safe...

cdietrich:
no i mean: replacing that dependency with equinox common: will it help you or not?
or is that  == equinox enchilada

sjka:
No, it won't help:

o.e.equinox.commons imports

```
# Bad:
org.eclipse.equinox.log;version="[1.0,2.0)",
org.eclipse.osgi.framework.log;version="[1.1, 2.0)"
org.eclipse.osgi.service.datalocation
org.eclipse.osgi.service.debug
org.eclipse.osgi.service.localization
org.eclipse.osgi.service.urlconversion
org.eclipse.osgi.util

# Good:
org.osgi.framework
org.osgi.service.log;version="[1.3.0, 2.0)"
org.osgi.service.packageadmin
org.osgi.service.url
org.osgi.util.tracker
```

cdietrich:
but org.eclipse.xtext has the same required dependency

cdietrich:
https://github.com/eclipse/xtext-core/blob/077b26c5caad58f0e07b35e7001415b8db93bc33/org.eclipse.xtext/META-INF/MANIFEST.MF#L106

sjka:
[...]

PR #447, opened 2017-08-29
made o.e.xtext.ide dep. to org.eclipse.core.runtime optional #446

cdietrich:
made o.e.xtext.ide dep. to org.eclipse.core.runtime optional #446

### DIFF
--- a/org.eclipse.xtext.ide/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.ide/META-INF/MANIFEST.MF
@@ -8,7 +8,8 @@ Bundle-ActivationPolicy: lazy
 Bundle-Vendor: Eclipse Xtext
 Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.eclipse.xtend.lib,
- org.eclipse.core.runtime;bundle-version="3.6.0",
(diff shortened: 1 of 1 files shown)


cdietrich:
please review

org.eclipse.xtext.ide/META-INF/MANIFEST.MF
```diff
@@ -8,7 +8,8 @@ Bundle-ActivationPolicy: lazy
 Bundle-Vendor: Eclipse Xtext
 Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.eclipse.xtend.lib,
- org.eclipse.core.runtime;bundle-version="3.6.0",
+ org.eclipse.core.runtime;bundle-version="3.6.0";resolution:=optional;x-installation:=greedy,
+ org.eclipse.equinox.common;bundle-version="3.5.0",
```
kthoms: why 3.5 here while core.runtime would be 3.6?

cdietrich: copy&paste from xtext-core?!?

kthoms: OK, if this valid for o.e.x then it should be fine.

PR #454, opened 2017-09-08
[refactoring] Core part https://github.com/eclipse/xtext-eclipse/issues/284

JanKoehnlein:
Reference descriptions should always use path URIs for the source. 

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
@@ -31,6 +31,7 @@ import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.emf.common.notify.Notifier;
 import org.eclipse.emf.common.util.AbstractTreeIterator;
 import org.eclipse.emf.common.util.EList;
+import org.eclipse.emf.common.util.SegmentSequence;
 import org.eclipse.emf.common.util.TreeIterator;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EClass;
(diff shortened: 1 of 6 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
```diff
@@ -773,4 +775,58 @@ protected EObject computeNext() {
...
+		if (!isContained) {
+			return "/-1";
+		}
+		builder.append(getURIFragmentRootSegment(internalEObject));
+		builder.append("");
+		builder.reverse();
+		return builder.toSegmentSequence().toString();
+	}
+	
+	protected static String getPathURIFragment(InternalEObject container, EStructuralFeature feature, InternalEObject contained) {
```
meysholdt: should this be named `getPathURISegment`?

org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
```diff
@@ -773,4 +775,58 @@ protected EObject computeNext() {
 			}
 		};
 	}
+	
+	public static URI getPathURI(final EObject object) {
```
meysholdt: maybe call this `getIndexBasedPathURI` or `getNameAgnosticURI` since the segments of n URI always form a *path* and the term *path* does not express that the URI is agnostic of names.

org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
JanKoehnlein: I have often heard the name _fragment path_ for fragments like this, so I changed it to this.

PR #455, opened 2017-09-11
[refactoring] refactored resource relocation and added CopyParticipant

JanKoehnlein:
I refactored the entire thing and added a CopyParticipant (in xtext-eclipse).

- Clients should now extend IResourceRelocationStrategy.AbstractImpl
- Clients can hook into loading of resources as well
- The execution now goes in two phases:
  - for all changes `loadAndWatchResource` using the first relocation strategy that is responsible
  - for all changes `applyChange` using the first relocation strategy that is responsible and apply all side effects from all strategies
- This rendered the `ResourceModification` interface obsolete.
- A `ResourceRelocationChange` now carries information about the kind of user event that triggered it (RENAME, MOVE or COPY) and whether it is a folder or file change.

TBD: There still is a slight asymmetry when to start watching resources, as the change serializer impl doesn't allow watching resources that are not yet loaded.


JanKoehnlein:
See also https://github.com/eclipse/xtext-eclipse/pull/362

PR #457, opened 2017-09-12
Fixed regression in URIExtension.

### LINKED ISSUE #456: Regression in `UriExtensions`: `inmemory:/` URI scheme is not supported any more
After merging GH-403 into the master branch, `inmemory:/` URI scheme cannot be used for testing as those will be converted into `file:///` URI scheme.

Test case:
```java
@Test
def void testInMemory_GH_403() {
  initialize
  val uri = 'inmemory:/foo/bar.testlang';
  open(uri, '''
    type Foo { }
  ''');
  assertNotNull('''Diagnostics could be empty but never null.''', diagnostics.get(uri));
}
```

Additional details are [here](https://github.com/eclipse/xtext-core/pull/403).

kittaakos:
Task: #456 
Closes #456.

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
@@ -9,11 +9,11 @@ package org.eclipse.xtext.ide.tests.server
 
 import com.google.common.base.CharMatcher
 import com.google.common.base.StandardSystemProperty
+import com.google.inject.Inject
 import java.io.File
 import java.net.URI
 import java.nio.file.Files
 import java.nio.file.Paths
-import javax.inject.Inject
 import org.eclipse.xtext.ide.server.UriExtensions
 import org.eclipse.xtext.ide.tests.testlanguage.TestLanguageIdeInjectorProvider
 import org.eclipse.xtext.testing.InjectWith
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
@@ -256,7 +256,7 @@ class UriExtensionsTest {
 		var directory = new File("./test-data/test-project")
 		assertTrue(directory.exists)
 		assertTrue(directory.directory)
-		var uri = createEmfFileURI(directory.canonicalPath).toPath.toUri
+		var uri = createEmfFileURI(directory.absolutePath).toPath.toUri
 		assertTrue(uri.isPrefix)
 	}
 	
@@ -265,7 +265,7 @@ class UriExtensionsTest {
 		var directory = new File("./test-data/test-project")
 		assertTrue(directory.exists)
 		assertTrue(directory.directory)
-		var uri = createEmfFileURI(directory.canonicalPath + "/").toPath.toUri
+		var uri = createEmfFileURI(directory.absolutePath + "/").toPath.toUri
 		assertTrue(uri.isPrefix)
 	}
 	
@@ -275,7 +275,7 @@ class UriExtensionsTest {
 		assertTrue(directory.exists)
 		assertTrue(directory.directory)
 		assertTrue(CharMatcher.WHITESPACE.matchesAnyOf(directory.name));
-		var uri = createEmfFileURI(directory.canonicalPath).toPath.toUri
+		var uri = createEmfFileURI(directory.absolutePath).toPath.toUri
 		assertTrue(uri.isPrefix)
 	}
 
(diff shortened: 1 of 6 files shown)


org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
```diff
@@ -50,8 +254,46 @@ class UriExtensionsTest {
...
+		assertTrue(directory.exists)
+		assertTrue(directory.directory)
+		assertTrue(CharMatcher.WHITESPACE.matchesAnyOf(directory.name));
+		var uri = createEmfFileURI(directory.absolutePath).toPath.toUri
 		assertTrue(uri.isPrefix)
 	}
-}
 
+	private def createEmfURI(String uri) {
+		return org.eclipse.emf.common.util.URI.createURI(uri);
```
akosyakov: you could use static extension import

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
```diff
@@ -50,8 +254,46 @@ class UriExtensionsTest {
...
+	}
+	
+	private def createEmfFileURI(String uri) {
+		return org.eclipse.emf.common.util.URI.createFileURI(uri);
```
akosyakov: the same here

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
```diff
@@ -21,32 +27,100 @@ import java.nio.charset.StandardCharsets
...
+	 * Converts the EMF URI argument into a string path.
+	 * <p>
+	 * If the argument is a URI with {@link URI#isFile() file} scheme, then the produced result value will always have this
+	 * form {@code file:///path/to/resource, otherwise it returns with the string representation
+	 * of the argument.
+	 */
 	def String toPath(URI uri) {
-		return URI.createURI(uri.path.toPath).toString
+		if (uri.file) {
+			val segments = uri.toSegments;
```
akosyakov: I am not sure about it, could it have some query params via emf URI?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
```diff
@@ -21,32 +27,100 @@ import java.nio.charset.StandardCharsets
...
+	 * </p>
+	 */
 	def URI toUri(String pathWithScheme) {
-		// URI is used to get path of the uri without scheme
-		var path = URI.createURI(pathWithScheme).path
-		return URI.createURI(path.toPath)
+		// This will take care of the encoding.
+		val uri = URI.createURI(pathWithScheme.checkNotNull('pathWithScheme'));
+		if (uri.file) {
+			val segments = uri.toSegments;
```
akosyakov: I am a bit afraid of manual parsing/serialization, we could lose some info from URIs, like query params

kittaakos: Although I totally agree with you, I am positive the current state [does not handle this either](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend#L26).

The ball is in your court...

akosyakov: ok, the state before as well

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/UriExtensions.xtend
```diff
@@ -21,32 +27,100 @@ import java.nio.charset.StandardCharsets
...
+			val prefix = try {
```
svenefftinge: I think the locations that assume a uri with a trailing empty segment (prefix-uri) should take care. 
Handling it in a central manner will alter URIs surprisingly in other cases, also it won't work for non existing directories but should.
We already do stuff like that here : https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext/src/org/eclipse/xtext/workspace/FileProjectConfig.xtend#L98

svenefftinge: I couldn't find a location where the `org.eclipse.xtext.util.UriUtil.isPrefixOf(URI, URI)` is used with a uri that has not been converted. Why was it added here?
If you find such place, please use `org.eclipse.xtext.util.UriUtil.toFolderURI(URI)`.

svenefftinge: Maybe not `isPrefixOf` but `resolve` caused the issue. We should make sure that `FileProjectConfig` is in all cases initialized with a "FolderUri" (see URIUtils).

kittaakos:
Closing in favor of #458.

PR #458, opened 2017-09-13
Fix for the weird uri conversions

svenefftinge:


svenefftinge:
do not review yet

svenefftinge:
@kittaakos could you have a look?

kittaakos:
Sure.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
```diff
@@ -7,51 +7,263 @@
...
+			URI.create('file://dir/\u0424\u0443%20\u0411\u0430\u0440').toUriString
+		);
+	}
+
+
+	@Test
+	def void test_toUriString_05() {
+		assertEquals(
+			'file://path/to/resource/',
+			URI.create('file://path/to/resource/').toUriString
```
kittaakos: This test case contradicts with the documentation you have written:
```
converts a java.net.URI into a string representation, if it is a file uri, the scheme will be followed with <code>///</code> (i.e. empty authority) 
```

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
```diff
@@ -7,51 +7,263 @@
...
+			'file:///dir/\u0424\u0443%20\u0411\u0430\u0440',
+			URI.create('file:/dir/\u0424\u0443%20\u0411\u0430\u0440').toUriString
+		);
+	}
+
+	@Test
+	def void test_toUriString_04() {
+		assertEquals(
+			'file://dir/\u0424\u0443%20\u0411\u0430\u0440',
+			URI.create('file://dir/\u0424\u0443%20\u0411\u0430\u0440').toUriString
```
kittaakos: Same as below. I am a bit puzzled, so when the file URI has two slashes it remains the same otherwise it will be three forward slashes? Is that the logic?

svenefftinge: Yes, in that case the first segment is the authority. I will update the documentation.

svenefftinge: The other conversion is ok, according to https://en.wikipedia.org/wiki/Talk%3AFile_URI_scheme

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/OpenDocumentTest.xtend
```diff
@@ -132,4 +132,15 @@ class OpenDocumentTest extends AbstractTestLangLanguageServerTest {
...
     }
+
+    @Test
+    def void testInMemory_GH_456() {
+        initialize
+        val uri = 'inmemory:/foo/bar.testlang';
+        open(uri, '''
+            type Foo { }
+        ''');
+        assertNotNull('''Diagnostics could be empty but never null.''', diagnostics.get(uri));
```
kittaakos: Just in case you make any additional modification in the PR, could you please fix my misleading assertion message. That is eventually not true. It can be null, but not in this case.

kittaakos:
Looks good. The problem in the downstream project is resolved with your changes. Thank you for cleaning it up!

kittaakos:
Wait a second. I have critical issues on Windows with these changes. Let me double check before you merge it.

kittaakos:
I have double checked it and worked correctly with this version. Thanks for your patience!

cdietrich:
tests are red on windows

cdietrich:
(well all the change serializer stuff as well, we should consider something to get more stable tests on windows)

Issue #470, opened 2017-09-25
[generator] Contribute org.eclipse.core.contenttype.contentTypes extension


kthoms:
See eclipse/xtext-eclipse#379.

Xtext's code generator has to provide a content type descriptor for DSL files.

cdietrich:
anything left?

kthoms:
I guess we have updated all languages now.

PR #471, opened 2017-09-25
[#470] Contribute org.eclipse.core.contenttype.contentTypes extension

kthoms:
Define a content type for a DSL file with base type text. This will avoid that a text search would try to determin whether the content is binary. This can impact search performance.


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ImplicitFragment.xtend
```diff
@@ -276,6 +276,15 @@ package class ImplicitFragment extends AbstractStubGeneratingFragment {
 				</activeWhen>
 			</handler>
 		</extension>
+		<extension point="org.eclipse.core.contenttype.contentTypes">
+			<content-type
+				base-type="org.eclipse.core.runtime.text"
+				file-extensions="�language.fileExtensions.join(",")�"
+				id="�name�.contenttype"
+				name="�it.simpleName� File"
```
cdietrich: inconsistent use of it.simpleName

cdietrich:
would like to see all tests running with xtext xbase xtend etc to make sure this has no negative side effects

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ImplicitFragment.xtend
cdietrich: also it would be nice to opt out of this. e.g. our customer has defined a bunch of content types and this one makes merging plugin.xml harder

kthoms: where should this been configured? Maybe PluginXmlAccess?
```
			project = model.project.StandardProjectConfig {
				baseName = projectName
				rootPath = ".."
				eclipsePlugin = {
					pluginXml = {
						contentType = false
					}
				}
			}
```

cdietrich: maybe a custom BundleProjectConfig subclass for 

```
eclipsePlugin = {
...
```

kthoms: Hmm, not sure. This would mean to specify some special class there which no-one knows of. Makes things complicated.
Does your customer contribute content type for DSL files? Would it be an option to remove that? I'm also not sure if it would harm if there are multiple descriptors. Do they do something different?

cdietrich: problem is: we have different content types for different dsls. we can merge plugin.xmls manually but this complicates it. i dont know how we deal with other stuff e.g. the registation of resource factories which is in a own fragment. and again its disabling is painful. maybe we just live with it cause disabling implicit stuff is total mess in the new generator

cdietrich: maybe we need a disabled flag in AbstractXtextGeneratorFragment

kthoms: Disabling by configuring PluginXmlAccess as I described above would work (just tried that) and is rather low invasive. Maybe more configuration switches there could be used also what is added implicitly?

cdietrich: i think we should discuss that in a separate issue. ill open one. leave it as is for now

PR #474, opened 2017-09-26
https://github.com/Microsoft/language-server-protocol/blob/master/pro…

RealTYPICAL:
…tocol.md Server should exit with success or fail code when exit has been called.

Exit Notification

A notification to ask the server to exit its process. The server should exit with success code 0 if the shutdown request has been received before; otherwise with error code 1.

Notification:

method: 'exit'
params: void

from https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md

Signed-off-by: James Tooley jetter019@msn.com


kthoms:
Yeah, finally!

RealTYPICAL:
>.>

kthoms:
There are now two Sign-Off lines. Could you update the commit message and just leave the latter one?

cdietrich:
I miss the generated Java file in the commit

cdietrich:
i have no idea on how to tell vscode to call that method ....

cdietrich:
thanks for the contribution

PR #482, opened 2017-09-28
First shot at ChangeSerializer based rename refactoring

JanKoehnlein:
Includes rename for Xtext LSPs


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/refactoring/IRenameNameValidator.xtend
```diff
@@ -0,0 +1,56 @@
...
+
+		@Inject IValueConverterService valueConverterService
+
+		override validate(EObject target, String newName, RefactoringIssueAcceptor issues) {
+			try {
+				var String value = getNameAsValue(newName)
+				var String text = getNameAsText(value)
+				if (!equal(text, newName)) {
+					issues.add(ERROR, '''Illegal name: '�newName�'. Consider using '�text�' instead.''')
+				}
```
meysholdt: In case a value converter normalizes values this validation enforces that users specify the already-normalized value. 
For example, for the following value converter users would not be allowed to use "null" as a value even though during parsing it would be allowed.
```
int getValue(String it) {
  case "0": return 0;
  case "null": return 0;
}
String getText(int it) {
  return String.valueOf(it)
}
```

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/refactoring/IRenameStrategy.xtend
```diff
@@ -0,0 +1,69 @@
...
+ * 
+ * @author koehnlein - Initial contribution and API
+ * @since 2.13
+ */
+@ImplementedBy(IRenameStrategy.DefaultImpl)
+interface IRenameStrategy {
+
+	def void applyRename(RenameContext context)
+	
+	def String getCurrentName(EObject element)
```
meysholdt: Not sure how `getCurrentName(EObject element)` makes sense. 
* The user triggers rename refactoring on a token, not on an EObject
* EObjects can have multiple names at the same time, i.e. multiple renamable EAttributes.
* Since rename refactoring is triggered on a token, the token's value is the "current name", isn't it?

Maybe what we need here is method such as 
```
boolean canRename(ISemanticRegion token)
```
to check rename refactoring can be triggered on the current token?

org.eclipse.xtext.testlanguages.ide/src/org/eclipse/xtext/testlanguages/fileAware/ide/refactoring/FileAwareTestLanguageResourceRelocationStrategy.xtend
```diff
@@ -0,0 +1,29 @@
...
+	override applyChange(ResourceRelocationContext context) {
+		context.changes.filter[ canHandle ].forEach [ change | 
+			context.addModification(change) [ resource |
+				val rootElement = resource.contents.head
+				if (rootElement instanceof PackageDeclaration) {
+					val newPackage = change.toURI.trimSegments(1).segmentsList.drop(2).join('.')
+					rootElement.name = newPackage
+				}
+			]			
+		]
```
meysholdt: beautiful!

I'm undecided if we should support the use case to load a resource without recording a modification on it. 
Example: Load it, inspect model, discover nothing needs to be renamed, do nothing. 
Creating an empty recording will never be as fast as not creating a recording. Not clear though how much the difference is.

JanKoehnlein: We could make it protected for now.

PR #484, opened 2017-09-28
#483: Remove last whitespaces only when appending separator in for-loops

### LINKED ISSUE #483: TemplateNode removes whitespaces for for-loops with separators
Consider following code snippet:
````
@Traced def header(FunctionDefinition definition) {
	'''«FOR x : definition.parameters SEPARATOR ', '»«x.type» «x.name»«ENDFOR»);'''
}
````
The generated text does not contain the whitespaces between «x.type» and «x.name».

It seems like ```TemplateNote.appendImmediate()``` is deleting all previous whitespaces instead of the last ones only when appending the separator.

### PR DESCRIPTION
**Pull request by tkutz**
Fixes #483

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/generator/trace/node/TemplateNodeTest.xtend
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/generator/trace/node/TemplateNodeTest.xtend
@@ -77,6 +77,33 @@ class TemplateNodeTest {
 		''')
 	}
 	
+	@Test def void testSeparatorLoopWithWhitespace() {
+		val strings = #['a', 'b', 'c']
+		assertEquals('''
+			«FOR s : strings SEPARATOR ', '»«s» «s»«ENDFOR»
+		''')
+	}
+	
+	@Test def void testSeparatorLoopWithWhitespace2() {
+		val strings = #['a', 'b', 'c']
+		assertEquals('''
+			«FOR s : strings SEPARATOR ','»
+				«s» «s»
+			«ENDFOR»
+		''')
+	}
+
+	@Test def void testSeparatorLoopWithWhitespace3() {
+		val strings = #['a', 'b', 'c']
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by cdietrich** [comment]
there seem to be more cases that dont work correct. even with your fix e.g.

```
@Test def void testSeparatorLoopWithWhitespace2() {
		val strings = #['a', 'b', 'c']
		assertEquals('''
			«FOR s : strings SEPARATOR ','»
				«s» «s»
			«ENDFOR»
		''')
	}
```

**Comment by cdietrich** [comment]
i propose something like

``` 
	override appendImmediate(Object object, String indentation) {
		var removeNLCount = 0;
		for (var i = currentParent.children.size - 1; i >= 0; i--) {
			val node = currentParent.children.get(i)
			if (node instanceof TextNode && !(node as TextNode).text.hasContent) {
				currentParent.children.remove(i)
			} else if (node instanceof NewLineNode) {
				currentParent.children.remove(i)
				removeNLCount++;
			} else {
				append(object, indentation)
				for (var j=0; j<removeNLCount; j++) {
					newLine
				}
				return
			}
		}
		append(object, indentation)
	}
```

cdietrich:
@tkutz for ip validation: make sure you
* have signed the cla
* do the commit with -s (signoff)
* git user email == cla == signoff email

tkutz:
Another case is the combination of whitespace and newline:

````
@Test def void testSeparatorLoopWithWhitespace4() {
	val strings = #['a', 'b', 'c']
	assertEquals('''
		«FOR s : strings SEPARATOR ','»«s» 
		
		«ENDFOR»
	''')
}
````
(with whitespace after «s»)

@cdietrich That one is not working even with your extended fix.

cdietrich:
@tkutz did you miss to add test case 4?

tkutz:
@cdietrich it is incorporated in test case 3 (added a whitespace after «s»)

cdietrich:
Thx @tkutz

PR #486, opened 2017-10-02
Jk/rename cleanup

JanKoehnlein:


org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/refactoring/IRenameStrategy2.java
```diff
@@ -33,10 +32,9 @@
  * @author koehnlein - Initial contribution and API
  * @since 2.13
  */
-@ImplementedBy(IRenameStrategy.DefaultImpl.class)
 @SuppressWarnings("all")
-public interface IRenameStrategy {
-  public static class DefaultImpl implements IRenameStrategy {
+public interface IRenameStrategy2 {
```
meysholdt: The "2" is to avoid confusion with "org.eclipse.xtext.ui.refactoring.IRenameStrategy", isn't it? I'd be happy to no append the "2" because people who don't target Eclipse don't see the first version anyway. People who do target Eclipse should be able to distinguish the classes by qualified name. Discuss.

JanKoehnlein: In the Eclipse code we have to dispatch to both versions, so I think it is better to do the distinction.

meysholdt: alright.

PR #492, opened 2017-10-10
 [ChangeSerializer] update references with FQNs and semi-FQNs properly

meysholdt:


meysholdt:
build is green

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ReferenceUpdater.java
```diff
@@ -51,6 +56,54 @@
 	@Inject
 	private IValueConverterService valueConverter;
 
+	protected boolean containsReferenceText(Delta delta, IUpdatableReference ref) {
+		String text = ref.getReferenceRegion().getText();
```
szarnekow: getText says its the text of the document. Does it return the verbatim text or the text that was produced by a value converter. IQualifiedNameConverter expects the semantic text from the document, if I'm not mistaken.

Could you please add a testcase, that uses names like ^foo.^bar ?

meysholdt: Indeed, value conversion was missing. Fixed in latest patch set.

PR #497, opened 2017-10-12
Me more potential np es

meysholdt:


kthoms:
OK, but please let PRs review and approve by someone.

Issue #501, opened 2017-10-12
Java 9 and ASM don’t work together. 


cdietrich:
This lets mwe fail

https://www.eclipse.org/forums/index.php/t/1089402/

Did not find asm 6 on orbit.
If I manually add it it seems to work after widening our asm version ranges

Workaround: remove asm from workflow classpath

cdietrich:
@svenefftinge any comments?

cdietrich:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=525344

cdietrich:
Fixed in 2.13

PR #504, opened 2017-10-13
widened version range for asm to include version 6

cdietrich:
widened version range for asm to include version 6
https://github.com/eclipse/xtext-core/issues/501


cdietrich:
please review

PR #507, opened 2017-10-16
 [ChangeSerializer] added progress monitor, added documentation

meysholdt:


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/IChangeSerializer.java
```diff
@@ -41,21 +44,51 @@
...
+	 * other cross references known by {@link IResourceDescriptions} are updated.
+	 * 
+	 * Example: When a modification changes <code>myElement.name</code> from <code>"foo"</code> to <code>"bar"</code>,
+	 * all references pointing to <code>"foo"</code> are updated to <code>"bar"</code>. Yes, that's a rename
+	 * refactoring.
+	 * 
+	 * @see IReferenceUpdater
+	 * @see #setUpdateRelatedFiles(boolean)
+	 */
 	void setUpdateCrossReferences(boolean value);
```
szarnekow: What is the usecase of setting one of the properties to false?

meysholdt: `setUpdateCrossReferences(false)`:
* repair broken cross references
* better performance when the modification is known to not affect cross references
* have same behavior as the old semantic quickfix / ITextEditComposer implementations
`setUpdateRelatedFiles(false)`
* don't let a broken Xtext index cause wrong results.
* don't risk changing other files except the one that's currently modified.

and probably more...

Issue #509, opened 2017-10-18
[Java 9] CodeGenUtil2.getJavaDefaultTypes() does not reflect new types


cdietrich:

cdietrich:
Fixed in 2.14

PR #511, opened 2017-10-18
added missing types to CodeGenUtil2.getJavaDefaultTypes() #509

cdietrich:
added missing types to CodeGenUtil2.getJavaDefaultTypes() #509


cdietrich:
please review

cdietrich:
@svenefftinge i am not sure if this should already go into 2.13

cdietrich:
please review

szarnekow:
Should we change that into a Set implementation, that does Class.forName(..) internally on a contains(String)?

cdietrich:
am not sure: usecase: with a jdk 8 generate for jdk 9 bree

szarnekow:
Hmm true. Maybe a combination of both to be a little more robust against new types in the JDK?

cdietrich:
i wonder anyway if there is a contstand list somewhere in javalang ...

szarnekow:
Not that I'm aware of.

kthoms:
LGTM

Issue #512, opened 2017-10-19
[LSP] Input line to start LSP is too long


kdrozd:
Hi,

I have language that is in project with a bit long name (in.drozd.pdp.lang.tables) and it works as Eclipse plugin.
Now I added LSP support with wizard from 2.13 with following configuration:
* Eclipse support on
* maven build
* plain source layout
* regular LSP (not fatJar)

It works fine on macOS but on Windows 7 LSP is not starting du following errors:

> The input line is too long.
> The syntax of the command is incorrect.
> [Info  - 09:46:38] Connection to server got closed. Server will restart.
> The input line is too long.
> The syntax of the command is incorrect.
> [Info  - 09:46:38] Connection to server got closed. Server will restart.
> The input line is too long.
> The syntax of the command is incorrect.
> [Info  - 09:46:39] Connection to server got closed. Server will restart.
> The input line is too long.
> The syntax of the command is incorrect.
> [Error - 09:46:39] Connection to server got closed. Server will not be restarted.

My Configuration:
Eclipse 4.7.1a
Xtext 2.13-RC1 (tested with night builds as well)
Windows 7
Java 8

kdrozd:
Example VSCode plugin: https://marketplace.visualstudio.com/items?itemName=krzysiekdrozd.pdp-table-definition

kthoms:
If the start script produces a too long classpath, consider to use the Fat Jar.

cdietrich:
i wonder what is so long. the path to the start.bat? the classpath should be handled by the gradle application plugin correct?

kthoms:
In "Regular" format a start script is produced which lists all dependencies. When a DSL has many dependencies and stored in a deep path, tha classpath can grow long. This is a typical windows problem.
With a Fat Jar this becomes obsolete, the classpath is a single jar.

cdietrich:
i wonder if https://issues.gradle.org/browse/GRADLE-2992 should be reopended at github.com/gradle/gradle or has already been reopened

kdrozd:
But I use just Maven for build

cdietrich:
@kthoms does the maven plugin we use have an option to create a `classpath` file?

kthoms:
Please try to configure the [appassembler-maven-plugin](http://www.mojohaus.org/appassembler/appassembler-maven-plugin/assemble-mojo.html) in the IDE project:

```
			<plugin>
				<groupId>org.codehaus.mojo</groupId>
[...]

PR #514, opened 2017-10-19
[#512] Maven AppAssembler - Reduce length of classpath

kthoms:
On windows the generated classpath in start scripts might become too
long. This can be avoided by additional configuration of the
appassembler-maven-plugin plugin:
- added option repositoryLayout=flat
- added option useWildcardClassPath=true


kthoms:
If possible I suggest this for 2.13

cdietrich:
+1 for 2.13 => please merge when build is green

Issue #533, opened 2017-11-08
[LSP] Fat Jars: Dont Exclude xmi, xtextbin and tokens


cdietrich:
this files might be actually needed by xtext

cdietrich:
cc @cwehrheim

cdietrich:
fixed in 2.14

PR #536, opened 2017-11-08
dont exclude required xtextbin and tokens files from fat jars #533

cdietrich:
dont exclude required xtextbin and tokens files from fat jars #533


cdietrich:
please review

Issue #537, opened 2017-11-09
[Serializer] Add null guard


kthoms:
When a null reference is passed into Serializer#serialize a NPE is raised in ConcreteSyntaxValidator. It would be better to check this invalid argument up front.

The stack trace shows:
```
java.lang.NullPointerException
	at org.eclipse.xtext.validation.impl.ConcreteSyntaxValidator.isEObjectTransient(ConcreteSyntaxValidator.java:88)
	at org.eclipse.xtext.validation.impl.ConcreteSyntaxValidator.validateObject(ConcreteSyntaxValidator.java:102)
	at org.eclipse.xtext.validation.impl.ConcreteSyntaxValidator.validateRecursive(ConcreteSyntaxValidator.java:148)
	at org.eclipse.xtext.serializer.impl.Serializer.serialize(Serializer.java:126)
	at org.eclipse.xtext.serializer.impl.Serializer.serialize(Serializer.java:192)
	at org.eclipse.xtext.serializer.impl.Serializer.serialize(Serializer.java:58)

```

cdietrich:
Fixed in 2.14

PR #539, opened 2017-11-09
Fix for #537: Added null guards

FStolte:
Fix for issue #537: Added null guards to all relevant public methods of the Serializer class.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
@@ -45,6 +45,8 @@ import org.eclipse.xtext.validation.IConcreteSyntaxValidator;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 /**
  * @author Moritz Eysholdt - Initial contribution and API
  */
@@ -52,6 +54,7 @@ public class Serializer implements ISerializer {
 
 	@Override
 	public String serialize(EObject obj) {
+		checkNotNull(obj, "obj must not be null.");
 		return serialize(obj, SaveOptions.defaultOptions());
 	}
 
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -185,6 +197,10 @@ public String serialize(EObject obj, SaveOptions options) {
 
 	@Override
 	public void serialize(EObject obj, Writer writer, SaveOptions options) throws IOException {
+		if (obj == null || writer == null || options == null) {
+			throw new IllegalArgumentException(
```
kittaakos: This is exactly the place where you do not want to guess from a log file, which argument was eventually the `null` when your code is in production. I belive, this would be better.
```java
if (obj == null) { throw new IAE("obj"); }
if (writer == null ) { throw new IAE("write); }
// others...
```

Same for above.

FStolte:
Thank you for your suggestions.
I replaced the if-statements with `checkArgument()` calls and specified exactly which parameter is null in the exception message.

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -52,6 +52,7 @@
 
 	@Override
 	public String serialize(EObject obj) {
+		com.google.common.base.Preconditions.checkArgument(obj != null, "obj must not be null.");
```
cdietrich: any reason you dont import the type?

FStolte: No, not really, just copy-pasted the method. I just added the import.

kthoms:
@meysholdt Why `checkArgument` and not `checkNotNull` ?

kthoms:
Yes, if I would code it without Guava, I would go with an IAE. But the signature of checkNotNull is again clearer in the stack.

kittaakos:
I do not understand why Guava matters in this case, but as I said, I am with you. +1 for NPE.

cdietrich:
the problem is that a NPE happens way down the call hierarchy and not that the initial place.
so the check should happen at the api entry points

kittaakos:
I believe that was clear from the beginning. The question is whether we throw NPE or IAE in the `Serializer`. Which eventually, does not truly matter, right?

kthoms:
checkNotNull throws a NPE. I think it is not a matter of which exception to throw, the entry points have to be checked. I think @FStolte just should change to checkNotNull and it's fine then. I guess

cdietrich:
yes

FStolte:
The `checkArgument` calls are now replaced by `checkNotNull`.

FStolte:
There was a mistake in the parameter of `checkForNull`, it was still a null-comparison instead of the plain parameter. That is now fixed.

meysholdt:
Thank you for the contribution!

Issue #540, opened 2017-11-10
Export Junit4Fragment2 and SimpleProjectWizardFragment2 from org.eclipse.xtext.xtext.generator


ckulla:
I wanted to created my own StandardLanguage like language generator configuration. The classes Junit4Fragment2 and SimpleProjectWizardFragment2  are not accessible due to missing export entries in the MANIFEST.MF of org.eclipse.xtext.xtext.generator.

Furthermore in needed to change the error level of the discouraged access warning to all the other fragments. You should consider to remove the "x-internal" in the MANIFEST.MF.

see also comments here: http://typefox.io/xtexts-new-generator

cdietrich:
Fixed in 2.14

PR #541, opened 2017-11-10
export all packages as interal to ease subclassing #540

cdietrich:
export all packages as interal to ease subclassing #540


cdietrich:
please review

Issue #542, opened 2017-11-12
[Compare] syntax coloring in CompareEditor doesn't work with Xtext 2.13.0


tomaswolf:
This is caused by the addition of the `org.eclipse.core.contenttype.contentTypes` binding in [ImplicitFragment.xtend](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ImplicitFragment.xtend) in issue #470.

To fix, one needs to generate content-type bindings for `org.eclipse.compare.contentViewers` and `org.eclipse.compare.contentMergeViewers` in [CompareFragment2.xtend](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ui/compare/CompareFragment2.xtend), too.

cdietrich:
Cc @kthoms

tomaswolf:
I didn't override the compare fragment for my use case but just added the binding in the UI bundle's plugin.xml. The [Eclipse Help](https://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fextension-points%2Forg_eclipse_compare_contentMergeViewers.html) on these extension points tells what to add. Basically link the language's content type ID to the viewer ID using a &lt;contentTypeBinding> element.

cdietrich:
That what I meant. How does your plugin.xml look like after your change

tomaswolf:
What's so unclear about the Eclipse Help?
```
	<extension point="org.eclipse.compare.contentMergeViewers">
		<viewer .... as generated>
		</viewer>
		<contentTypeBinding
			contentTypeId="fully.qualified.myDSL.contenttype"
			contentMergeViewerId="fully.qualified.myDSL.compare.contentMergeViewers" />
	</extension>
```
and likewise for `contentViewers`. Try it out with xtend -- syntax coloring in the compare editor doesn't work for xtend, too.

kthoms:
I think Christian meant that if you have a working example it is easier for us to do the fix. Thanks for the hint! We will fix it soon.

brodo:
I'm working on it and will submit a PR today.

PR #547, opened 2017-11-15
Add contentTypeBinding to content[Merge]Viewers, refs #542 (v2)

brodo:
As I see it, this is everything that needs to be done here. Everything else is in other repos. Here is a screenshot of the compare view with the home automation example language:
compare


kthoms:
Thank you for this contribution!

Issue #555, opened 2017-11-22
TextRegionsToString handles windows line endings incorrectly


FStolte:
In `org.eclipse.xtext.formatting2.debug.TextRegionsToString::quote(String region, String maxLength)`, strings that exceed the maxLength are supposed to be shortened to exactly maxLength. In that case, the last three characters are also replaced with dots "...".

The dots are currently not added if there are windows line endings in the string. This also causes one test to fail when executed on a windows machine.

I already identified the cause for this problem and will be providing a fix shortly.

cdietrich:
@kthoms  anything left for this issue?

kthoms:
Fixed in 2.14

PR #556, opened 2017-11-22
Fix for issue [#555]

FStolte:
Since windows line endings are replaced by unix line endings - i.e. the `\r` character is removed -, the result string can be shorter than the input string.
Since the loop iterating over the characters of the input string is limited to `maxLength + 1` iterations, the check `result.size() > maxLength` did not give the desired result in the presence of windows line endings.

Now, the number of iterations is incremented when a character is deleted from the string.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionsToString.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionsToString.java
@@ -86,10 +86,10 @@ public class TextRegionsToString {
 				result.add("\\n");
 				break;
 			case '\r':
-				if (!ignoreCarriageReturnInQuotes) {
-					result.add("\\r");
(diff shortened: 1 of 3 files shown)


kthoms:
Please add a test for the changed method.

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionsToString.java
```diff
@@ -88,6 +88,8 @@ protected String quote(String string, int maxLength) {
 			case '\r':
 				if (!ignoreCarriageReturnInQuotes) {
```
cdietrich: maybe invert the conditions for better readability?

FStolte: Conditions are now swapped.

Issue #558, opened 2017-11-23
[LSP] Execute Command may not return null


cdietrich:
if execute command returns null, we get the following in vscode

> Received message which is neither a response nor a notification message

```
{
    "jsonrpc": "2.0",
    "id": "1"
}
```

i propose to return an object instead of null.

svenefftinge:
The protocol says it can be null | any. 
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#execute-a-command
Is it an lsp4j issue that we cannot distinguish between undefined field and value `null`? (cc @spoenemann)

cdietrich:
we decided to do so. see the discussion in #240

cdietrich:
> Response Message

Response Message sent as a result of a request. If a request doesn't provide a result value the receiver of a request still needs to return a response message to conform to the JSON RPC specification. The result property of the ResponseMessage should be set to null in this case to signal a successful request.

svenefftinge:
It's not sound though. This change suggests that lsp4j can distinguish if we use `https://github.com/eclipse/lsp4j/commit/35607f80d64abd42d77f33ed16f0a2a640d5ffec` (I am not sure, though). 
Just in case the command has a semantic on null vs. empty object.

cdietrich:
yes if lsp4j can handle this inbound it should be able to handle it outbound as well

spoenemann:
> Is it an lsp4j issue that we cannot distinguish between undefined field and value null?

I'd say it's a Java issue :-D

cdietrich:
so do you think we need a "NULL" literal/constant in lsp4j and do a special treatment for it?

spoenemann:
See eclipse/lsp4j#137.

cdietrich:
thx @spoenemann

PR #559, opened 2017-11-23
return new Object instead of null in ExecuteCommand Service #558

cdietrich:
return new Object instead of null in ExecuteCommand Service #558


cdietrich:
@svenefftinge would you be so kind to review?

cdietrich:
replaced by change in lsp

PR #561, opened 2017-11-27
[eclipse/xtext-eclipse#463] Fix for issue

FStolte:
- Created EditorFragment2 to generate language-specific sub-classes of
XtextEditor

- Added the new fragment to StandardLanguage

- Configured the language-specific editor in XbaseGeneratorFragment2, which solves [eclipse/xtext-eclipse#463]

- Added methods to XtextGeneratorNaming to get the qualified names of XtextEditor, XbaseEditor and
language-specific editors

- Added binding from XtextEditor to language-specific editors in ImplicitFragment


meysholdt:
Is it true that you're introducing an XtextEdiorFragment and enforce a custom generated subclass of XtextEditor for every Xtext language just so that `adaptableType="org.eclipse.xtext.ui.editor.XtextEditor"` has a unique type to match?
```
<extension point="org.eclipse.core.runtime.adapters">
    <factory class="org.eclipse.xtext.example.domainmodel.ui.DomainmodelExecutableExtensionFactory:org.eclipse.xtext.builder.smap.StratumBreakpointAdapterFactory2"
        adaptableType="org.eclipse.xtext.ui.editor.XtextEditor">
        <adapter type="org.eclipse.debug.ui.actions.IToggleBreakpointsTarget"/>
    </factory> 
</extension>
```
If that's the case - did you look into more lightweight solutions?

cdietrich:
@meysholdt we did not find any. see the discussion in the eclipse bugs as well

kthoms:
@meysholdt It is required for all languages deriving from Xbase, not all Xtext languages. And yes, it is required. And actually, for Xtend it was already the case that we needed to introduce an empty subclass to have a dedicated target for adapters for bug#505760,#434688.

If you have a concrete idea for a smarter solution, please let us know. Otherwise I would like to accept this approach.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ui/editor/EditorFragment2.xtend
```diff
@@ -0,0 +1,50 @@
...
+package org.eclipse.xtext.xtext.generator.ui.editor
+
+import com.google.inject.Inject
+import org.eclipse.xtext.xtext.generator.AbstractStubGeneratingFragment
+import org.eclipse.xtext.xtext.generator.XtextGeneratorNaming
+import org.eclipse.xtext.xtext.generator.model.FileAccessFactory
+import org.eclipse.xtext.xtext.generator.xbase.XbaseUsageDetector
+
+/**
+ * @author fstolte - Initial contribution and API
```
cdietrich: `@since 2.14` shhould be exported in manifest

cdietrich:
maybe we should restrict this to xbase languages?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ui/editor/EditorFragment2.xtend
FStolte: I added the package export to the manifest.

FStolte:
Limiting the generation of the language-specific editor class to xbase languages should easily be possible.

I am not sure if the language-specific editor might be useful for non-xbase languages in any way. If there is a conceivable use-case, then I would leave it as is since I also do not see any problems from having it generated. Otherwise, I would favor removing the language-specific editor from non-xbase languages.

cdietrich:
you need it if you use tracing and want to debug java, so it is a rare edge case

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ui/editor/EditorFragment2.xtend
cdietrich: @since is still missing @FStolte

FStolte:
After discussion with @kthoms, we arrived at the following solution: The editor class will be generated for all languages. It is generated into the src-gen folder, and if a concrete language needs to customize the editor, it can inherit from the generated one.
Since Xbase aleady has a custom editor, the generation is deactivated for it via the `generateStub` flag in the `EditorFragment`.

This version of the implementation is currently checked in.

cdietrich:
@meysholdt  as i told you folow the discussions in the bugzilla you did not dare to care about for the last 2-4 years https://bugs.eclipse.org/bugs/show_bug.cgi?id=471773 https://bugs.eclipse.org/bugs/show_bug.cgi?id=422633

szarnekow:
Patch looks good to me.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ui/editor/EditorFragment2.xtend
FStolte: Oh right, I missed that. @Since is now added.

meysholdt:
Hi Christian,
sorry for not having been more responsive on those bugzillas in the past. Unfortunately, non of these bugzillas mentions or discusses a more simple solution for this issue. 
The fact that I proposed this as a workaround does not mean it's the best solution.

cdietrich:
the more simply solution would be:

we introduce a new registry/extension (binding) point that allows us the provide actual language specific implementations of the code. we hope that nobody does manually what we do inside the fix.
then we change the current impl to ask the currrent editor for its language, and then query our registry for that impl

szarnekow:
@meysholdt Seriously, what is not simple with the solution that is provided? It fits nicely with the Eclipse Adapter framework, does not contain any surprises and a language specific editor class helps with debugging, too.

FStolte:
Since the issue directly referred to this solution approach, I did not spend much time on thinking about alternatives.

However, considering how long this issue has been around without a more elegant solution coming up (and the fact that this workaround was already in place for Xtend), I think that it is reasonable to implement this solution.

cdietrich:
we could consider to move this to xbaseeditor itself. but i am not sure if this will work

meysholdt:
I was hoping you guys had investigated a solution that is not based on generating even more code.

e.g. the `org.eclipse.core.runtime.IAdapterFactory` has access to the editor instance and could do a language-specific dispatch.

cdietrich:
yes but letting a random instance doing the dispatch is bad.
and moving to getAdapter in XbaseEditor would break existing clients.

szarnekow:
@meysholdt Yes, there are more involving solutions that avoid generating 6 lines of code. All of them have their own drawbacks. Do you approve this PR or not?

meysholdt:
thx @cdietrich, that's roughly the list I was asking for. I also think that it would have been better if it would have been implemented in `xbaseditor.getAdapter()` in the first place.

The Eclipse Adapter Framework has this limitation that there can only be one factory for every (`adaptable.class`, `adapter.class`) combination which hits us in Guice-configured environments. 

One questions remains: Why have this empty generated editor class for languages that aren't debuggable and therefore don't need an `StratumBreakpointAdapterFactory`? That's probably the big majority of languages.

meysholdt:
anything else besides:
> language specific editor class helps with debugging
?

cdietrich:
the question: does it hurt for non xbase languages.
we can generate the subclasses fo xbase langs only

meysholdt:
> the question: does it hurt for non xbase languages.

cons that come to my mind:
- it forces people that actually subclass the editor to break with the naming convention MyLanugageXtextEditor because that name is now occupied by the generated super-class.
- it's another file that needs to be compiled, index, appears in inheritance hierarchies
- the class doesn't do anything. 

pro: 
- consistency across Xtext projects, fewer switches in the generator
- the class is there in case maybe at some point somebody wants to register an adapter factory for it

neutral:
- debugging. A better debugging experience can also be provided by meaningful toString() impls.

Issue #544, opened 2017-11-15
Traced generation - Configurable issue code is not registered


NDuemmel:
Hello,

I am writing a DSL which is generated to Java.
Whenever I save my DSL file I get the following error:

> 498710 [Worker-20] ERROR org.eclipse.xtext.validation.IssueSeverities  - Configurable issue code 'org.eclipse.xtext.builder.copyJavaProblems' is not registered. Check the binding for org.eclipse.xtext.validation.ConfigurableIssueCodesProvider
> 
> !ENTRY org.apache.log4j 4 0 2017-11-14 15:33:48.136
> !MESSAGE org.eclipse.xtext.validation.IssueSeverities  - Configurable issue code 'org.eclipse.xtext.builder.copyJavaProblems' is not registered. Check the binding for org.eclipse.xtext.validation.ConfigurableIssueCodesProvider

Everything seems to work fine however, the Java code is generated.
If several classes are generated I get this error several times.
For example the following code generates two classes:

> type Point : structure {
> 	attribute x : integer;
> 	attribute y : integer;
> }
> 
> main {	
> 	variable p : Point;
> 	p.x <-- 2; 
> 	p.y <-- 3; 
> 	_output(p);
> }   

On save I get following errors:

> 756293 [Worker-20] ERROR org.eclipse.xtext.validation.IssueSeverities  - Configurable issue code 'org.eclipse.xtext.builder.copyJavaProblems' is not registered. Check the binding for org.eclipse.xtext.validation.ConfigurableIssueCodesProvider
> 
> !ENTRY org.apache.log4j 4 0 2017-11-14 15:38:05.712
> !MESSAGE org.eclipse.xtext.validation.IssueSeverities  - Configurable issue code 'org.eclipse.xtext.builder.copyJavaProblems' is not registered. Check the binding for org.eclipse.xtext.validation.ConfigurableIssueCodesProvider
> 
> 756298 [Worker-20] ERROR org.eclipse.xtext.validation.IssueSeverities  - Configurable issue code 'org.eclipse.xtext.builder.copyJavaProblems' is not registered. Check the binding for org.eclipse.xtext.validation.ConfigurableIssueCodesProvider
> 
> !ENTRY org.apache.log4j 4 0 2017-11-14 15:38:05.714
[...]

PR #563, opened 2017-11-30
moved 'copy java problems' configurable issue code to xtext core 

cdietrich:
moved 'copy java problems' configurable issue code to xtext core 
https://github.com/eclipse/xtext-core/issues/544

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/validation/ConfigurableIssueCodesProvider.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/validation/ConfigurableIssueCodesProvider.java
@@ -7,22 +7,53 @@
  *******************************************************************************/
 package org.eclipse.xtext.validation;
 
-import java.util.Collections;
 import java.util.Map;
 
 import org.eclipse.xtext.preferences.PreferenceKey;
+import org.eclipse.xtext.util.IAcceptor;
+
+import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/validation/ConfigurableIssueCodesProvider.java
```diff
@@ -7,22 +7,53 @@
...
  * @author Sven Efftinge - Initial contribution and API
  * @since 2.4
  */
 public class ConfigurableIssueCodesProvider {
+	
+	private Map<String, PreferenceKey> issueCodes;
 
+	public ConfigurableIssueCodesProvider() {
+		final Map<String, PreferenceKey> map = Maps.newLinkedHashMap();
+		initialize(new IAcceptor<PreferenceKey>() {
```
szarnekow: I wonder if there is a use-case where we would want to use another acceptor. Maybe a factory method should be present, e.g. the acceptor would be obtained via `createAcceptor(map)`

cdietrich: we did not need that in the project where we customized this. we simpy did the copy&paste that is now no longer needed

szarnekow: ok

Issue #359, opened 2017-06-01
[tracing] NPE thrown when a file is generated without traces


spoenemann:
Using the GeneratorNodeProcessor to serialize a generator tree without any TraceNode leads to
```
java.lang.NullPointerException
	at org.eclipse.xtext.builder.EclipseResourceFileSystemAccess2.updateTraceInformation(EclipseResourceFileSystemAccess2.java:422)
	at org.eclipse.xtext.builder.EclipseResourceFileSystemAccess2.generateFile(EclipseResourceFileSystemAccess2.java:269)
	at org.eclipse.xtext.builder.EclipseResourceFileSystemAccess2.generateFile(EclipseResourceFileSystemAccess2.java:219)
```
Because an implementation of ITraceRegionProvider must not return null when asked for its trace region.

The Javadoc of ITraceRegionProvider says it should throw a TraceNotFoundException if no trace can be found. However, this case is not considered in EclipseResourceFileSystemAccess2.

cdietrich:
other clients like

`org.eclipse.xtext.generator.JavaIoFileSystemAccess.generateTrace(String, String, CharSequence)`

ignore this contract as well.

cdietrich:
@szarnekow what do you think? and can we still change the contract?

szarnekow:
I'd go for consistently throwing a TraceNotFoundException in the different implementations of ITraceRegionProvider. EclipseResourceFileSystemAccess2.updateTraceInformation could simply be adjusted to handle that contract correctly.

andreasmuelder:
+1

cdietrich:
@szarnekow how to handle the exception correct? can the trace file be empty? write no trace file/delte it?

cdietrich:
Fixed in Xtext 2.14

PR #564, opened 2017-12-01
[tracing] NPE thrown when a file is generated without traces

cdietrich:
[tracing] NPE thrown when a file is generated without traces 
https://github.com/eclipse/xtext-core/issues/359

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/generator/JavaIoFileSystemAccess.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/generator/JavaIoFileSystemAccess.java
@@ -123,17 +123,20 @@ public class JavaIoFileSystemAccess extends AbstractFileSystemAccess2 {
 		try {
 			if (contents instanceof ITraceRegionProvider) {
 				String traceFileName = traceFileNameProvider.getTraceFromJava(generatedFile);
-				File traceFile = getFile(traceFileName, outputConfigName);
-				OutputStream out = new BufferedOutputStream(new FileOutputStream(traceFile));
+				OutputStream out = null;
 				try {
 					AbstractTraceRegion traceRegion = ((ITraceRegionProvider) contents).getTraceRegion();
+					File traceFile = getFile(traceFileName, outputConfigName);
+					out = new BufferedOutputStream(new FileOutputStream(traceFile));
 					traceSerializer.writeTraceRegionTo(traceRegion, out);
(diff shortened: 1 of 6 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/generator/JavaIoFileSystemAccess.java
```diff
@@ -129,7 +130,9 @@ protected void generateTrace(String generatedFile, String outputConfigName, Char
 					traceSerializer.writeTraceRegionTo(traceRegion, out);
 					if(callBack != null) 
 						callBack.fileAdded(traceFile);
-				} finally {
+				} catch (TraceNotFoundException e) {
```
spoenemann: `getTraceRegion()` should be called before the output stream is created in order to avoid opening the trace file unnecessarily.

org.eclipse.xtext/src/org/eclipse/xtext/generator/URIBasedFileSystemAccess.xtend
```diff
@@ -93,9 +94,13 @@ class URIBasedFileSystemAccess extends AbstractFileSystemAccess2 {
 		if (isGenerateTraces && contents instanceof ITraceRegionProvider) {
 			var String traceFileName = traceFileNameProvider.getTraceFromJava(generatedFile)
 			val out = new ByteArrayOutputStream()
-			var AbstractTraceRegion traceRegion = (contents as ITraceRegionProvider).getTraceRegion()
-			traceRegionSerializer.writeTraceRegionTo(traceRegion, out)
-			generateFile(traceFileName, outputConfigName, new ByteArrayInputStream(out.toByteArray))
+			try {
+				var AbstractTraceRegion traceRegion = (contents as ITraceRegionProvider).getTraceRegion()
```
spoenemann: Same as above: call `getTraceRegion()` first to avoid unnecessary computations.

cdietrich:
updated the pr? ok so now?

org.eclipse.xtext/src/org/eclipse/xtext/generator/URIBasedFileSystemAccess.xtend
szarnekow: I agree - it would make sense to postpone `getTraceFromJava` and do it only if there is a traceRegion.

Issue #520, opened 2017-10-28
[LSP] Rename not working accross files


cdietrich:
Given the Grammar

```
grammar org.xtext.example.mydsl.MyDsl with org.eclipse.xtext.common.Terminals

generate myDsl "http://www.xtext.org/example/mydsl/MyDsl"

Model:
	greetings+=Greeting*;
	
Greeting:
	'Hello' name=ID ('from' from=[Greeting])? '!';

```

and test models
`a.mydsl`
```
Hello Xtext2!
Hello VSCode from Xtext2!
Hello ThisFile from Other!
Hello you!
```

and 
`b.mydsl`
```
Hello Other!
```

renaming `Other` to `OtherNew` does not work

```
[Trace - 12:41:11] Sending request 'textDocument/hover - (37)'.
Params: {
    "textDocument": {
        "uri": "file:///Users/dietrich/git/xtext-languageserver-example/demo/b.mydsl"
    },
    "position": {
        "line": 1,
        "character": 11
    }
}
```

```
[Trace - 12:41:18] Received response 'textDocument/rename - (38)' in 4982ms.
Result: {
    "changes": {
        "file:///Users/dietrich/git/xtext-languageserver-example/demo/b.mydsl": [
            {
                "range": {
                    "start": {
                        "line": 1,
                        "character": 6
                    },
                    "end": {
                        "line": 1,
                        "character": 11
                    }
                },
                "newText": "OtherNeu"
            }
        ],
        "file:///Users/dietrich/git/xtext-languageserver-example/demo/a.mydsl": [
            {
                "range": {
                    "start": {
                        "line": 2,
                        "character": 20
                    },
                    "end": {
                        "line": 2,
                        "character": 25
                    }
                },
                "newText": "Other"
            }
        ]
    }
}
```

as you can see ` "newText": "Other"` still produces the old value (coming from `org.eclipse.xtext.ide.serializer.impl.ReferenceUpdater.updateReference(ITextRegionDiffBuilder, IUpdatableReference)`)

the scope has still the old entry. so searching the object there will result in the old value

standalone tests might have missed this cause the dont user Serializable Resource Descriptions

cdietrich:
cc @meysholdt @JanKoehnlein

cdietrich:
added test with b43934a0efdde29a492cbba7713f1b19715b3327

cdietrich:
@JanKoehnlein can this issue be closed now?

JanKoehnlein:
Fixed

PR #567, opened 2017-12-06
GH-520 more tests for live scoping

JanKoehnlein:

Here are more tests and a fix for new containers, as wished by @szarnekow .

Note that I am not aware of any EMF contract that defines what happens to the persistent state when you rename / delete a resource form a resource set. So these tests can only be done on the local state, i.e. I cannot rename / delete a resource form the shadowed global state by modifying the local resource set. Correct me if I am wrong, but AFAIK this doesn't work in the existing live scopes either.


JanKoehnlein:
See #568

szarnekow:
Thank you. I'm using the wrong words apparently :D Would it make sense to add these additional resources to the resource-move tests, too?

JanKoehnlein:
https://github.com/eclipse/xtext-core/pull/569

Issue #570, opened 2017-12-06
Missing error on invalid Override in grammar


szarnekow:
`@Override` in a grammar without a super grammar does not produce any errors / warnings.

dhuebner:
Should be an error

szarnekow:
Thanks for clarifying.

dhuebner:
You are welcome :)

cdietrich:
fixed in 2.14

PR #578, opened 2017-12-13
[#570] Added error message for invalid override

FStolte:
Fix for [#570]: Rules with an @Override annotation in a grammar without super grammar
now correctly display an error.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
@@ -1255,7 +1255,14 @@ public class XtextValidator extends AbstractDeclarativeValidator {
 		final boolean isOverride =
 				rule.getAnnotations().stream().anyMatch(e -> AnnotationNames.OVERRIDE.equals(e.getName()));
 		
-		for (Grammar g : GrammarUtil.getGrammar(rule).getUsedGrammars()) {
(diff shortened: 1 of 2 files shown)


cdietrich:
thx for the contribution. could you provide a test for this as well. see `XtextValidationTest`

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
```diff
@@ -1255,6 +1255,11 @@ public void checkOverridingRule(AbstractRule rule) {
 		final boolean isOverride =
 				rule.getAnnotations().stream().anyMatch(e -> AnnotationNames.OVERRIDE.equals(e.getName()));
 		
+		if (GrammarUtil.getGrammar(rule).getUsedGrammars().isEmpty() && isOverride) {
```
szarnekow: Nitpicking: Why do you dublicate `GrammarUtil.getGrammar(rule).getUsedGrammars()` instead of using a local var? It would also make sense to guard with `if (isOverride)` before doing the other stuff. Both are minor things, though.

FStolte: I will add the local variable, but I am not sure what you mean with "the other stuff" that should be guarded by `isOverride`. The for-loop that comes after my contribution contains a case for `!isOverride`, so such a guard does not apply to it.

FStolte:
I added a test for the new error message.

cdietrich:
thx @FStolte

Issue #582, opened 2017-12-13
[J9] Wizard + gradle build: update to  1.0.20 of xtext-gradle-plugin


cdietrich:

cdietrich:
update the xtext gradle plugin the wizard creates to 1.0.20
update the tests
update the version used in xtext gradle files

cdietrich:
fixed in 2.14

PR #584, opened 2017-12-13
[#582] Updated version of xtext-gradle-plugin

FStolte:
Related to issue [eclipse/xtext-core#582]
- updated from 1.0.19 to 1.0.20
- updated the version in the gradle file
- updated the version used by the wizard
- updated the version in one example

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -8,7 +8,7 @@ ext.versions = [
 	'xtext': version,
 	'xtext_bootstrap': '2.13.0',
 	'gradle_plugins': '0.1.0',
-	'xtext_gradle_plugin': '1.0.19',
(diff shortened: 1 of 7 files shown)


cdietrich:
please adjust all the tests as well

FStolte:
For some reason I did not find the other tests using the old xtext-gradle-plugin version yesterday. I will adjust them now.

FStolte:
Ok, now all tests should be covered.

gradle/versions.gradle
```diff
@@ -1,29 +1,29 @@
-/*
- * Base project version and versions of common dependencies.
- */
-
-version = '2.14.0-SNAPSHOT'
-
```
cdietrich: am not sure which kind of line endings we want to have here @szarnekow @kthoms any idea

PR #599, opened 2018-01-05
[xtext-eclipse#145] Create new implementations in core.

ArneDeutsch:
Move implementation from CopyQualifiedNameService and
DefaultCopyQualifiedNameService to xtext core plugin.

Change-Id: I3ffdbbcda641cc4f70396f4274e42934e4248278
Signed-off-by: Arne Deutsch <arne@idedeluxe.com>

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/naming/DefaultCopyQualifiedNameService.xtend
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/naming/DefaultCopyQualifiedNameService.xtend
@@ -1,0 +1,76 @@
+/*******************************************************************************
+ * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.naming
+
+import com.google.inject.Inject
+import java.util.List
+import org.eclipse.emf.ecore.EObject
+import org.eclipse.xtext.naming.IQualifiedNameConverter
+import org.eclipse.xtext.naming.IQualifiedNameProvider
+import org.eclipse.xtext.naming.QualifiedName
+
+/**
+ * Default implementation for {@link ICopyQualifiedNameService}. Clients might use this implementation as base class.
+ * 
+ * @author Anton Kosyakov - Initial contribution and API
+ * @author Arne Deutsch - Moved to new package
+ * @since 2.14
+ */
+class DefaultCopyQualifiedNameService implements ICopyQualifiedNameService {
+
+	@Inject
+	IQualifiedNameProvider qualifiedNameProvider
+
+	@Inject
+	IQualifiedNameConverter qualifiedNameConverter
+
+	def dispatch getQualifiedName(EObject it, EObject context) {
+		toFullyQualifiedName
+	}
+
+	def dispatch getQualifiedName(EObject it, Void context) {
+		toFullyQualifiedName
+	}
+
+	def protected dispatch getQualifiedName(Void it, EObject context) {
+		null
+	}
+
+	def protected dispatch getQualifiedName(Void it, Void context) {
+		null
+	}
+
+	def protected <T> toQualifiedNames(List<T> it, (T)=>String toQualifiedNameFunction) {
+		if (it === null || size == 0) {
+			return ""
+		}
+		'''«FOR element : it SEPARATOR ', '»«toQualifiedNameFunction.apply(element)»«ENDFOR»'''
+	}
+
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/naming/CopyQualifiedNameService.java
```diff
@@ -0,0 +1,23 @@
...
+import org.eclipse.emf.ecore.EObject;
+
+/**
+ * Replacement for org.eclipse.xtext.ui.editor.copyqualifiedname.CopyQualifiedNameService
+ * 
+ * @author Anton Kosyakov - Initial contribution and API
+ * @author Arne Deutsch - Moved to new package
+ * @since 2.14
+ */
+public interface CopyQualifiedNameService {
```
cdietrich: @szarnekow when do we start interfaces with `I` and when don't we.
how was this decided in the past?

org.eclipse.xtext/src/org/eclipse/xtext/naming/CopyQualifiedNameService.java
```diff
@@ -0,0 +1,23 @@
...
+
```
kthoms: when becoming public API the interface has to be documented

org.eclipse.xtext/src/org/eclipse/xtext/naming/CopyQualifiedNameService.java
```diff
@@ -0,0 +1,23 @@
...
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.naming;
+
+import org.eclipse.emf.ecore.EObject;
+
+/**
+ * Replacement for org.eclipse.xtext.ui.editor.copyqualifiedname.CopyQualifiedNameService
```
kthoms: Document the interface
Remove "Replacement for..."

org.eclipse.xtext/src/org/eclipse/xtext/naming/CopyQualifiedNameService.java
kthoms: Please add prefix "I" as it is most common

org.eclipse.xtext/src/org/eclipse/xtext/naming/CopyQualifiedNameService.java
```diff
@@ -0,0 +1,23 @@
+/*******************************************************************************
+ * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.naming;
```
kthoms: put into a subpackage naming.copy
otherwise the classes would be automatically public and I'd like to keep them restricted API as before

org.eclipse.xtext/src/org/eclipse/xtext/naming/CopyQualifiedNameService.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Replacement for org.eclipse.xtext.ui.editor.copyqualifiedname.CopyQualifiedNameService
+ * 
+ * @author Anton Kosyakov - Initial contribution and API
+ * @author Arne Deutsch - Moved to new package
+ * @since 2.14
+ */
+public interface CopyQualifiedNameService {
+
+	String getQualifiedName(EObject selectedElement, EObject context);
```
kthoms: The context parameter is not used so far in the implementation and I doubt that it makes sense at all. Please remove the parameter here and add a default method for the deprecated interface

szarnekow: The context parameter allows to support use cases with aliases. Assuming you have an element with two different properties that can be names, so it has two qualified names, let's say an IP address and a URL.
When you refer to that object by ip, you probably want to copy the IP on copy QN. When you refer to it via text, you probably want to copy the URL. The context parameter is passed into the service to allow this distinction.

org.eclipse.xtext/src/org/eclipse/xtext/naming/CopyQualifiedNameService.java
szarnekow: There is no strict rule throughout the code base, but heuristically it's better use the I prefix.

org.eclipse.xtext/src/org/eclipse/xtext/naming/CopyQualifiedNameService.java
szarnekow: I doubt that we need a subpackage here. The API was stable since a couple of years already.

Issue #608, opened 2018-01-09
local test failures in  org.eclipse.xtext.parser.indentation.tests when called gradle


cdietrich:
i have local test failures in  org.eclipse.xtext.parser.indentation.tests when called gradle.

cdietrich:
```
com.google.inject.ProvisionException: Guice provision errors:

1) Error injecting constructor, org.eclipse.emf.common.util.WrappedException: java.lang.RuntimeException: Cannot create a resource for 'classpath:/org/eclipse/xtext/parser/indentation/IndentationAwareTestLanguage.xtextbin'; a registered resource factory is needed
  at org.eclipse.xtext.common.services.TerminalsGrammarAccess.<init>(TerminalsGrammarAccess.java:31)
  at org.eclipse.xtext.common.services.TerminalsGrammarAccess.class(TerminalsGrammarAccess.java:31)
  while locating org.eclipse.xtext.common.services.TerminalsGrammarAccess
    for parameter 1 at org.eclipse.xtext.parser.indentation.services.IndentationAwareTestLanguageGrammarAccess.<init>(IndentationAwareTestLanguageGrammarAccess.java:174)
  at org.eclipse.xtext.parser.indentation.services.IndentationAwareTestLanguageGrammarAccess.class(IndentationAwareTestLanguageGrammarAccess.java:174)
  while locating org.eclipse.xtext.parser.indentation.services.IndentationAwareTestLanguageGrammarAccess
  while locating org.eclipse.xtext.IGrammarAccess
    for field at org.eclipse.xtext.linking.lazy.LazyLinker.grammarAccess(LazyLinker.java:300)
  while locating org.eclipse.xtext.linking.lazy.LazyLinker
  while locating org.eclipse.xtext.linking.ILinker
    for field at org.eclipse.xtext.resource.XtextResource.linker(XtextResource.java:461)
  while locating org.eclipse.xtext.linking.lazy.LazyLinkingResource
  while locating org.eclipse.xtext.resource.XtextResource
Caused by: org.eclipse.emf.common.util.WrappedException: java.lang.RuntimeException: Cannot create a resource for 'classpath:/org/eclipse/xtext/parser/indentation/IndentationAwareTestLanguage.xtextbin'; a registered resource factory is needed
	at org.eclipse.xtext.parser.BaseEPackageAccess.loadResource(BaseEPackageAccess.java:57)
```

cdietrich:
java main showing (maybe) the cause

```
[...]

PR #616, opened 2018-01-15
[#608] Fixed reuse issue with WrappingInjectorProvider

cdietrich:
[#608] make sure that WrappingInjectorProvider records changes to the global registry even if the delegate itself is a IRegistryConfigurator and its injector is already created so that createInjector does not do anything that could be recorded


org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/smoketest/internal/WrappingInjectorProvider.java
```diff
@@ -45,8 +45,14 @@ public WrappingInjectorProvider(IInjectorProvider delegate) {
 		this.delegate = delegate;
 		stateBeforeInjectorCreation = GlobalRegistries.makeCopyOfGlobalState();
 		this.injector = createInjector();
+		if (delegate instanceof IRegistryConfigurator) {
```
cdietrich, line 48: is it correct to call this AFTER createInjector?

cdietrich: nope inverting also seems to break other stuff

Issue #550, opened 2017-11-16
AbstractInternalAntlrParser.createAllRules should be cached.


cdietrich:
AbstractInternalAntlrParser.createAllRules is done for each and every time a parser is created.
so in a code base with many files and bigger grammars it might lead to a significant decrease in parsing performance. e.g. when parsing 10000 empty xtend classes it is consuming nearly 50% of the parse time. (3.3s of 7.8s measured on my machine).
we should consider make this cachable by default.

cdietrich:
@szarnekow is the way we create the parsers considered to be stable or can we rework the parser creation.

cdietrich:
we have usecases with 400k model files. in this case we win about 200sec on this

cdietrich:
Fixed in 2.14

PR #624, opened 2018-01-26
[#550] Cache AbstractInternalAntlrParser.createAllRules

cdietrich:
[#550] Cache AbstractInternalAntlrParser.createAllRules


knutwannheden:
OK, thanks. I just wanted to make sure we didn't introduce some overlap here.

cdietrich:
if i knew what other place two weeks ago (or 5 month when i actually did the change) i would have unified it but now after the release its too late unfortunately

knutwannheden:
Makes sense, unfortunately :-)

Issue #627, opened 2018-01-30
Wizard cannot compute Xtext version in dev mode


kthoms:
Steps to reproduce:
- Create a new Xtext project with Maven build
- in the parent pom.xml there is the entry `<xtext-version>unspecified</xtext-version>`

It is expected that the `unspecified` is replaced by the current version with `-SNAPSHOT` qualifier. For releases, the qualifier is stripped.

It should be possible to derive the right version from the Manifest's Bundle-Version attribute.

kthoms:
Yes, but it should also work in devenv, shouldn't it?

cdietrich:
yes but you should change this only if Maven-Version returns unspecfied for the reasons i explained in the pr

cdietrich:
Fixed in 2.14

PR #628, opened 2018-01-30
[#627] Compute Xtext version from Bundle-Version

kthoms:
In development mode Xtext read the version from the 'Maven-Version'
attribute from org.eclipse.xtext.util manifest, which has a static value
of 'unspecified'. This gets replaced in a Gradle build, but does not
work when using the bundle in an Eclipse development environment.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.xtend
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.xtend
@@ -84,6 +84,7 @@ class XtextVersion {
 					return version.substring(0, version.lastIndexOf('.'))
 				}
 			}
+			return version
 		} catch (Exception e) {
 			return null;
 		} finally {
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.xtend
```diff
@@ -72,7 +72,13 @@ class XtextVersion {
 			val url = new URL(Plugin.INSTANCE.baseURL + 'META-INF/MANIFEST.MF')
 			is = url.openStream()
 			val manifest = new Manifest(is)
-			return manifest.getMainAttributes().getValue('Maven-Version')
+			val bundleVersion = manifest.getMainAttributes().getValue('Bundle-Version')
```
cdietrich: i am not sure what the reasoning behind this was. and this wont work for nightlys. their bundle version looks like `/Users/dietrich/Downloads/EclipseOxygenXtext214Snapshots.app/Contents/Eclipse/plugins/org.eclipse.xtext_2.14.0.v20180126-1316.jar`

so maybe you can introduce looking at BundleVersion as fallback if Maven version returns unspecified

cdietrich: same for release candidates/milestones

kthoms:
Updated. Bundle-Version is only taken as fallback in dev mode.

cdietrich:
please fix failing tests

cdietrich:
@kthoms see my changes

kthoms:
Ah thanks!

PR #633, opened 2018-02-08
[refactoring] Deprecate methods with correspondence in Guava

kthoms:
Also refactored methods to use Guava

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2008, 2018 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -8,14 +8,11 @@
  *******************************************************************************/
 package org.eclipse.xtext.util;
 
-import java.io.BufferedInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileFilter;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
-import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
```diff
@@ -139,64 +149,30 @@ public static String readFileIntoString(String filename) {
...
-			byte[] buffer = new byte[2048];
-			int bytesRead = 0;
-			StringBuffer b = new StringBuffer();
-			do {
-				bytesRead = source.read(buffer);
-				if (bytesRead != -1)
-					b.append(new String(buffer, 0, bytesRead));
-			} while (bytesRead != -1);
-			return b.toString();
+			return new String(ByteStreams.toByteArray(inputStream));
```
szarnekow: No longer buffered

org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
```diff
@@ -139,64 +149,30 @@ public static String readFileIntoString(String filename) {
...
-			byte[] data = new byte[16384];
-	
-				while ((nRead = inputStream.read(data, 0, data.length)) != -1) {
-				  buffer.write(data, 0, nRead);
-				}
-	
-			buffer.flush();
-	
-			return buffer.toByteArray();
+			return ByteStreams.toByteArray(inputStream);
```
szarnekow: Please add buffering to the inputStream if it is not yet buffered

org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
```diff
@@ -139,64 +149,30 @@ public static String readFileIntoString(String filename) {
 	
 	/**
 	 * @since 2.3
+	 * @deprecated Use com.google.common.io.ByteStreams.toByteArray(InputStream) instead
```
szarnekow: The deprecation comment is not valid if we do the additional buffering

kthoms:
Thanks for these comments. Will change.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
kthoms: toByteArray calls copy(in,out), which does buffering
com.google.common.io.ByteStreams.copy(InputStream, OutputStream)

Am I missing something?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
kthoms: is buffered by toByteArray/copy, right?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
szarnekow: I suspect that a buffered input stream with a buffer size of 16k is the better choice than a custom buffering in ByteStreams.copy with a buffer size of 4k. But that is only an assumption that would need to be verified.

kthoms: Actually there are no references to this method. Also an indicator to deprecate it?

szarnekow: Yes

org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
szarnekow: true

org.eclipse.xtext.util/src/org/eclipse/xtext/util/Files.java
szarnekow: My comment is obsolete.

cdietrich:
@szarnekow  what is the state of review?

kthoms:
Thx, @szarnekow

Issue #638, opened 2018-02-14
[LanguageServer] FormattingOptions are ignored on format request


xcariba:
in `FormattingService` on format request `DocumentFormattingParams` are not used at all. It is useful to use `tabSize` value in formatter.

cdietrich:
Fixed in Xtext 2.14

PR #642, opened 2018-02-20
[#638] Use Formatting options passed by the client

cdietrich:
[#638] Use Formatting options passed by the client


kthoms:
I assume that reflection is necessary again here and I don't have to mention that we have to reduce usage to a bare minimum. How do we assure that we detect breakage? I guess the unit tests would fail then?

cdietrich:
@kthoms reflection is used to maintain api. and no i could write an addional unit test but dont have yet

cdietrich:
@kthoms and if you break the reflection you break the api
so it does not really matter

PR #651, opened 2018-03-05
startsWith: return false for null argument

kthoms:
Avoid NPE when accidently invoked with null

Or shall we throw IllegalArgumentException?


szarnekow:
It’ll likely cause follow up exceptions in client code. What’s wrong with an NPE?

kthoms:
It looks like that case is not expected and actually is not a bug in Xtext, but in the client code. A NPE would also be OK is we provide an additional message to indicate the erraneous call to the client.

I felt that returning false for a null argument could be also valid, since no QN is starting with null.

kthoms:
Updated: Throw IAE by using Preconditions.checkArgument

PR #663, opened 2018-03-21
[361216] Improved logic that registers scope wrappers

szarnekow:


org.eclipse.xtext.tests/src/org/eclipse/xtext/scoping/impl/DelegatingScopeProviderTest.java
```diff
@@ -0,0 +1,159 @@
...
+import org.junit.Test;
+
+/**
+ * @author Sebastian Zarnekow - Initial contribution and API
+ */
+public class DelegatingScopeProviderTest implements IScopeWrapper {
+
+	static class TestableDelegatingScopeProvider implements IDelegatingScopeProvider {
+
+		private IScopeProvider delegate;
```
cdietrich: final?

szarnekow: Done

Issue #666, opened 2018-03-22
JUnit 5 support


kthoms:
Xtext's testing framework should support writing test for JUnit 5.

Most notably is that the `@RunWith` annotation is not available anymore and [has been replaced by the more flexible `@ExtendWith` annotation](http://www.baeldung.com/junit-5-runwith). However, we still need the logic that is employed by `XtextRunner`, and preserve `@InjectWith`.

This basically means that `XtextRunner`'s logic has to be transferred to the new JUnit 5 extension API. This concept should be added to `org.eclipse.xtext.testing`, introducing new dependencies on the JUnit 5 framework.

Clients may use either JUnit 4 or 5, or for transition both. Since JUnit 5 API is providing different packages it is possible to manage both.

As a result clients should be able to leverage JUnit 5 to write and execute Xtext based tests with typical dependency injection.

kthoms:
CQ for JUnit 5.0 usage, approved: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15984

kthoms:
I think we are done. Documentation, of course, must be added.

kthoms:
The test projects do not need direct dependencies on the JUnit engine bundles. Tycho is smart enough to use these dependencies at runtime. It even harms, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=534587 for that.

Instead, the test projects only need dependencies on the JUnit API. When using the JUnit library container in the project configuration the required bundles are automatically inserted, including the engine bundles for JU5.

JU4
<img width="793" alt="screenshot 101" src="https://user-images.githubusercontent.com/265597/40128066-0c2e40fc-5931-11e8-83d2-da5414837470.png">

JU5
<img width="786" alt="screenshot 103" src="https://user-images.githubusercontent.com/265597/40128072-0ed8a874-5931-11e8-832b-6ace358cd3ef.png">

cdietrich:
Junit 5 tests are not executed in pure maven projects

cdietrich:
Junit5 test are NOT executed in pure gradle projects

cdietrich:
for gradle the following seems to work

```
dependencies {
	compile project(':org.xtext.example.mydsl1')
	testCompile "org.junit.jupiter:junit-jupiter-api:5.1.0"
	testCompile "org.junit.jupiter:junit-jupiter-params:5.2.0"
	testCompile "org.eclipse.xtext:org.eclipse.xtext.testing:${xtextVersion}"
[...]

PR #670, opened 2018-03-23
[#666] JUnit 5 support

kthoms:
- provide new InjectionExtension as replacement for XtextRunner
- added unit test for InjectionExtension
- added JUnit 5.0 dependencies


org.eclipse.xtext.testing/META-INF/MANIFEST.MF
```diff
@@ -8,6 +8,7 @@ Bundle-Vendor: Eclipse Xtext
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Export-Package: org.eclipse.xtext.testing,
  org.eclipse.xtext.testing.builder,
+ org.eclipse.xtext.testing.extensions,
```
cdietrich: export as internal?

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/extensions/InjectionExtension.java
```diff
@@ -0,0 +1,86 @@
...
+ * tests.
+ * <p>
+ * The extension retrieves an {@link Injector} to inject members in the instance
+ * under test before best execution and resets registries after execution.
+ * </p>
+ * 
+ * @author Karsten Thoms - Initial contribution and API
+ * @since 2.14
+ */
+public class InjectionExtension implements BeforeTestExecutionCallback, AfterTestExecutionCallback {
```
cdietrich: mark as beta?

cdietrich:
@kthoms can you create and push downstream branches to test?
`gitAll`and `adjustBuildPipelines` in umbrella are your friends for that.

kthoms:
Good points, will update. Surely we need to adjust target definitions. Will come soon.

szarnekow:
What's the status of this?

gradle/versions.gradle
```diff
@@ -26,5 +26,6 @@ ext.versions = [
 	'antlr': '3.2',
 	'asm': '6.0',
 	'icu': '52.1',
-	'junit': '4.12'
+	'junit': '4.12',
+	'junit5': '5.0.0'
```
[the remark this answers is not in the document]
kthoms: JDT Photon will be bundled with 5.1. We are allowed to do so also now, since we have an approved CQ. Will update now.

org.eclipse.xtext.testing/META-INF/MANIFEST.MF
```diff
@@ -22,6 +23,7 @@ Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.junit;bundle-version="4.12.0",
  org.eclipse.xtend.lib
 Import-Package: org.apache.log4j;version="1.2.15",
- org.apache.log4j.spi;version="1.2.15"
+ org.apache.log4j.spi;version="1.2.15",
+ org.junit.jupiter.api.extension;version="5.0.0";resolution:=optional
```
cdietrich: still 5.0.0?

cdietrich: wil you do a pr for umbrella (repackageing of orbit) too

kthoms: I think the lower bound of 5.0.0 is correct here. There are no limitations rgd. usage of JUnit 5.0.0, only that we build against 5.1. The bundle is tested against 5.0, though.

cdietrich: am note sure how we test

kthoms: InjectionExtensionTest tests the new extension, or what kind of test do you think of? The lower bound just says that there is no restriction on something that 5.1 added, and works also for 5.0. And the InjectionExtensionTest can be executed also when 5.0.0 is stated in the versions.gradle.

kthoms:
The new commit extends the project configuration and generator to support JUnit 5. 

Since the JUnit version is configurable the fragment Junit4Fragment2 seems not to be the right name anymore. Created JUnitFragment now, copied code from the old fragment, derive the old fragment from the new, deprecated the old one and changed StandardLanguage to the new one. Since it is a base class for Junit4Fragment2, it should still be valid when users have configured or customized `junitSupport` in their generator workflows.

To enable JUnit 5, the workflow has to be configured:
screenshot 20
Default will be JUnit 4, so no change to old behavior and no configuration in the workflow needed then.

When configured, the Test plugin will have JUnit 5 dependencies and MyDslParsingTest will be a JUnit 5 test case.
screenshot 21

When executing the test, it will be executed with the JUnit 5 engine:
screenshot 22

org.eclipse.xtext.testing/tests/org/eclipse/xtext/testing/tests/InjectionExtensionTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+	private static boolean injectorCreated = false;
+	private static boolean registrySaved = false;
+	private static boolean registryRestored = false;
+	
+	@Inject
+	private Boolean fieldsInjected = false;
+
+	public static class MyInjectorProvider implements IRegistryConfigurator, IInjectorProvider {
+
```
szarnekow: Stylistic: There are quite some unnecessary blank lines in this file

org.eclipse.xtext.testing/tests/org/eclipse/xtext/testing/tests/InjectionExtensionTest.java
```diff
@@ -0,0 +1,95 @@
...
+	@Before
+	public void beforeShouldBeExecutedAfterTheRegistriesAreInitialized(){
+		assertTrue(registrySaved);
+		assertTrue(injectorCreated);
+		assertTrue(fieldsInjected);
+	}
+	
+	@Test
+	public void shouldSaveRegistriesBeforeCreatingAnInjector() {
+		// tests are performed in MyInjectorProvider
```
szarnekow: Should we inject the injector provider itself and assert on its internal state rather than with statics?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/JUnitVersion.java
```diff
@@ -0,0 +1,34 @@
...
+	private JUnitVersion (int version) {
+		this.version = version;
+	}
+	
+	public int getVersion () {
+		return version;
+	}
+	
+	public static JUnitVersion fromString (String version) {
+		switch (Integer.parseInt(version)) {
```
szarnekow: You could switch on the string itself

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/JUnitFragment.xtend
```diff
@@ -0,0 +1,262 @@
+package org.eclipse.xtext.xtext.generator.junit
```
szarnekow: Copyright header missing

org.eclipse.xtext.util/src/org/eclipse/xtext/util/JUnitVersion.java
```diff
@@ -0,0 +1,34 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.util;
+
+/**
+ * @author Karsten Thoms (karsten.thoms@itemis.de) - Initial contribution and API
+ * @since 2.14
+ */
+public enum JUnitVersion {
```
szarnekow: Please introduce a static DEFAULT that points to JUNIT_4 and use that on in the fragment.

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TestProjectDescriptor.xtend
```diff
@@ -44,14 +45,58 @@ abstract class TestProjectDescriptor extends ProjectDescriptor {
...
+				maven.groupId = "junit"
+				maven.artifactId = "junit"
+				maven.version = "4.12"
+				maven.scope = Scope.TESTCOMPILE
+			]
+		}
+		if (config.junitVersion == JUnitVersion.JUNIT_5) {
+			deps += new ExternalDependency()=>[
+				p2.bundleId = "org.junit.jupiter.api"
+				p2.version = "5.1.0"
```
szarnekow: This appears to be inconsistent with the 5.0.0 in the manifest

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/WizardConfiguration.xtend
```diff
@@ -32,6 +33,7 @@ class WizardConfiguration {
 	boolean needsGradleWrapper = true
 	JavaVersion javaVersion = JavaVersion.JAVA8
 	LanguageServer languageServer = LanguageServer.NONE
+	JUnitVersion junitVersion = JUnitVersion.JUNIT_4;
```
szarnekow: Can we check for Junit5 in the TP and use that one if available?

org.eclipse.xtext.testing/tests/org/eclipse/xtext/testing/tests/InjectionExtensionTest.java
kthoms: All copied from XtextRunnerTest

org.eclipse.xtext.testing/tests/org/eclipse/xtext/testing/tests/InjectionExtensionTest.java
kthoms: I don't know how I should implement this. I just copied from XtextRunnerTest and was assuming it was correct there. If you have a concrete idea we might refactor the 2 XtextRunnerTest classes with this change also, or open a new issue to refactor all 3 tests later?

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TestProjectDescriptor.xtend
kthoms: OK, will use 5.1.0 consistent

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/WizardConfiguration.xtend
kthoms: xtext-core has no idea about the TP. Will use JUnitVersion.DEFAULT here.
The wizard knows about presence of JUnit 5.

org.eclipse.xtext.testing/tests/org/eclipse/xtext/testing/tests/InjectionExtensionTest.java
szarnekow: OK. Any chance to extract a common super type for both tests?

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/extensions/InjectionExtension.java
```diff
@@ -0,0 +1,88 @@
...
+	 * <p>
+	 * {@link #getOrCreateInjectorProvider(TestClass)} is used to obtain
+	 * an valid injector provider in the context of a given {@link TestClass}.
+	 * </p>
+	 */
+	private static ClassToInstanceMap<IInjectorProvider> injectorProviderClassCache = MutableClassToInstanceMap.create();
+
+	@Override
+	public void beforeTestExecution(ExtensionContext context) throws Exception {
+		IInjectorProvider injectorProvider = getOrCreateInjectorProvider(context);
```
szarnekow: Would it be worthwhile to make these default methods in IInjectorProvider or at least extract them to a common code for JUnit4 and JUnit5. No dealbreaker, though.

kthoms:
Refactored the both test classes and introduced a base class.

PR #643, opened 2018-02-20
[#638] Use Formatting options passed by the client

cdietrich:
[#638] Use Formatting options passed by the client
Signed-off-by: Christian Dietrich <christian.dietrich@itemis.de>

PR #673, opened 2018-03-26
[formating] use indentation information

JanKoehnlein:

- use IIndentationInformation for default
- don't use an immutable map to be able to override preferences in IFormatter2 impl


JanKoehnlein:
see https://github.com/eclipse/xtext-core/pull/643#discussion_r177144457

PR #675, opened 2018-03-27
[eclipse/xtext-eclipse#616] Use more efficient algorithm.

ArneDeutsch:
Do a depth first traversal the node model in reversed direction and stop
at the first found node before the current offset with specific
properties. This way complete composite nodes can be skipped. This
changes the algorithm complexity from n to log(n) in most cases
(degenerated case ... flat node model).


cdietrich:
Breaks `CompletionTest`

PR #677, opened 2018-03-27
Added nullsafe readOnly and modify operations

tivervac:
Signed-off-by: Titouan Vervack <titouan.vervack@sigasi.com>

Solution to https://github.com/eclipse/xtext-eclipse/issues/587

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/AbstractReadWriteAcces.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/AbstractReadWriteAcces.java
@@ -12,12 +12,13 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.WrappedException;
+import org.eclipse.xtext.util.concurrent.IUnitOfWork.Void;
 
 /**
  * @author Sven Efftinge - Initial contribution and API
  * @author Jan Koehnlein
  */
-public abstract class AbstractReadWriteAcces<P> implements IReadAccess<P> {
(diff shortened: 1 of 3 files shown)


tivervac:
This introduces deprecation warnings in all callsites (there are loads of them). These can be fixed once these patches are considered to be acceptable.

meysholdt:
If you really want to enhance this API, it should also let units of work control whether they want to update cross references (and maybe more). See `ICompositeModificationContext`.

szarnekow:
Thank you for the PR, @tivervac .
What do you think about an alternative design:
In IReadAccess, I would have tried to add a default method
```
default <Result> Result tryReadOnly(.. uow, Function<? super Exception, ? extends Result> defaultResult) {
  try {
    return readOnly((state)->{
      if (state == null) {
        return defaultResult.get();
      }
      return uow.exec(state);
    });
  } catch(Exception e) {
    return defaultResult.apply(e);
  }
}
```
Probably with an overload without an exception handler.
IWriteAccess symmetrically.
I think I wouldn't deprecate the others, but emphasize in the JavaDoc that the given state may be null.

tivervac:
@szarnekow Seems like a better idea, Function has no `get()` though, did you mean this?
```
default <Result> Result tryReadOnly(.. uow, Supplier<Result> defaultResult, Function<? super Exception, ? extends Result> exceptionHandler) {
  try {
    return readOnly((state)->{
      if (state == null) {
        return defaultResult.get();
      }
      return uow.exec(state);
    });
  } catch(Exception e) {
    return exceptionHandler.apply(e);
  }
}```

tivervac:
@meysholdt That seems rather out of scope for eclipse/xtext-eclipse#587, so I'd prefer to just make the current IWriteAccess and IReadAccess nullsafe.

meysholdt:
@tivervac alright, let's not extend the scope here. New issue: https://github.com/eclipse/xtext/issues/2480

PR #682, opened 2018-03-29
Added null- and exception-safe variants of (priority)readOnly and modify

tivervac:
Also added I[Read|Write]Access where applicable

This is based on the feedback of #677

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/AbstractReadWriteAcces.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/AbstractReadWriteAcces.java
@@ -17,11 +17,11 @@ import org.eclipse.emf.common.util.WrappedException;
  * @author Sven Efftinge - Initial contribution and API
  * @author Jan Koehnlein
  */
-public abstract class AbstractReadWriteAcces<P> implements IReadAccess<P> {
+public abstract class AbstractReadWriteAcces<State> implements IReadAccess<State>, IWriteAccess<State> {
 
 	private static Logger log = Logger.getLogger(AbstractReadWriteAcces.class);
 
-	protected abstract P getState();
+	protected abstract State getState();
 
 	/**
 	 * This field should be <code>private</code>. It is <code>protected</code> for API compatibility only. Never access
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/AbstractReadWriteAcces.java
```diff
@@ -17,11 +17,11 @@
  * @author Sven Efftinge - Initial contribution and API
  * @author Jan Koehnlein
  */
-public abstract class AbstractReadWriteAcces<P> implements IReadAccess<P> {
+public abstract class AbstractReadWriteAcces<State> implements IReadAccess<State>, IWriteAccess<State> {
```
szarnekow: Thank you for renaming the type parameters. I like the attention to detail 👍

org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/IReadAccess.java
```diff
@@ -8,22 +8,140 @@
...
+	 * @param exceptionHandler Supplies a result in case an exception is raised during execution
+	 * 
+	 * @return The result of executing {@code work},
+	 *         the result of querying {@code defaultResult} if the State is null, or
+	 *         the result of executing {@code exceptionHandler} in case an exception is raised
+	 * @since 2.14
+	 */
+	default <Result> Result tryReadOnly(
+		IUnitOfWork<Result, State> work,
+		Supplier<Result> defaultResult,
```
szarnekow: We could use Supplier<? extends Result> similar to the signature of the exceptionHandler.

Same for the other try's

szarnekow (approved):
LGTM

cdietrich:
@szarnekow what is the state of this pr?

szarnekow:
Ready to be merged.

cdietrich:
thx @tivervac

cdietrich:
hi can you please add the sample project?

szarnekow:
I assume that the `@Delegate` does not implement a correct detection of equivalent signatures, e.g type parameter-names are not anonymised.

```
interface J<P, Q> {
}

interface I<X> {
	def <Y> Y m(J<Y, X> j);
}

class C<T> implements I<T> {
	@Delegate
	I<T> delegate;
	
	override <V> V m(J<V, T> j) {}
} 
```

Reproduces this bug.

cdietrich:
i am not sure reverting the change would be a good idea....

szarnekow:
Neither am I. We should rather file a ticket for the broken impl in `@Delegate` and try to fix that.

cdietrich:
created https://github.com/eclipse/xtext/issues/2246

sailingKieler:
> Neither am I. We should rather file a ticket for the broken impl in @Delegate and try to fix that.

I would prefer that, too.

PR #687, opened 2018-04-02
Avoid usage of new guava APIs

szarnekow:


org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.xtend
```diff
@@ -270,7 +270,7 @@ class UriExtensionsTest {
 		var directory = createTempDir('some name with whitespaces').toFile;
 		assertTrue(directory.exists)
 		assertTrue(directory.directory)
-		assertTrue(CharMatcher.whitespace.matchesAnyOf(directory.name));
+		assertTrue(directory.name.contains(' '));
```
cdietrich: why change this?

szarnekow: My TP did not contain a Guava version with that method after I re-oomphed.

cdietrich: test should be gradle oomphed and that one should use a newer guava

szarnekow: So this is to be rejected?

cdietrich:
Since it is a test I don’t care as long thing thing what should be tested is still tested

PR #696, opened 2018-04-11
[eclipse/xtext-eclipse#599] Support waiting operation.

ArneDeutsch:
ActiveAnnotationProcessors might run in parallel. Read operations should
wait until all write operations are ready to avoid race conditions.


ArneDeutsch:
Initially I thought this would not be to much work and started an implementation. But now I think this is quite a lot of work. Hence I follow your suggestion and created a ticket for a refactoring.

https://github.com/eclipse/xtext-core/issues/709

PR #701, opened 2018-04-19
Updated to LSP4J 0.4.0

### LINKED ISSUE #694: Update to LSP4J 0.4.0 / Protocol Version 3.6
we should consider a update to LSP4J 0.4.0 / Protocol Version 3.6 

@spoenemann @akosyakov are you going to provide a pr for this?

spoenemann:
Fixes #694.

In LSP 3.6 the parameter type of `textDocument/completion` has changed from `TextDocumentPositionParams` to `CompletionParams`. While this is compatible in TypeScript (the new type adds only optional properties), it is a breaking change in Java :-/

LSP4J 0.4.0 is released next week. I set the dependency to the snapshot version here so we can include the update in the first RC. I'll change it to the released version as soon as it's available.

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -9,7 +9,7 @@ ext.versions = [
 	'xtext_bootstrap': '2.13.0',
 	'gradle_plugins': '0.1.0',
 	'xtext_gradle_plugin': '1.0.20',
-	'lsp4j': '[0.3.0,0.4)',
+	'lsp4j': '0.4.0-SNAPSHOT',
 	'log4j': '1.2.16',
 	'equinoxCommon' : '3.8.0',
 	'equinoxRegistry' : '3.6.100',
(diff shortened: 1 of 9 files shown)


cdietrich:
@spoenemann is it intended that you did no adapt the wizard + its expectations

spoenemann:
I updated the MANIFEST.MF of org.eclipse.xtext.ide.

Why does IdeProjectDescriptor list org.eclipse.lsp4j in the POM dependencies? Doesn't it get in transitively?

cdietrich:
i think it should but it is still used in some poms. dont know why. maybe @kthoms does.

cdietrich:
may cause how gradle/maven deal with transitive/optional deps?

spoenemann:
The dependency is added only if the project is built as Eclipse plugin, so probably it's some Tycho issue. It would be good to know whether this is really necessary so we can decide whether to update the version there or to remove the redundant dependency.

cdietrich:
thx @spoenemann

PR #703, opened 2018-04-20
Java 10 integration

cdietrich:


org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/pom.xml
```diff
@@ -12,7 +12,7 @@
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<!-- Tycho settings -->
-		<tycho-version>1.1.0</tycho-version>
+		<tycho-version>1.2.0-SNAPSHOT</tycho-version>
```
holgerschill: Why do we have to use tycho 1.2.0-SNAPSHOT here? Accident?

cdietrich: cause tycho 1.2.0 is not final yet. and it is needed for java 10. so we want to deliver that so that one the snapshot is completely java 10 capable it can be tried

PR #723, opened 2018-05-02
Issue/722 validation2 fragment

holgerschill:
I have written a fake Xbase grammar in the abstract testclass since loading real Xbase from xtext.tests does not work and duplicating the whole thing for Xbase is a lot of code in addition. However the problems got tested.


org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/generator/ValidatorFragment2Tests.xtend
```diff
@@ -172,6 +172,41 @@ class ValidatorFragment2Tests extends AbstractGeneratorFragmentTests {
...
 				protected void initialize(IAcceptor<PreferenceKey> acceptor) {
+					super.initialize(acceptor);
+					acceptor.accept(create(DEPRECATED_MODEL_PART, SeverityConverter.SEVERITY_WARNING));
+				}
+			}
+		'''.toString, concatenationClientToString(fragment.generateIssueProvider))
+	}
+	
+	@Test
+	def testGenerateConfigurableIssueProvideXbaser() {
```
LorenzoBettini: a very small cosmetic remark: I guess it should be "ProviderXbase" instead of "ProvideXbaser"

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/validation/ValidatorFragment2.xtend
```diff
@@ -232,6 +245,7 @@ class ValidatorFragment2 extends AbstractInheritingFragment {
 			
 				@Override
 				protected void initialize(�IAcceptor�<�PreferenceKey�> acceptor) {
+					super.initialize(acceptor);
```
LorenzoBettini: so we always call super.initialize(acceptor) even for non Xbase languages, is that right?
This sounds sensible, but just to be sure: in the current version for super.initialize was never called

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/validation/ValidatorFragment2.xtend
```diff
@@ -75,18 +77,29 @@ class ValidatorFragment2 extends AbstractInheritingFragment {
...
 	 * @since 2.14
 	 */
 	protected def getAbstractValidatorConfigurationBlockClass(){
 		return new TypeReference("org.eclipse.xtext.ui.validation.AbstractValidatorConfigurationBlock")
 	}
+	
+	/**
+	 * @since 2.14
+	 */
+	protected def getXbaseValidationConfigurationBlockClass(){
```
LorenzoBettini: I was wondering if it would be better to follow the other approach and have something similar to getSuperConfigurableIssueCodesProviderClass (e.g., getSuperValidatorConfigurationBlockClass) instead of having getAbstractValidatorConfigurationBlockClass and getXbaseValidationConfigurationBlockClass

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/validation/ValidatorFragment2.xtend
```diff
@@ -243,11 +257,12 @@ class ValidatorFragment2 extends AbstractInheritingFragment {
 		val javaFile = fileAccessFactory.createGeneratedJavaFile(validatorConfigurationBlockClass)
 		javaFile.content = '''
 			@SuppressWarnings("restriction")
-			public class �validatorConfigurationBlockClass� extends �abstractValidatorConfigurationBlockClass� {
+			public class �validatorConfigurationBlockClass� extends �if(language.grammar.inheritsXbase) xbaseValidationConfigurationBlockClass else abstractValidatorConfigurationBlockClass� {
```
LorenzoBettini: see the above comment on having getSuperValidatorConfigurationBlockClass

org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/generator/ValidatorFragment2Tests.xtend
holgerschill: Sure. Will correct.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/validation/ValidatorFragment2.xtend
holgerschill: Yes. In the current version this is not done since it wasn't necessary. Doing now for all langs does not hurt.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/validation/ValidatorFragment2.xtend
holgerschill: Would work if could bind an impl of XbaseValidationConfigurationBlockClass class but we have to bind and impl of AbstractValidatorConfigurationBlockClass instead. This is why I did it this way to make a clear difference.

LorenzoBettini: OK, I suspected it was something like that :)

holgerschill:
Thanks @LorenzoBettini !

PR #729, opened 2018-05-04
[eclipse/xtext-web#53] updated to a newer jetty version that works better with asm 6.1

cdietrich:
[eclipse/xtext-web#53] updated to a newer jetty version that works better with asm 6.1


cdietrich:
please review

PR #731, opened 2018-05-08
use tycho 1.2.0 in the wizard

cdietrich:

### DIFF
--- a/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
+++ b/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
@@ -10,6 +10,7 @@ package org.eclipse.xtext.xtext.wizard
 import com.google.common.base.Charsets
 import com.google.common.io.Resources
 import org.eclipse.xtend.lib.annotations.FinalFieldsConstructor
+import org.eclipse.xtext.util.JavaVersion
 
 @FinalFieldsConstructor
 class ParentProjectDescriptor extends ProjectDescriptor {
@@ -69,7 +70,7 @@ class ParentProjectDescriptor extends ProjectDescriptor {
 	}
 	
 	def String getTychoVersion() {
-		'1.2.0-SNAPSHOT'
+		'1.1.0'
 	}
 	
 	def private CharSequence loadResource(String resourcePath) {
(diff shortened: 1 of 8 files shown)


org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
```diff
@@ -69,7 +69,7 @@ class ParentProjectDescriptor extends ProjectDescriptor {
 	}
 	
 	def String getTychoVersion() {
-		'1.2.0-SNAPSHOT'
+		'1.2.0'
```
kthoms: Could we only use 1.2.0 when Java 10 is selected, otherwise 1.1.0?
And then leave the comment out?

cdietrich: well ill have a look when i have 5 hours time to adjust the tests. we need a better way to adjust them

cdietrich:
@kthoms updated

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
```diff
@@ -69,7 +70,7 @@ class ParentProjectDescriptor extends ProjectDescriptor {
 	}
 	
 	def String getTychoVersion() {
-		'1.2.0-SNAPSHOT'
+		'1.1.0'
```
kthoms: I would expect the condition here:
if (config.javaVersion == JavaVersion.JAVA10) '1.2.0' else '1.1.0'

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
```diff
@@ -215,7 +216,12 @@ class ParentProjectDescriptor extends ProjectDescriptor {
 					<maven.compiler.target>�javaVersion�</maven.compiler.target>
 					�IF config.needsTychoBuild�
 						<!-- Tycho settings -->
-						<tycho-version>�tychoVersion�</tycho-version>
+						�IF config.javaVersion == JavaVersion.JAVA10�
+							<!-- Java 10 support requires Tycho 1.2.0, which was not released at release date of Xtext 2.14.0. When not available downgrade to Java 9 and Tycho 1.1.0. -->
+							<tycho-version>1.2.0</tycho-version>
```
kthoms: Only the comment would be conditional here

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
kthoms: WDYT?

cdietrich: Won’t catch the comment

cdietrich: Since we throw this away anyway ....

PR #735, opened 2018-05-16
[#666] JUnit 5 dependencies

kthoms:
- Use Tycho 1.2 for JUnit 5
- Use only org.junit.jupiter.api dependency on test projects
- Remove implicit dependencies for JUnit 4 also


cdietrich:
works from maven but i dont get any run as context menu for junit 5

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/JUnitFragment.xtend
```diff
@@ -100,22 +100,12 @@ class JUnitFragment extends AbstractStubGeneratingFragment {
 		].filterNull.forEach [
 			if (junitVersion == JUnitVersion.JUNIT_4) {
 				importedPackages.addAll(
-					"org.junit;version=\"4.5.0\"",
```
cdietrich: do we need import package and require bundle=

kthoms: matter of taste. we might be able to drop the import package. A newly created project with JU4 already adds a Required-Bundle for org.junit 4.12

The Import-Package is added later by the JUnitFragment. Seems redundant to me.

screenshot 104

cdietrich: yes this is that i meant

kthoms: removed that now. the wizard already creates test bundles with JU require bundle dependency, so the JUnitFragment does not need to do this also.

cdietrich: Ok

kthoms:
> works from maven but i dont get any run as context menu for junit 5

I have a context menu for Run As / JUnit test. Or what do you mean?

This PR only handles the dependencies created by the generator. I updated the PR, please review again.

screenshot 105

PR #747, opened 2018-05-22
[serializer] relax API

JanKoehnlein:


org.eclipse.xtext/src/org/eclipse/xtext/formatting2/regionaccess/internal/TextRegionAccessBuildingSequencer.java
```diff
@@ -141,6 +145,11 @@ protected StringHiddenRegion createHiddenRegion() {
...
 	}
 
+	protected StringSemanticRegion createSemanticRegion(EObject element, String token,
+			AbstractEObjectRegion obj, int offset) {
+		return new StringSemanticRegion(regionAccess, obj, element, offset, token.length());
+	}
+
 	protected StringSemanticRegion createSemanticRegion(AbstractElement element, String token,
 			AbstractEObjectRegion obj, int offset) {
 		return new StringSemanticRegion(regionAccess, obj, element, offset, token.length());
```
meysholdt: I think this method should be deprecated and delegate to the one you introduced.

JanKoehnlein:
We agreed offline on deprecating and renaming the old method overload.

Issue #506, opened 2017-10-13
Java 11+ hates some of the reflection stuff we do


cdietrich:
e.g.

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.eclipse.xtext.util.MergeableManifest (file:/C:/Users/aart/eclipse/plugins/org.eclipse.xtext.util_2.13.0.v20171005-0628.jar) to field java.util.jar.Manifest.attr
WARNING: Please consider reporting this to the maintainers of org.eclipse.xtext.util.MergeableManifest
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

cdietrich:
no news on guice side

cdietrich:
move to 2.22

cdietrich:
still waiting for guice fix to be released. move to 2.24

wmanth:
Any news on that? Would be reducing the extensive use of dependency injection an option?
Some people (or is it just me?) anyway have problems understanding what's going on and how to fix things if they don't work as expected.

And obviously Guice is not a well behaving first class citizen on Java >8 anymore.

cdietrich:
there is a guice 5 beta out containing a fix.
but that brings other problems
so it will take some time to integrate it
once it is release

see also https://github.com/eclipse/xtext-core/issues/393
you might also experiment using guice 3 noaop

cdietrich:
as there is still no guice release: move to 2.25

cdietrich:
as there is still no guice release: move to 2.26

melaasar:
I think there is a release now: https://github.com/google/guice/releases/tag/5.0.1

cdietrich:
@melaasar fee free to test with the information in https://github.com/eclipse/xtext-core/issues/393

cdietrich:
TODO: milestone & bootstrapping test

cdietrich:
there is also still some warnings in tycho/pde that needs tested with new version

cdietrich:
done

PR #750, opened 2018-05-24
[#506] Removed reflection call.

ArneDeutsch:
Do not inherit and modify (by reflection) the original
java.util.jar.Manifest but instead C&P and modify the code. Adapt
ManifestMergerTest by importing the new classes.


ArneDeutsch:
Something went wrong ...

PR #754, opened 2018-05-25
 [release] Increment version to 2.15.0

cdietrich:
[release] Increment version to 2.15.0

### DIFF
--- a/releng/releng-target/xtext-core.target.target
+++ b/releng/releng-target/xtext-core.target.target
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.8"?>
-<target name="org.eclipse.xtext.core.target" sequenceNumber="0">
+<target name="org.eclipse.xtext.helios.target" sequenceNumber="0">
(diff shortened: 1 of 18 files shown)


releng/releng-target/xtext-core.target.target
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.8"?>
-<target name="org.eclipse.xtext.core.target" sequenceNumber="0">
+<target name="org.eclipse.xtext.helios.target" sequenceNumber="0">
```
kthoms: this change is produced by the generateP2Build task and should be reverted

cdietrich: will d you that or should i?
we fix the generate p2 build task then

kthoms: OK, we'll put this on our list

kthoms: Tracked in https://github.com/eclipse/xtext/issues/1220

PR #766, opened 2018-06-11
Added an extra api call that omits the default result

tivervac:


cdietrich:
thx @tivervac

szarnekow:
Looks like the entire API could greatly benefit from `Optional<Result>`

PR #768, opened 2018-06-14
Allow to not print source code in LSP

tivervac:
It is not always wanted to have your code be printed out when an error occurs

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
@@ -20,8 +20,22 @@ import org.eclipse.lsp4j.Range
 
     Integer version
     String contents
+    boolean printSourceOnError
 
-    def int getOffSet(Position position) {
+    new(Integer version, String contents) {
+        this(version, contents, true)
+    }
+
+    /**
+     * @since 2.15
+     */
+    new(Integer version, String contents, boolean printSourceOnError) {
+        this.version = version
+        this.contents = contents
+        this.printSourceOnError = printSourceOnError
+    }
+
+    def int getOffSet(Position position) throws IndexOutOfBoundsException {
         val l = contents.length
         val char NL = '\n'
         var line = 0
@@ -41,13 +55,13 @@ import org.eclipse.lsp4j.Range
         if (position.line === line && position.character === column) {
             return l
         }
-        throw new IndexOutOfBoundsException(position.toString + " text was : " + contents)
+        throw new IndexOutOfBoundsException(position.toString + if (printSourceOnError) "" else (" text was : " + contents))
     }
 
-    def Position getPosition(int offset) {
+    def Position getPosition(int offset) throws IndexOutOfBoundsException{
         val l = contents.length
         if (offset < 0 || offset > l)
-            throw new IndexOutOfBoundsException(offset + " text was : " + contents)
+            throw new IndexOutOfBoundsException(offset + if (printSourceOnError) "" else (" text was : " + contents))
 
         val char NL = '\n'
         var line = 0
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
```diff
@@ -20,8 +20,19 @@ import org.eclipse.lsp4j.Range
...
     Integer version
     String contents
+    boolean printSourceOnError
 
-    def int getOffSet(Position position) {
+    new(Integer version, String contents) {
+        this(version, contents, true)
+    }
+
+    new(Integer version, String contents, boolean printSourceOnError) {
```
cdietrich: misses `@since`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
```diff
@@ -20,8 +20,19 @@ import org.eclipse.lsp4j.Range
 
     Integer version
     String contents
+    boolean printSourceOnError
```
cdietrich: am not sure if we need explicit getters with since for the flag too

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
```diff
@@ -41,13 +52,13 @@ import org.eclipse.lsp4j.Range
         if (position.line === line && position.character === column) {
             return l
         }
-        throw new IndexOutOfBoundsException(position.toString + " text was : " + contents)
+        throw new IndexOutOfBoundsException(position.toString + if (printSourceOnError) "" else (" text was : " + contents))
```
cdietrich: what is the actual reason you want to change the behaviour.
the index out of bounces should actually never happen

tivervac: It should never happen, but it did in my implementation, probably because of a mistake on my part. However, we have clients where code is never allowed to leak, even when talkback is enabled. A small mistake on our side could mean that some code of one of our clients could leak back to us. I'm convinced that other Xtext users have similar clients and thus want a way to disable this.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
tivervac: I can see why we would want that, but it's quite annoying to add those in an @Data class

cdietrich: yes the same is for contructors. @szarnekow what is our general approach regarding api and active annotation produced code

org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/server/Document.java
```diff
@@ -25,7 +25,25 @@
   
   private final String contents;
   
-  public int getOffSet(final Position position) {
+  private final boolean printSourceOnError;
+  
+  /**
+   * @since 2.15
```
cdietrich: this is the old constructor so we wont need the `@since` here but only for the api wise new stuff

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
szarnekow: Usually one has to resort to manually written code to get the annotations right and sound.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
szarnekow: :+1:

Issue #69, opened 2016-08-10
[Content Assist] does not work properly without org.eclipse.xtext.common.Terminals


miklossy:
Given is the following xtext grammar:
```
grammar org.xtext.example.mydsl.MyDsl

generate myDsl "http://www.xtext.org/example/mydsl/MyDsl"

PolygonBasedShape:
	shape = PolygonBasedNodeShape;

enum PolygonBasedNodeShape:
	octagon | oval |
	parallelogram | pentagon | plain | plaintext | point | polygon | primersite | promoter | proteasesite | proteinstab |
	rarrow | record | rect | rectangle | restrictionsite | ribosite | rnastab | rpromoter;
```
Use Case1 (Content Assistant works properly):
1. Create an empty test.mydsl file based on the grammar above and open it with the generated editor.
2. Press Ctrl+Space to invoke the Content Assistant.
3. Type the letter 'p' and 'r'. 
The Content Assistant suggests correct proposals beginning with the prefix 'pr'.
![content_assistant_ok](https://cloud.githubusercontent.com/assets/20393472/17555787/28361d9a-5f11-11e6-9182-e28f74815c6c.png)

Use Case2 (Content Assistant does not work properly):
1. Create an empty test.mydsl file based on the grammar above and open it with the generated editor.
2. Type the letter 'p' and 'r'.
3. Press Ctrl+Space to invoke the Content Assistant. 
The Content Assistant suggests incorrect proposals beginning with the prefix 'r' (see screenshot content_assistant_nok.png). 
![content_assistant_not_ok](https://cloud.githubusercontent.com/assets/20393472/17555786/282069e6-5f11-11e6-81bd-982bc59e0dc9.png)

HINT: When modifying the grammar to use the "org.eclipse.xtext.common.Terminals", the content assistant works for both of the use cases properly.

meysholdt:
how is the behavior for a grammar that does not extend common.Terminals but does have the terminal rule `ID` defined (copied from common.Terminals)?

miklossy:
The same problem.

cdietrich:
@ArneDeutsch can this be closed?

ArneDeutsch:
We have found a working fix but it is actually quite ugly because the error message need to be inspected to get all cases green. https://github.com/eclipse/xtext-core/commit/35582ebd66efca10a718ce878f455bedcff910a9

[...]

PR #776, opened 2018-06-25
[#69] Fix content assist if lexer returns multiple invalid tokens.

ArneDeutsch:
If the lexer returns multiple invalid tokens in succession only the last
one is taken into account for content assist. Especially if there are
keywords only (no ID or similar) this is an issue because only the last
character is used as prefix.

The fix assumes that ALL tokens with syntax errors that come in
succession are relevant for content assist. It seems not to be useful to
execute content assist otherwise.

This can never be reproduced with any grammar that includes standart
"Terminals" because of the ID and the ANY_OTHER rule. The ID rule makes
sure that the start is usually read as only ONE token, whereas the
ANY_OTHER rule ensures there is never a lexer error.

As this is merely a side case for grammars without an ID it seems to much overhead to define an extra test language to implement tests to cover this.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/ContentAssistContextFactory.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/ContentAssistContextFactory.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2014, 2018 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -409,7 +409,18 @@ public class ContentAssistContextFactory implements Function<ContentAssistContex
 		if (prefixNode instanceof ILeafNode) {
 			if (((ILeafNode) prefixNode).isHidden() && prefixNode.getGrammarElement() != null)
 				return "";
-			return getNodeTextUpToCompletionOffset(prefixNode);
+			INode startingNode = prefixNode;
+			// As long as there are leaf nodes with syntax errors one after the other go back until the first one is
+			// found
+			while (startingNode.getSyntaxErrorMessage() != null) {
+				INode nodeBefore = startingNode.getPreviousSibling();
+				if (nodeBefore.getSyntaxErrorMessage() != null) {
+					startingNode = nodeBefore;
+				} else {
+					break;
+				}
+			}
+			return getNodeTextUpToCompletionOffset(startingNode);
 		}
 		StringBuilder result = new StringBuilder(prefixNode.getTotalLength());
 		doComputePrefix((ICompositeNode) prefixNode, result);
@@ -419,7 +430,7 @@ public class ContentAssistContextFactory implements Function<ContentAssistContex
 	public String getNodeTextUpToCompletionOffset(INode currentNode) {
 		int startOffset = currentNode.getOffset();
 		int length = completionOffset - startOffset;
-		String nodeText = ((ILeafNode) currentNode).getText();
+		String nodeText = currentNode.getRootNode().getText().substring(startOffset, startOffset + length);
 		String trimmedNodeText = length > nodeText.length() ? nodeText : nodeText.substring(0, length);
 		try {
 			String text = document.substring(startOffset, startOffset + trimmedNodeText.length());


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/ContentAssistContextFactory.java
```diff
@@ -409,7 +412,20 @@ public String getPrefix(INode prefixNode) {
 		if (prefixNode instanceof ILeafNode) {
 			if (((ILeafNode) prefixNode).isHidden() && prefixNode.getGrammarElement() != null)
 				return "";
-			return getNodeTextUpToCompletionOffset(prefixNode);
+			INode startingNode = prefixNode;
+			// As long as there are leaf nodes with syntax errors one after the other go back until the first one is
+			// found
+			while (startingNode instanceof LeafNodeWithSyntaxError
```
szarnekow: Please do not cast to concrete types but check node.hasSyntaxError instead.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/ContentAssistContextFactory.java
```diff
@@ -419,7 +435,16 @@ public String getPrefix(INode prefixNode) {
 	public String getNodeTextUpToCompletionOffset(INode currentNode) {
 		int startOffset = currentNode.getOffset();
 		int length = completionOffset - startOffset;
-		String nodeText = ((ILeafNode) currentNode).getText();
+		final String nodeText;
+		ICompositeNode root = currentNode.getRootNode();
+		if (root instanceof RootNode) {
```
szarnekow: PLease do not cast to concrete types. root.getText() will yield the same. Also root should never be something else than a RootNode

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/ContentAssistContextFactory.java
ArneDeutsch: There is no such method ... or at least I'm not able to find it.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/ContentAssistContextFactory.java
ArneDeutsch: Because 'getRoot' returns a 'ICompositeNode' I was not sure about that.

cdietrich:
@szarnekow @ArneDeutsch what is the state of this pr?

ArneDeutsch:
From my point of view this should be done. But someone else then me should open an Xtext file external to the workspace and re-check that everything is working as expected. Especially OpenDeclaration (Hyperlink), saving ... and what comes in mind additionally.

ArneDeutsch:
If Sebastian has no objections I would say this could be merged.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/ContentAssistContextFactory.java
```diff
@@ -409,7 +409,18 @@ public String getPrefix(INode prefixNode) {
 		if (prefixNode instanceof ILeafNode) {
 			if (((ILeafNode) prefixNode).isHidden() && prefixNode.getGrammarElement() != null)
 				return "";
-			return getNodeTextUpToCompletionOffset(prefixNode);
+			INode startingNode = prefixNode;
+			// As long as there are leaf nodes with syntax errors one after the other go back until the first one is
+			// found
+			while (startingNode.getSyntaxErrorMessage() != null) {
+				INode nodeBefore = startingNode.getPreviousSibling();
+				if (nodeBefore.getSyntaxErrorMessage() != null) {
```
szarnekow: getPrevSibling may return null.

szarnekow:
We'd need a testlanguage for this.

ArneDeutsch:
Ok, I will create a test language that does not use Terminals. Guess it will be a lot more effort then the fix ;).

cdietrich:
dont we have any?

ArneDeutsch:
At least I havn't found any.

ArneDeutsch:
This case needs a very special setup. `EncodingUiTestLanguage` for example is not usable here. Because of LEVEME and ANY_OTHER terminal.

szarnekow:
Any progress on this one?

ArneDeutsch:
No, have not yet started to implement tests for this one.

szarnekow:
Any news? Do you need help with this @ArneDeutsch ?

ArneDeutsch:
Test language and test case: https://github.com/eclipse/xtext-eclipse/pull/776

ArneDeutsch:
Test language and test case: https://github.com/eclipse/xtext-eclipse/pull/777 (previous comit was broken)

cdietrich:
tests are merged and now "red" @szarnekow @ArneDeutsch what is the state of this pr

PR #777, opened 2018-06-27
[xtend#474] Allow to skip generation of validator ui contributions

szarnekow:
An additional boolean flag was added to the generator fragment which allows to skip the generation of the validation block. This minimised the changes after regenerating Xtend.


cdietrich:
@holgerschill would be nice if you would have a look too

szarnekow:
I'll merge this one. @holgerschill Please take a look if you find the time. I'll fix things later on, if you request changes.

PR #781, opened 2018-07-04
[eclipse/xtext-eclipse#729] Adapt output folder layout.

ArneDeutsch:
Investigated how previous Xtext versions handled the output folders and
what values m2e expects. Adapted output folders accordingly. "Maven ->
Update Project..." does not change the output folders any more.


org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/SourceLayout.java
```diff
@@ -45,23 +45,23 @@ public String getOutputFor(Outlet outlet) {
 			case ROOT:
 				return "";
 			case MAIN_JAVA:
-				return "bin";
+				return "target/classes";
```
cdietrich: is this for maven or for all cases?

ArneDeutsch: There are two cases:
- `SourceLayout.PLAIN`
- `SourceLayout.MAVEN`
But in fact the output folders are now aligned between both. In contrast to the source folders.

ArneDeutsch: See my mistake. Have confidered only two cases (Maven+Eclipse Layout, Maven+Maven Layout) and skipped the last one (No Maven+Eclipse Layout). Will have a look about the last case.

cdietrich: you can have gradle with both layouts too

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ProjectDescriptor.xtend
```diff
@@ -174,12 +174,13 @@ abstract class ProjectDescriptor {
 		config.sourceLayout.getPathFor(outlet);
 	}
 
-	def outputFolder(Outlet outlet) {
-		config.sourceLayout.getOutputFor(outlet);
-	}
-
 	def isTest(Outlet outlet) {
-		config.sourceLayout.isTest(outlet);
+		for (o : Outlet.testOutlets) {
```
szarnekow: looks like we should do a contains() on the testOutlets. If it is an Array, Arrays.asLists() should do the trick.

ArneDeutsch: Changed as requested.

PR #806, opened 2018-07-30
GH-798: Increased the visibility of the stateful incremental builder.

### LINKED ISSUE #798: [ls] Convenient support for custom `InternalStatefulIncrementalBuilder` bindings
Currently, I need to do something like this if I would like to have a custom binding for my LS. 
```java
class MyServerModule extends ServerModule {

  override protected configure() {
    super.configure();
    binder.bindInternalStatefulIncrementalBuilder;
  }

  protected def bindInternalStatefulIncrementalBuilder(Binder binder) {
    val fromClass = IncrementalBuilder.classLoader.loadClass(
      'org.eclipse.xtext.build.IncrementalBuilder$InternalStatefulIncrementalBuilder') as Class<Object>;
    (binder.bind(fromClass) as AnnotatedBindingBuilder<Object>).to(MyInternalStatefulIncrementalBuilder);
  }

}
```

The `InternalStatefulIncrementalBuilder` class is a protected static class of the `IncrementalBuilder`. Please make it `public`, or I can provide a PR too.

kittaakos:
Added `protected` setters for the build context and the request.
Replaced the context and request assignments with setter calls.

Closes #798.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.xtend
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.xtend
@@ -48,10 +48,10 @@ import org.eclipse.xtext.generator.GeneratorContext
 		List<IResourceDescription.Delta> affectedResources
 	}
 	
-	@Log protected static class InternalStatefulIncrementalBuilder {
+	@Log static class InternalStatefulIncrementalBuilder {
 	
-		@Accessors(PROTECTED_SETTER) extension BuildContext context
-		@Accessors(PROTECTED_SETTER) BuildRequest request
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.xtend
```diff
@@ -48,10 +48,10 @@ import org.eclipse.xtext.generator.GeneratorContext
 		List<IResourceDescription.Delta> affectedResources
 	}
 	
-	@Log protected static class InternalStatefulIncrementalBuilder {
+	@Log static class InternalStatefulIncrementalBuilder {
 	
-		@Accessors(PROTECTED_SETTER) extension BuildContext context
```
cdietrich: what is the getter needed for?

cdietrich:
thx @kittaakos

PR #807, opened 2018-07-31
[ChangeSerializer] Don't record EObjectDescs if not needed.

meysholdt:
When CrossReferences are not updated, recording the descriptions should
not be needed.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/EObjectSnapshotProvider.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/EObjectSnapshotProvider.java
@@ -148,10 +148,14 @@ public class EObjectSnapshotProvider {
 		protected final Resource resource;
 		protected final URI uri;
 
-		public ResourceSnapshot(EObjectSnapshotProvider strategy, Resource resource) {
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/EObjectSnapshotProvider.java
```diff
@@ -236,8 +240,8 @@ public String toString() {
 
 	}
 
-	public IResourceSnapshot createResourceSnapshot(Resource resource) {
-		return new ResourceSnapshot(this, resource);
+	public IResourceSnapshot createResourceSnapshot(Resource resource, boolean recordReferences) {
+		return new ResourceSnapshot(this, resource, recordReferences);
```
spoenemann: Add an overload without the new parameter which defaults to true for compatibility?

cdietrich: do we need that.afaik this is not yet api

meysholdt: this is indeed not API... skipping.

meysholdt:
thanks for the review!

PR #810, opened 2018-08-07
Added TargetURIConverter to non UI

tivervac:
Will also make the UI one override from this one

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/findReferences/TargetURIConverter.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/findReferences/TargetURIConverter.java
@@ -1,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) 2014 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/findReferences/TargetURIConverter.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.eclipse.emf.common.util.URI;
+
+import com.google.inject.Inject;
+import com.google.inject.Provider;
+import com.google.inject.Singleton;
+
+/**
+ * @author Titouan Vervack - Moved from UI to non UI
+ * @since 2.15
```
cdietrich: i dont think we want this to be api so maybe we should remove the since

org.eclipse.xtext/src/org/eclipse/xtext/findReferences/TargetURIConverter.java
```diff
@@ -0,0 +1,37 @@
+/*******************************************************************************
```
cdietrich: can you use the original copyright header

tivervac: Didn't notice the template was different, will do

Issue #793, opened 2018-07-19
Update to LSP4J 0.5.0


cdietrich:
Lsp4j will participate with 0.5.0 in 2018-09.
We should thus update the version we consume once it is available

@spoenemann can you give some hints on your time schedule?

Tasks
- Update Dependency
- Update IdePluginDescriptor and Tests
- Update Impl (e.g. codeAction)

spoenemann:
https://projects.eclipse.org/projects/technology.lsp4j

cdietrich:
do you have plans to publish snapshots so that we can consume a range or make the consumable dependent of xtext release or nightly beeing used

https://oss.sonatype.org/content/repositories/snapshots/org/eclipse/lsp4j/org.eclipse.lsp4j/

spoenemann:
We could shift the release date of one of the two projects, or simply make the build artifacts of LSP4J 0.5.0 available a few days earlier.

cdietrich:
the main problem is that producing xtext releases is still far from beeing a 1 hour task but rather a 1+ days task so i'd prefer it to have lsp4j before

spoenemann:
I'll try to create a milestone version of LSP4J next week so Xtext can be tested with that. We'll create the release build for LSP4J on August 3rd so it can be consumed by the Xtext release.

dhuebner:
FYI: The Lsp4J 0.5.0M1 is contributed to the Photon 2018-09 staging repository

spoenemann:
LSP4J 0.5.0 is now available, so Xtext 2.15.0 can be built with it.

The Maven artifacts are already published, but the p2 repository is not yet copied to the Eclipse downloads section. Until that is done, the following URL can be used:

https://ci.eclipse.org/lsp4j/job/lsp4j-release/20/artifact/build-result/p2.repository/

cdietrich:
ill give it a try

cdietrich:
artifacts are not mirrored to jcenter yet

cdietrich:
Fixed in 2.15. thx @spoenemann @kittaakos

PR #815, opened 2018-08-14
GH-812: Implemented the semantic highlighting for the Xtext LS.

### LINKED ISSUE #812: [ls] Support the semantic highlighting with the Xtext LS
There is a proposed LSP extension for the semantic highlighting here: https://github.com/Microsoft/vscode-languageserver-node/pull/367.

I would like to start working on the Xtext LS part.

kittaakos:
Implementation of the LS-based semantic highlighting. It does **not** support deltas.

The PR is in **WIP** because of the `lsp4j` dependency. @spoenemann, can you please advise how to tackle the `0.4.0` vs. the `0.5.0` LSP4J version mismatch?

Closes #812.

CC: @svenefftinge

**Update:**
 - ~Depends on #793.~
 - Depends on #821.

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -9,7 +9,11 @@ ext.versions = [
 	'xtext_bootstrap': '2.14.0',
 	'gradle_plugins': '0.1.0',
 	'xtext_gradle_plugin': '2.0.1',
-	'lsp4j': '0.4.0', // dont forget to adapt IdeProjectDescriptor + Tests
+	// When adjusting the `lsp4j` version here, do not forget to do the followings:
+	// - Update the version in `IdeProjectDescriptor#pom`.
+	// - Update the test expectations by running `CliWizardIntegrationTest#main`.
+	// - Update the versions in the `/org.eclipse.xtext.ide/META-INF/MANIFEST.MF` and `/org.eclipse.xtext.ide.tests/META-INF/MANIFEST.MF` files.
+	'lsp4j': '0.5.0.M1',
 	'log4j': '1.2.16',
 	'equinoxCommon' : '3.9.0',
 	'equinoxRegistry' : '3.7.0',
--- a/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/TestLanguageRuntimeModule.xtend
+++ b/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/TestLanguageRuntimeModule.xtend
@@ -8,9 +8,13 @@
 package org.eclipse.xtext.ide.tests.testlanguage
 
 import org.eclipse.xtext.formatting2.IFormatter2
+import org.eclipse.xtext.ide.editor.syntaxcoloring.ISemanticHighlightingCalculator
 import org.eclipse.xtext.ide.server.coloring.IColoringService
+import org.eclipse.xtext.ide.server.semanticHighlight.ISemanticHighlightingStyleToTokenMapper
 import org.eclipse.xtext.ide.server.signatureHelp.ISignatureHelpService
 import org.eclipse.xtext.ide.tests.testlanguage.coloring.ColoringServiceImpl
+import org.eclipse.xtext.ide.tests.testlanguage.editor.syntaxcoloring.SemanticHighlightingCalculatorImpl
+import org.eclipse.xtext.ide.tests.testlanguage.editor.syntaxcoloring.SemanticHighlightingStyleToTokenMapper
 import org.eclipse.xtext.ide.tests.testlanguage.formatting2.TestLanguageFormatter
 import org.eclipse.xtext.ide.tests.testlanguage.signatureHelp.SignatureHelpServiceImpl
 
@@ -23,12 +27,20 @@ class TestLanguageRuntimeModule extends AbstractTestLanguageRuntimeModule {
 		return TestLanguageFormatter;
 	}
 	
-    def Class<? extends ISignatureHelpService> bindSignatureHelpService() {
-        return SignatureHelpServiceImpl;
-    }
-    
-    def Class<? extends IColoringService> bindIColoringService() {
-    	return ColoringServiceImpl;
-    }
+	def Class<? extends ISignatureHelpService> bindSignatureHelpService() {
+		return SignatureHelpServiceImpl;
+	}
+
+	def Class<? extends IColoringService> bindIColoringService() {
+		return ColoringServiceImpl;
+	}
+
+	def Class<? extends ISemanticHighlightingCalculator> bindISemanticHighlightingCalculator() {
+		return SemanticHighlightingCalculatorImpl;
+	}
+
+	def Class<? extends ISemanticHighlightingStyleToTokenMapper> bindISemanticHighlightingStyleToTokenMapper() {
+		return SemanticHighlightingStyleToTokenMapper;
+	}
 
 }
--- a/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/editor/syntaxcoloring/SemanticHighlightingStyleToTokenMapper.xtend
+++ b/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/editor/syntaxcoloring/SemanticHighlightingStyleToTokenMapper.xtend
@@ -1,0 +1,22 @@
+/*******************************************************************************
+ * Copyright (c) 2018 TypeFox and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.testlanguage.editor.syntaxcoloring
+
+import org.eclipse.xtext.ide.server.semanticHighlight.ISemanticHighlightingStyleToTokenMapper
+
+class SemanticHighlightingStyleToTokenMapper implements ISemanticHighlightingStyleToTokenMapper {
+
+	override toScopes(String styleId) {
+		return #[styleId];
+	}
+
+	override getAllStyleIds() {
+		return SemanticHighlightingCalculatorImpl.STYLES;
+	}
+
+}
--- a/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.xtend
+++ b/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.xtend
@@ -29,6 +29,7 @@ import org.eclipse.xtext.util.CollectionBasedAcceptor
 import org.eclipse.xtext.util.StringInputStream
 
 import static org.eclipse.xtext.ide.tests.testlanguage.validation.TestLanguageValidator.*
+import org.eclipse.lsp4j.jsonrpc.messages.Either
 
 /**
  * @author Sven Efftinge - Initial contribution and API
@@ -46,7 +47,7 @@ class CodeActionService implements ICodeActionService {
 				case UNSORTED_MEMBERS: commands += d.fixUnsortedMembers(document, resource, params)
 			}
 		}
-		return commands
+		return commands.map[Either.forLeft(it)];
 	}
 
 	def private Command fixInvalidName(Diagnostic d, Document doc, XtextResource res, CodeActionParams params) {
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
@@ -35,6 +35,7 @@ import org.eclipse.lsp4j.DidSaveTextDocumentParams
 import org.eclipse.lsp4j.DocumentFormattingParams
 import org.eclipse.lsp4j.DocumentOnTypeFormattingParams
 import org.eclipse.lsp4j.DocumentRangeFormattingParams
+import org.eclipse.lsp4j.DocumentSymbol
 import org.eclipse.lsp4j.DocumentSymbolParams
 import org.eclipse.lsp4j.ExecuteCommandOptions
 import org.eclipse.lsp4j.ExecuteCommandParams
@@ -47,6 +48,7 @@ import org.eclipse.lsp4j.PublishDiagnosticsParams
 import org.eclipse.lsp4j.Range
 import org.eclipse.lsp4j.ReferenceParams
 import org.eclipse.lsp4j.RenameParams
+import org.eclipse.lsp4j.SemanticHighlightingServerCapabilities
 import org.eclipse.lsp4j.ServerCapabilities
 import org.eclipse.lsp4j.SignatureHelpOptions
 import org.eclipse.lsp4j.SymbolInformation
@@ -81,6 +83,7 @@ import org.eclipse.xtext.ide.server.formatting.FormattingService
 import org.eclipse.xtext.ide.server.hover.IHoverService
 import org.eclipse.xtext.ide.server.occurrences.IDocumentHighlightService
 import org.eclipse.xtext.ide.server.rename.IRenameService
+import org.eclipse.xtext.ide.server.semanticHighlight.SemanticHighlightingRegistry
 import org.eclipse.xtext.ide.server.signatureHelp.ISignatureHelpService
 import org.eclipse.xtext.ide.server.symbol.DocumentSymbolService
 import org.eclipse.xtext.ide.server.symbol.WorkspaceSymbolService
@@ -104,6 +107,7 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 	@Inject extension UriExtensions
 	@Inject extension IResourceServiceProvider.Registry languagesRegistry
 	@Inject ExecutableCommandRegistry commandRegistry
+	@Inject SemanticHighlightingRegistry semanticHighlightingRegistry
 	
 	// injected below
 	WorkspaceManager workspaceManager
@@ -151,7 +155,7 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 			// check if a language with code actions capability exists
 			codeActionProvider = allLanguages.exists[get(ICodeActionService)!==null] 
 			
-            signatureHelpProvider = new SignatureHelpOptions(#['(', ','])
+			signatureHelpProvider = new SignatureHelpOptions(#['(', ','])
 			textDocumentSync = TextDocumentSyncKind.Incremental
 			completionProvider = new CompletionOptions => [
 				resolveProvider = false
@@ -162,13 +166,18 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 			documentHighlightProvider = true
 			renameProvider = allLanguages.exists[get(IRenameService)!==null]
 
+			val clientCapabilities = params.capabilities;
 			// register execute command capability
-			if (params.capabilities?.workspace?.executeCommand !== null) {
-				this.commandRegistry.initialize(allLanguages, params.capabilities, client)
+			if (clientCapabilities?.workspace?.executeCommand !== null) {
+				this.commandRegistry.initialize(allLanguages, clientCapabilities, client)
 				executeCommandProvider = new ExecuteCommandOptions => [
 					commands = this.commandRegistry.getCommands()
 				]
 			}
+
+			semanticHighlightingRegistry.initialize(allLanguages, clientCapabilities, client);
+			semanticHighlighting = new SemanticHighlightingServerCapabilities(semanticHighlightingRegistry.allScopes);
+
 		]
 		for (language : allLanguages) {
 			language.get(ICapabilitiesContributor)?.contribute(capabilities, params)
@@ -383,12 +392,13 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 	}
 
 	override documentSymbol(DocumentSymbolParams params) {
-		return requestManager.<List<? extends SymbolInformation>>runRead[ cancelIndicator |
+		return requestManager.<List<Either<SymbolInformation, DocumentSymbol>>>runRead[ cancelIndicator |
 			val uri = params.textDocument.uri.toUri
 			val resourceServiceProvider = uri.resourceServiceProvider
 			val documentSymbolService = resourceServiceProvider?.get(DocumentSymbolService)
-			if (documentSymbolService === null)
+			if (documentSymbolService === null) {
 				return emptyList
+			}
 	
 			return workspaceManager.doRead(uri) [ document, resource |
 				return documentSymbolService.getSymbols(document, resource, params, cancelIndicator)
@@ -445,7 +455,7 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 		return requestManager.runRead [ cancelIndicator |
 			val uri = params.textDocument.uri.toUri;
 			val serviceProvider = uri.resourceServiceProvider;
-			val service =  serviceProvider?.get(IDocumentHighlightService);
+			val service = serviceProvider?.get(IDocumentHighlightService);
 			if (service === null)
 				return emptyList
 			
@@ -459,7 +469,7 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 		return requestManager.runRead [ cancelIndicator |
 			val uri = params.textDocument.uri.toUri;
 			val serviceProvider = uri.resourceServiceProvider;
-			val service =  serviceProvider?.get(ICodeActionService);
+			val service = serviceProvider?.get(ICodeActionService);
 			if (service === null)
 				return emptyList
 			
@@ -468,7 +478,7 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 			]
 		]
 	}
-	
+
 	private def void installURI(List<? extends CodeLens> codeLenses, String uri) {
 		for (lens : codeLenses) {
 			if (lens.data !== null) {
@@ -478,7 +488,7 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 			}
 		}
 	}
-	
+
 	private def URI uninstallURI(CodeLens lens) {
 		var URI result = null
 		if (lens.data instanceof String) {
@@ -667,7 +677,6 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 	}
 	
 	override afterBuild(List<Delta> deltas) {
-		if (client instanceof LanguageClientExtensions) {
 		deltas.filter[^new !== null].map[uri.toString].forEach [
 				access.<Void>doRead(it) [ ctx |
 					if (ctx.documentOpen) {
@@ -677,19 +686,21 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 							val serviceProvider = resource.URI.resourceServiceProvider;
 							val coloringService = serviceProvider?.get(IColoringService);
 							if (coloringService !== null) {
-								val doc = ctx.document;
-								val coloringInfos = coloringService.getColoring(resource, doc);
-								if (!coloringInfos.nullOrEmpty) {
-									val uri = resource.URI.toString;
-									client.updateColoring(new ColoringParams(uri, coloringInfos));
+								if (client instanceof LanguageClientExtensions) {
+									val doc = ctx.document;
+									val coloringInfos = coloringService.getColoring(resource, doc);
+									if (!coloringInfos.nullOrEmpty) {
+										val uri = resource.URI.toString;
+										client.updateColoring(new ColoringParams(uri, coloringInfos));
+									}
 								}
 							}
 						}
-						return /*void*/ null;
 					}
+					semanticHighlightingRegistry.update(ctx);
+					return /*void*/ null;
 				];
 			];
-		}
 	}
 
 }
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semanticHighlight/ISemanticHighlightingStyleToTokenMapper.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semanticHighlight/ISemanticHighlightingStyleToTokenMapper.xtend
@@ -1,0 +1,57 @@
+/*******************************************************************************
+ * Copyright (c) 2018 TypeFox and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ide.server.semanticHighlight
+
+import com.google.inject.ImplementedBy
+import com.google.inject.Singleton
+import java.util.List
+import java.util.Set
+import org.eclipse.xtext.ide.editor.syntaxcoloring.IHighlightedPositionAcceptor
+import org.eclipse.xtext.ide.editor.syntaxcoloring.ISemanticHighlightingCalculator
+import org.eclipse.xtext.ide.editor.syntaxcoloring.LightweightPosition
+
+/**
+ * Service for mapping the IDs of the highlighting styles to a list of <a href="https://manual.macromates.com/en/language_grammars">TextMate scopes</a>.
+ * 
+ * @see IHighlightedPositionAcceptor#addPosition
+ * @see LightweightPosition#getIds
+ * 
+ */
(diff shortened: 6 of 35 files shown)


gradle/versions.gradle
```diff
@@ -9,7 +9,7 @@ ext.versions = [
 	'xtext_bootstrap': '2.14.0',
 	'gradle_plugins': '0.1.0',
 	'xtext_gradle_plugin': '1.0.21',
-	'lsp4j': '0.4.0', // dont forget to adapt IdeProjectDescriptor + Tests
+	'lsp4j': '0.5.0-SNAPSHOT', // dont forget to adapt IdeProjectDescriptor + Tests
```
cdietrich: miss adaptation of IdeProjectDescriptor and tests. maybe we can use the snapshot version in the wizard if the Xtext version is a snapshot and the release version if it is a release

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.xtend
```diff
@@ -46,7 +47,7 @@ class CodeActionService implements ICodeActionService {
 				case UNSORTED_MEMBERS: commands += d.fixUnsortedMembers(document, resource, params)
 			}
 		}
-		return commands
+		return commands.map[Either.forLeft(it)];
```
cdietrich: this is just to make it compile right? so we dont consider to support the new api fully?

gradle/versions.gradle
kittaakos: Thanks for the note, correct. As I stated in the PR description, I need to some input on how to deal with the different LSP4J versions.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/SemanticHighlightingManager.xtend
```diff
@@ -0,0 +1,160 @@
...
+		textDocument.version = context.document.version;
+		return new SemanticHighlightingParams(textDocument, lines);
+	}
+
+	protected def void checkInitialized() {
+		Preconditions.checkState(scopes !== null, 'Not initialized.');
+	}
+
+	protected def boolean isNullOrUnknown(List<String> nullable) {
+		return nullable === null || nullable === UNKNOWN_SCOEPS;
```
cdietrich: typo

org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/server/SemanticHighlightingManager.java
```diff
@@ -0,0 +1,246 @@
...
+  
+  /**
+   * Reserved TextMate scope identifier for styles that cannot be handled.
+   */
+  public final static String UNKNOWN_SCOPE = "scope.unknown.xtext";
+  
+  /**
+   * TextMate scopes indicating an unhandled style ID to TextMate scope mapping.
+   */
+  public final static List<String> UNKNOWN_SCOEPS = Collections.<String>unmodifiableList(CollectionLiterals.<String>newArrayList(SemanticHighlightingManager.UNKNOWN_SCOPE));
```
cdietrich: typo

org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/server/SemanticHighlightingManager.java
```diff
@@ -0,0 +1,246 @@
...
+    final Procedure2<List<String>, Integer> _function = (List<String> scope, Integer index) -> {
+      builder.put(index, scope);
+    };
+    IterableExtensions.<List<String>>forEach(IterableExtensions.<List<String>>toSet(scopes), _function);
+    this.scopes = builder.build();
+  }
+  
+  /**
+   * Returns with a list of TextMate scopes for the internal scope index. Returns the
+   * {@link SemanticHighlightingManager#UNKNOWN_SCOEPS unknown scopes} if no scopes are registered for the argument.
```
cdietrich: typo

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/syntaxcoloring/ISemanticHighlightingIdMapper.xtend
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	def List<String> map(String id);
+
+	/**
+	 * The default ID mapper. Does nothing at all but appends the {@link LanguageInfo#getShortName() short name}
+	 * of the language.
+	 * 
+	 * <p>
+	 * For instance, if the ID of the highlighting style is {@code foo} and the name of your language
+	 * is {@code org.your.language.Bar}, then the resulting ID will be {@code foo.Bar.xtext}.
```
cdietrich: is the shortname sufficient enhough? or do you assume its unique?

kittaakos: No, ID uniqueness is not a must.

If a language maps a style string to `['foo', 'bar']` and another language maps a different style string to `['foo', 'bar']` too, then they [will be merged together](https://github.com/eclipse/xtext-core/pull/815/files#diff-4c70201045ee55f33ea5a3df3a6944a1R88) by the highlighting manager.

Here is the corresponding test case: https://github.com/eclipse/xtext-core/pull/815/files#diff-efe3ee25966b9c9a9b9ce3b5f496274eR47

If it needs a bit more explanation, please let me know.

cdietrich: why is the shortname then needed at all?

kittaakos: It is not needed, but it is still easier to debug and track the style-TextMate scope mappings.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -169,6 +175,22 @@ import static org.eclipse.xtext.diagnostics.Severity.*
...
+			val semanticHighlighting = params?.capabilities?.textDocument?.semanticHighlightingCapabilities?.semanticHighlighting ?: false;
+			if (semanticHighlighting) {
+				val scopes = allLanguages
+					.map[get(ISemanticHighlightingCalculator) -> get(ISemanticHighlightingIdMapper)]
+					.filter[key !== null && value !== null]
+					.map[pair | pair.key.highlightingStyles.map[pair.value.map(it)]]
+					.flatten
+					.filter[!nullOrEmpty]
+					.toList;
+				semanticHighlightingManager.initialize(scopes);
```
cdietrich: any reasoning behind the naming scheme. here its called manager for commands its called registry.

kittaakos: No, in fact, I like the registry better. I do not mind renaming it.

cdietrich: @svenefftinge what do you think?

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/TestLanguageRuntimeModule.xtend
```diff
@@ -26,9 +28,13 @@ class TestLanguageRuntimeModule extends AbstractTestLanguageRuntimeModule {
     def Class<? extends ISignatureHelpService> bindSignatureHelpService() {
         return SignatureHelpServiceImpl;
     }
-    
+
     def Class<? extends IColoringService> bindIColoringService() {
     	return ColoringServiceImpl;
     }
 
+	def Class<? extends ISemanticHighlightingCalculator> bindISemanticHighlightingCalculator() {
```
szarnekow: Mixed indentations

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/syntaxcoloring/HighlightingStyles.xtend
```diff
@@ -21,5 +21,21 @@ interface HighlightingStyles {
 	val DEFAULT_ID = "default"
 	val INVALID_TOKEN_ID = "error"
 	val TASK_ID = "task"
+	
+	/**
+	 * A view of all default styles.
+	 *
+	 * @since 2.15
+	 */
+	val ALL_STYLES = #[
```
szarnekow: In the org.eclipse.xtext.ui.editor.syntaxcoloring.IHighlightingConfiguration, we use an acceptor to collect all the known styles, e.g. the preference page in Eclipse is populated from the same source of information as the text attribute provider. Personally I prefer that compared to a static list of things that needs to be maintained. Why did you do it like that?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/syntaxcoloring/ISemanticHighlightingCalculator.java
```diff
@@ -30,4 +32,23 @@
...
+	 * to the corresponding TextMate scopes.
+	 *
+	 * <p>
+	 * Clients can ignore this method and use the default implementation if they do not want to contribute to the Xtext-based
+	 * language server or the LSP-based semantic highlighting is not supported for the language.
+	 *
+	 * @see ISemanticHighlightingIdMapper
+	 * @since 2.15
+	 */
+	default List<String> getHighlightingStyles() {
```
szarnekow: The calculation of the regions is a different aspect compared to a service that provides all styles. I don't see how this method fits into this interface.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/syntaxcoloring/ISemanticHighlightingIdMapper.xtend
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @since 2.15
+ */
+@ImplementedBy(Default)
+interface ISemanticHighlightingIdMapper {
+
+	/**
+	 * Maps the highlighting style ID to the corresponding TextMate scopes.
+	 */
+	def List<String> map(String id);
```
szarnekow: A name 'toScopes' would be more appropriate given the documentation comment.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/syntaxcoloring/ISemanticHighlightingIdMapper.xtend
```diff
@@ -0,0 +1,64 @@
...
+	 * For instance, if the ID of the highlighting style is {@code foo} and the name of your language
+	 * is {@code org.your.language.Bar}, then the resulting ID will be {@code foo.Bar.xtext}.
+	 */
+	@Singleton
+	static class Default implements ISemanticHighlightingIdMapper {
+
+		@Inject
+		Provider<LanguageInfo> info;
+
+		Supplier<String> shortNameSupplier = Suppliers.memoize[info.get.shortName];
```
szarnekow: A setter `@Inject setLanguageInfo(LanguageInfo info) { this.languageShortName = info.shortName }` may be more concise here.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -169,6 +175,22 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 					commands = this.commandRegistry.getCommands()
 				]
 			}
+
+			val semanticHighlighting = params?.capabilities?.textDocument?.semanticHighlightingCapabilities?.semanticHighlighting ?: false;
```
szarnekow: params is already dereferenced unconditionally above. No need for `?.`.
Also: should extract capabilities into a local var and ruse that one.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -169,6 +175,22 @@ import static org.eclipse.xtext.diagnostics.Severity.*
...
+			if (semanticHighlighting) {
+				val scopes = allLanguages
+					.map[get(ISemanticHighlightingCalculator) -> get(ISemanticHighlightingIdMapper)]
+					.filter[key !== null && value !== null]
+					.map[pair | pair.key.highlightingStyles.map[pair.value.map(it)]]
```
szarnekow: Could you please rewrite this and use descriptive names for locals?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/SemanticHighlightingManager.xtend
```diff
@@ -0,0 +1,160 @@
...
+	 * Initializes this manager with all known semantic highlighting TextMate scopes.
+	 */
+	def void initialize(List<List<String>> scopes) {
+		Preconditions.checkState(this.scopes === null, 'Already initialized.');
+		if (scopes === null) {
+			scopes = ImmutableBiMap.of;
+			return;
+		}
+		val builder = ImmutableBiMap.builder;
+		scopes.toSet.forEach [ scope, index |
```
szarnekow: If the scopes should be a unique, why isn't it a Set in the first place?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/SemanticHighlightingManager.xtend
```diff
@@ -0,0 +1,160 @@
...
+		scopes = builder.build;
+	}
+
+	/**
+	 * Returns with a list of TextMate scopes for the internal scope index. Returns the
+	 * {@link SemanticHighlightingManager#UNKNOWN_SCOPES unknown scopes} if no scopes are registered for the argument.
+	 */
+	def List<String> getScopes(int scopeIndex) {
+		checkInitialized;
+		return scopes.getOrDefault(scopeIndex, org.eclipse.xtext.ide.server.SemanticHighlightingManager.UNKNOWN_SCOPES);
```
szarnekow: QualifiedName necessary?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/SemanticHighlightingManager.xtend
```diff
@@ -0,0 +1,160 @@
...
+		textDocument.version = context.document.version;
+		return new SemanticHighlightingParams(textDocument, lines);
+	}
+
+	protected def void checkInitialized() {
+		Preconditions.checkState(scopes !== null, 'Not initialized.');
+	}
+
+	protected def boolean isNullOrUnknown(List<String> nullable) {
+		return nullable === null || nullable === UNKNOWN_SCOPES;
```
szarnekow: You probably want to use `nullable == UNKNOWN_SCOPES`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/SemanticHighlightingManager.xtend
kittaakos: 👍 Xtend refactorings.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/SemanticHighlightingManager.xtend
kittaakos: The order matters, I will check.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/syntaxcoloring/HighlightingStyles.xtend
kittaakos: > we use an acceptor to collect all the known styles
> ...
> Why did you do it like that?

I was not aware of it, thanks for the pointer.

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/TestLanguageRuntimeModule.xtend
kittaakos: It was already mixed, I'll switch to tabs.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/syntaxcoloring/HighlightingStyles.xtend
cdietrich: IHighlightedPositionAcceptor ???

kittaakos: @cdietrich, are you suggesting something or what does this mean? If so, I did not get how does the `IHighlightedPositionAcceptor` solve the problem here. Could you please clarify this? Thanks!

cdietrich: sry confused two problem. i am not sure if the 3 IHighlightingConfigurationAcceptor (2 existing + 1 new) could be unified somehow

kittaakos:
I have made a few changes based on the feedback. Please check. Thank you!

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semanticHighlight/ISemanticHighlightingStyleToTokenMapper.xtend
```diff
@@ -0,0 +1,58 @@
...
+	 * Must not return with {@code null} but an empty set instead.
+	 * 
+	 * @see IHighlightedPositionAcceptor#addPosition
+	 */
+	def Set<String> getAllStyleIds();
+
+	@Singleton
+	static class Noop implements ISemanticHighlightingStyleToTokenMapper {
+
+		override toScopes(String id) {
```
cdietrich: param should be styleId

cdietrich:
i am not sure if we should expose this stuff as api yet (`@since`)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semanticHighlight/SemanticHighlightingRegistry.xtend
```diff
@@ -0,0 +1,220 @@
...
+				return new HighlightedRange(start, end, scope);
+			]
+		].flatten;
+
+		val lines = ranges.toLines(document);
+		val textDocument = context.toTextDocument;
+		notifyClient(new SemanticHighlightingParams(textDocument, lines));
+	}
+
+	protected def List<SemanticHighlightingInformation> toLines(Iterable<? extends HighlightedRange> ranges, Document document) {
```
cdietrich: am not sure if this is better called toSemanticHighlightingInformation(s)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semanticHighlight/SemanticHighlightingRegistry.xtend
```diff
@@ -0,0 +1,220 @@
...
+			}
+		];
+		return builder.build.asMap.entrySet.map [
+			val line = key;
+			val tokens = value;
+			new SemanticHighlightingInformation(line, tokens.encode);
+		].toList;
+	}
+
+	protected def VersionedTextDocumentIdentifier toTextDocument(ILanguageServerAccess.Context context) {
```
cdietrich: rename toVersionedTextDocumentIdentifier and move as member to context?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semanticHighlight/SemanticHighlightingRegistry.xtend
```diff
@@ -0,0 +1,220 @@
...
+		return new VersionedTextDocumentIdentifier => [
+			uri = context.resource.URI.toUriString;
```
cdietrich: do you know why the context does not store the original uri. if yes we could move this as member to context

cdietrich:
should we mark IColoringService and as deprecated?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semanticHighlight/SemanticHighlightingRegistry.xtend
kittaakos: No, I have no idea.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semanticHighlight/SemanticHighlightingRegistry.xtend
```diff
@@ -0,0 +1,221 @@
...
+import static extension org.eclipse.lsp4j.util.SemanticHighlightingTokens.encode
+
+/**
+ * Shared semantic highlighting manager per language server.
+ * Responsible for converting the semantic highlighted ranges into the LSP standard by producing a compact,
+ * {@code base64} encoded token string.
+ * 
+ */
+@Log
+@Singleton
```
kittaakos: It must not be `@Singleton`. Otherwise, the initialize method will throw an ISE when the client tries to reconnect. I am verifying this locally and will update the PR according to it.

kittaakos: Yes, it is a bug.

spoenemann:
> can you please advise how to tackle the 0.4.0 vs. the 0.5.0 LSP4J version mismatch?

See #793: I'll try to make an LSP4J milestone next week that can be consumed by Xtext.

kittaakos:
> It does not support deltas.

It won't work like that. We either support the delta with the Xtext LS too, or the _delta_/_non-delta_ capability information must be part of the `InitializeResult`. The client should behave differently when it receives delta tokens and when the full token set. (Or we send an empty token for each line that does not have highlighted ranges, but that smells like a massive hack.)

The problem is the followings.

Let assume, we have a simple DSL, and we want to color the name of the operation. Nothing more:
```
type A {
  op foo() {}
  op foo(a: A): string {}
}
```

The client will get two ranges representing both `foo`s. After commenting out the first `foo`, we send only one token (with the second one) as the notification to the client. So the client does not know, the first `foo` must not be highlighted anymore. With the JDT LS, where we support the delta tokens, we send out an "empty" info, and the client knows the state of that line has to be discarded.

For now, I go with the delta to see how it works. But adding this to the LSP would be better I believe.

CC: @svenefftinge

svenefftinge:
> Or we send an empty token for each line that does not have highlighted ranges, but that smells like a massive hack.

I don't think that is a massive hack, but simply a full update for the document, no?

kittaakos:
> I don't think that is a massive hack, but simply a full update for the document, no?

👍 Yes, it makes sense. Thanks for the feedback.

kittaakos:
I have updated the Xtext LS to reflect the [most recent state](https://github.com/eclipse/lsp4j/commit/721d5ab6345d3f2cc756c629af9c10df417b3c6e) of the lsp4j API. This PR has no intention to fully support those changes, just to compile. This PR depends on #793.

cdietrich:
@kittaakos @spoenemann looks like there are still compile errors:

ERROR:Type mismatch: cannot convert from CompletableFuture<Either<List<? extends SymbolInformation>, List<? extends DocumentSymbol>>> to CompletableFuture<List<Either<SymbolInformation, DocumentSymbol>>> (file:/jobs/genie.xtext/xtext-core/branches/PR-815/workspace/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend line : 395 column : 10)

cdietrich:
@kittaakos can you switch to the milestone, and use it in the wizard (IdeProjectDescriptor) too (+adapt wizard test = CliWizardIntegrationTest)

kittaakos:
> can you please switch to the milestone, and use it in the wizard (IdeProjectDescriptor) too (+adapt wizard test = CliWizardIntegrationTest)

OK.

kittaakos:
> switch to the milestone, and use it in the wizard (IdeProjectDescriptor) too (+adapt wizard test = CliWizardIntegrationTest)

Done. Let me know if I have overlooked something.

 - [Switched to the milestone version of LSP4J for Gradle](https://github.com/eclipse/xtext-core/pull/815/files#diff-25205fe6f2e0d330abcbc45df6b96797R12).
 - [The version change in the `IdeProjectDescriptor`](https://github.com/eclipse/xtext-core/pull/815/files#diff-cb8a0173e26e6832fc89c0a814a4f00aR58).
 - Adapted the `pom.xml`s for the tests: [here](https://github.com/eclipse/xtext-core/pull/815/files#diff-3d7dbd491e9741b48e263723ae7ccb7dR21) and [here](https://github.com/eclipse/xtext-core/pull/815/files#diff-c1367418cf505ba438ca3c5bd69dc86aR21).

cdietrich:
wha about org.eclipse.xtext.ide/META-INF/MANIFEST.MF ?

kittaakos:
Good point, thanks! What should be the desired range? I will set it.

cdietrich:
and org.eclipse.xtext.testing/META-INF/MANIFEST.MF

cdietrich:
am not sure @spoenemann what do you propose?

spoenemann:
I'd keep a range like now: `[0.5.0,0.6.0)`

kittaakos:
I have switched to the `[0.5.0,0.6.0)` range in the [`MANIFEST.MF`](
https://github.com/eclipse/xtext-core/pull/815/files#diff-4c56e69b8720d8334edc4120b9c1cf7e) and I have also declared the same range in the [`o.e.x.ide.test`](
https://github.com/eclipse/xtext-core/pull/815/files#diff-6f80aa5d880a88ece1a342e501b6598d).

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/TestLanguageRuntimeModule.xtend
```diff
@@ -23,12 +27,20 @@ class TestLanguageRuntimeModule extends AbstractTestLanguageRuntimeModule {
...
-    }
-    
-    def Class<? extends IColoringService> bindIColoringService() {
-    	return ColoringServiceImpl;
-    }
+	def Class<? extends ISignatureHelpService> bindSignatureHelpService() {
+		return SignatureHelpServiceImpl;
+	}
+
+	def Class<? extends IColoringService> bindIColoringService() {
```
holgerschill: Is this binding still needed. I assume we don't need it.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -677,19 +686,21 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 							val serviceProvider = resource.URI.resourceServiceProvider;
 							val coloringService = serviceProvider?.get(IColoringService);
 							if (coloringService !== null) {
```
holgerschill: I think this needs to be removed since IColoringService is deprecated and then we don't need the binding in the TestLanguageModule anymore.

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/TestLanguageRuntimeModule.xtend
kittaakos: Why not? Although the service is deprecated we did not remove it; hence it makes sense to test it, right?

holgerschill: That would imply that we keep the call in the LanguageServerImpl. I would prefer to remove it. Any reason besides testing against that?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
kittaakos: > I think this needs to be removed

"this" what? Did you mean the`null` check or the entire logic?

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/TestLanguageRuntimeModule.xtend
kittaakos: If we remove the call from the LS, why do we keep the `IColoringService` at all? I mean, what is the benefit of a deprecated interface (and its default implementation) if it is not called from anywhere?

holgerschill: Ok. If someone has used it then we still need to call it in LS - fair enough.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
holgerschill: "this" means the entire logic but as mentioned above -> Ok. If someone has used it then we still need to call it in LS. fair enough.

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.xtend
kittaakos: See: #821.

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/editor/syntaxcoloring/SemanticHighlightingStyleToTokenMapper.xtend
```diff
@@ -0,0 +1,22 @@
...
+
+import org.eclipse.xtext.ide.server.semanticHighlight.ISemanticHighlightingStyleToTokenMapper
+
+class SemanticHighlightingStyleToTokenMapper implements ISemanticHighlightingStyleToTokenMapper {
+
+	override toScopes(String styleId) {
+		return #[styleId];
+	}
+
+	override getAllStyleIds() {
```
szarnekow: Do you have plans to work on a different impl of getAllStyleIds, e.g. similar to how we  collect the style ids in the eclipse case?

kittaakos: No, not in the scope of this PR. As I see, even if we would come up with an adapter layer (to collect all the style IDs just like in the Eclipse-case), the actual implementations are in the UI code for the downstream projects. So it would be useless. But, if you have already something in mind, a silver-bullet solution, I am not against implementing it, so please share your idea.

kittaakos: CC: @meysholdt

szarnekow: My main concern is not about the reuse of some piece of code, but the general structure of the API. A method `getAllStyleIds` that has to be in sync with the style ids anyways, is prone to be forgotten to override. I'd like to reduce the risk and the burden on the implementor by collecting the style ids generically.

szarnekow: Anyways, I'm not passionate about this. If this is the API you are happy with, than I'm ok with it, too.

PR #820, opened 2018-08-17
Xtext core 69 content assist on syntax errors

ArneDeutsch:


cdietrich:
So you did not find any Solution?

ArneDeutsch:
Yes, the commit https://github.com/eclipse/xtext-core/pull/820/commits/35582ebd66efca10a718ce878f455bedcff910a9 fixes the bug and all tests are green. But it is quite fragile, hence we decided to revert.

Issue #771, opened 2018-06-22
ValidatorFragment2 generates an overridden version of addComboBox for Xbase languages


LorenzoBettini:
Xbase already defines this method 

```
org.eclipse.xtext.xbase.ui.validation.XbaseValidationConfigurationBlock.addComboBox(String, String, Composite, int)
```

but the configuration block generated for Xbase languages overrides this method, just like for any other languages (indeed there's a warning in the generated configuration block, though src-gen is usually configured to exclude additional warnings).

Is this the intended behavior? Shouldn't Xbase languages reuse the one from XbaseValidationConfigurationBlock? @holgerschill

cdietrich:
for me this looks like a oversight.

cdietrich:
@holgerschill  any news on this issue?

cdietrich:
ping @holgerschill

holgerschill:
Fixed in 2.15

PR #840, opened 2018-08-28
ValidatorFragment2 generates an overridden version of addComboBox for…

holgerschill:
… Xbase languages #771


holgerschill:
Thre is an open PR in xtext-eclipse#808 that depends on this PR after all.

PR #843, opened 2018-08-29
[eclipse/xtext-eclipse#809] Cache task marker pattern for performance

ArneDeutsch:

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/tasks/DefaultTaskParser.xtend
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/tasks/DefaultTaskParser.xtend
@@ -8,22 +8,21 @@
 package org.eclipse.xtext.tasks
 
 import java.util.HashMap
-import java.util.regex.Pattern
-import org.eclipse.xtext.util.Strings
 import java.util.Map
+import org.eclipse.xtext.util.Strings
 
 /**
  * @author Stefan Oehme - Initial contribution and API
  * @since 2.6
  */
 class DefaultTaskParser implements ITaskParser {
-
+	
 	override parseTasks(String source, TaskTags taskTags) {
 		if (taskTags.empty) return #[]
 
 		val taskTagsByName = getTaskTagsByName(taskTags)
 
-		val matcher = taskTags.toPattern.matcher(source)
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/tasks/DefaultTaskParser.xtend
```diff
@@ -8,15 +8,17 @@
...
 import org.eclipse.xtext.util.Strings
-import java.util.Map
 
 /**
  * @author Stefan Oehme - Initial contribution and API
  * @since 2.6
  */
 class DefaultTaskParser implements ITaskParser {
+	
+	var patternCache = new HashMap<TaskTags, Pattern>
```
kthoms: could use val here

org.eclipse.xtext/src/org/eclipse/xtext/tasks/DefaultTaskParser.xtend
```diff
@@ -60,11 +62,16 @@ class DefaultTaskParser implements ITaskParser {
 	}
 
 	protected def toPattern(TaskTags taskTags) {
```
kthoms: Move Pattern into TaskTags as field, would make cache obsolete

ArneDeutsch: Done.

org.eclipse.xtext/xtend-gen/org/eclipse/xtext/tasks/DefaultTaskParser.java
```diff
@@ -39,7 +37,7 @@
         return Collections.<Task>unmodifiableList(CollectionLiterals.<Task>newArrayList());
       }
       final Map<String, TaskTag> taskTagsByName = this.getTaskTagsByName(taskTags);
-      final Matcher matcher = this.toPattern(taskTags).matcher(source);
+      final Matcher matcher = taskTags.toPattern().matcher(source);
```
szarnekow: If we do no longer call toPattern, we do not need to keep that method.

PR #854, opened 2018-09-02
[#793] update to lsp4j 0.5.0 final

cdietrich:
[#793] update to lsp4j 0.5.0 final


cdietrich:
needs to be tested once the releas is out

Issue #858, opened 2018-09-04
[LS] Support the hierarchical document symbols


kittaakos:
Part of the issue description was copied from [here](
https://github.com/eclipse/xtext-core/issues/821#issuecomment-418292716).

 - https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol
 - "_The client support hierarchical document symbols._"
 - "_[...] Document symbols can be hierarchical and they have two ranges [...]_"
 - `TextDocumentClientCapabilities.documentSymbol.hierarchicalDocumentSymbolSupport`
 - The API is already [available in LSP4J](https://github.com/eclipse/lsp4j/blob/02ddbda6e65d23a7f31c7eaf6f7e81263d2ecc8c/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/Protocol.xtend#L463) (`0.5.0`).

cdietrich:
it would be nice if the api could be designed in a way that is independent from lsp so that the same implementation can be used in eclipse and lsp via adapters

PR #862, opened 2018-09-06
GH-858: [LS] Added support for the hierarchical document symbols.

kittaakos:
Closes: eclipse/xtext-eclipse#858.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/DocumentExtensions.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/DocumentExtensions.xtend
@@ -75,6 +75,18 @@ class DocumentExtensions {
 		return resource.newLocation(textRegion)
 	}
 
+	/**
+	 * Returns with the {@link Location location} that represents the {@link ILocationInFileProvider#getFullTextRegion full text region}
+	 * of the argument.
+	 * 
+	 * @since 2.16
+	 */
+	def Location newFullLocation(EObject object) {
+		val resource = object.eResource
+		val textRegion = locationInFileProvider.getFullTextRegion(object)
+		return resource.newLocation(textRegion)
+	}
+
 	def Location newLocation(EObject owner, EStructuralFeature feature, int indexInList) {
 		val resource = owner.eResource
 		val textRegion = locationInFileProvider.getSignificantTextRegion(owner, feature, indexInList)
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
@@ -86,6 +86,8 @@ import org.eclipse.xtext.ide.server.rename.IRenameService
 import org.eclipse.xtext.ide.server.semanticHighlight.SemanticHighlightingRegistry
 import org.eclipse.xtext.ide.server.signatureHelp.ISignatureHelpService
 import org.eclipse.xtext.ide.server.symbol.DocumentSymbolService
+import org.eclipse.xtext.ide.server.symbol.HierarchicalDocumentSymbolService
+import org.eclipse.xtext.ide.server.symbol.IDocumentSymbolService
 import org.eclipse.xtext.ide.server.symbol.WorkspaceSymbolService
 import org.eclipse.xtext.resource.IResourceDescription.Delta
 import org.eclipse.xtext.resource.IResourceServiceProvider
@@ -395,7 +397,7 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 		return requestManager.<List<Either<SymbolInformation, DocumentSymbol>>>runRead[ cancelIndicator |
 			val uri = params.textDocument.uri.toUri
 			val resourceServiceProvider = uri.resourceServiceProvider
-			val documentSymbolService = resourceServiceProvider?.get(DocumentSymbolService)
+			val documentSymbolService = resourceServiceProvider.IDocumentSymbolService
 			if (documentSymbolService === null) {
 				return emptyList
 			}
@@ -404,6 +406,25 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 				return documentSymbolService.getSymbols(document, resource, params, cancelIndicator)
 			]
 		]
+	}
+
+	/**
+	 * @since 2.16
+	 */
+	protected def IDocumentSymbolService getIDocumentSymbolService(IResourceServiceProvider serviceProvider) {
+		if (serviceProvider === null) {
+			return null;
+		}
+		val documentSymbolServiceClass = if(hierarchicalDocumentSymbolSupport) HierarchicalDocumentSymbolService else DocumentSymbolService;
+		return serviceProvider.get(documentSymbolServiceClass);
+	}
+
+	/**
+	 * {@code true} if the {@code TextDocumentClientCapabilities} explicitly declares the hierarchical document symbol support
+	 * at LS initialization time. Otherwise, false.
+	 */
+	protected def boolean isHierarchicalDocumentSymbolSupport() {
+		return this.params.capabilities?.textDocument?.documentSymbol?.hierarchicalDocumentSymbolSupport ?: false;
 	}
 
 	override symbol(WorkspaceSymbolParams params) {
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/HierarchicalDocumentSymbolService.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/HierarchicalDocumentSymbolService.xtend
@@ -1,0 +1,106 @@
+/*******************************************************************************
+ * Copyright (c) 2018 TypeFox and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ide.server.symbol
+
+import com.google.common.annotations.Beta
+import com.google.common.base.Optional
+import com.google.inject.Inject
+import com.google.inject.Singleton
+import java.util.Iterator
+import org.eclipse.emf.ecore.EObject
+import org.eclipse.emf.ecore.resource.Resource
+import org.eclipse.emf.ecore.util.EcoreUtil
+import org.eclipse.lsp4j.DocumentSymbol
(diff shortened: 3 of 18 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/IDocumentSymbolService.xtend
```diff
@@ -0,0 +1,47 @@
...
+ * 
+ * <p>
+ * This interface is not meant to be injected. Use the {@link DocumentSymbolService} and the {@link HierarchicalDocumentSymbolService} directly instead.
+ * 
+ * @since 2.16
+ */
+@ImplementedBy(Noop)
+interface IDocumentSymbolService {
+
+	def List<Either<SymbolInformation, DocumentSymbol>> getSymbols(Document document, XtextResource resource,
```
cdietrich: this is one of the apis where i am not sure if i like it. e.g. this could not be used to produce an eclipse outline without using lsp4j

kittaakos: I am not against modifying, adjusting it, but I would like to understand what is wrong with the `lsp4j` model API? If I understood you correctly, you would like to introduce another API in `xtext-core` that would be used by `xtext-eclipse` and will be transformed into the [`IOutlineNode`](https://github.com/eclipse/xtext-eclipse/blob/master/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/outline/IOutlineNode.java), right? But `xtext-eclipse` already has `lsp4j` as a dependency, right? Then why to add another API?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/HierarchicalDocumentSymbolService.xtend
```diff
@@ -0,0 +1,108 @@
...
+
+		return getSymbols(resource, cancelIndicator);
+	}
+
+	def getSymbols(XtextResource resource, CancelIndicator cancelIndicator) {
+		val allSymbols = newHashMap;
+		val rootSymbols = newArrayList;
+		val itr = getAllContents(resource);
+		while (itr.hasNext) {
+			operationCanceledManager.checkCanceled(cancelIndicator);
```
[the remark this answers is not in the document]
kittaakos: Sure, we can do this but let's move this conversation to [here](
https://github.com/eclipse/xtext-core/pull/862#discussion_r215621247). I think they are related.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/IDocumentSymbolService.xtend
cdietrich: no it does not thats the point

cdietrich: => for api that might be useful toi be reused in eclipse it might help to have a lsp4j independent api e.g. as we do it for content assist

cdietrich: => we should decide that at per api level if it can be usefully used in more than the lsp4j context. currently we are inconsistent with that cc @tivervac

cdietrich: => i want at least having this thought over before introducing a new api that can be used from lsp4j only

kittaakos: Also, please explain to me, what would the user do with Xtext `2.16`? Let assume, there is this new API but they already have the outline implementation in `xtext-eclipse`. What would they do then? I am obviously missing a step... That's why I cannot see the benefit of introducing this new non-`lsp4j` API.

cdietrich: this is about new projects that target lsp4j and eclipse. right now they would have toi implement the outline twice

kittaakos: > no it does not thats the point

So `xtext-eclipse` does not have the `lsp4j` dependency; I was not aware of it. Now I see your point. Sure, let's do it, I do not mind introducing a new API here, and the LS can map it to LSP4J and Eclipse can transform it to `IOutlineNode`s.

cdietrich: see discussion at https://github.com/eclipse/xtext-eclipse/issues/711#issuecomment-419072815

cdietrich: if we come to the conclusion that it is impossible to come up with a lsp4j independent api that would be ok but then i would prefer to implement it twice

svenefftinge: In Xtext we always allowed different levels of abstraction. We need this LSP-level. If you want something on top that abstracts over LSP, please build it. But it doesn't need to be part of this PR.

cdietrich: i can live with that if we dont offer this as api to be used in eclipse

svenefftinge: Ok

kittaakos: > i can live with that if we dont offer this as api to be used in eclipse

Should I remove the `@since 2.16` from the source? How should we proceed with this PR?

Although the PR is not super urgent, I would like to get this done if you do not have any objections. Thanks!

One thing that we have to keep in mind before we merge is the LSP4J API. See here https://github.com/eclipse/lsp4j/issues/252.

cdietrich: we marked provisional api as `@Beta` before. correct @szarnekow @svenefftinge ?

svenefftinge: Yes, we did that in some places.

svenefftinge (approved):

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/DocumentExtensions.xtend
```diff
@@ -75,6 +75,12 @@ class DocumentExtensions {
 		return resource.newLocation(textRegion)
 	}
 
+	def Location newFullLocation(EObject object) {
```
kthoms: missing `@since 2.16` here

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -406,6 +408,22 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 		]
 	}
 
+	protected def IDocumentSymbolService getIDocumentSymbolService(IResourceServiceProvider serviceProvider) {
```
kthoms: also add `@since` for new protected methods in public API packages

org.eclipse.xtext.testing/xtend-gen/org/eclipse/xtext/testing/DocumentSymbolConfiguraiton.java
```diff
@@ -31,11 +33,11 @@ public void setExpectedSymbols(final String expectedSymbols) {
   }
   
   @Pure
-  public Procedure1<? super List<? extends SymbolInformation>> getAssertSymbols() {
+  public Procedure1<? super List<Either<SymbolInformation, DocumentSymbol>>> getAssertSymbols() {
```
kthoms: add `@since`

org.eclipse.xtext.testing/xtend-gen/org/eclipse/xtext/testing/DocumentSymbolConfiguraiton.java
```diff
@@ -31,11 +33,11 @@ public void setExpectedSymbols(final String expectedSymbols) {
...
     return this.assertSymbols;
   }
   
-  public void setAssertSymbols(final Procedure1<? super List<? extends SymbolInformation>> assertSymbols) {
+  public void setAssertSymbols(final Procedure1<? super List<Either<SymbolInformation, DocumentSymbol>>> assertSymbols) {
```
kthoms: add `@since`

kittaakos:
Thanks for the feedback on this PR. One general question, shall we use [`@Beta`](
https://github.com/eclipse/xtext-core/pull/862#discussion_r217642497) or [`@since`](
https://github.com/eclipse/xtext-core/pull/862#discussion_r218820251). Thanks!

kthoms:
I'd say a class once made public available can't be `@Beta`, and adding API methods can't either. And if I see this correctly, `@Beta` is only used on type level in the codebase.

So adding API methods to API classes can't be beta, thus we have to be carefully.

cdietrich:
No Beta says available but not API

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -682,7 +702,7 @@ class DocumentHighlightConfiguration extends TextDocumentPositionConfiguration {
 @Accessors
 class DocumentSymbolConfiguraiton extends TextDocumentConfiguration {
```
kthoms: Typo in class name :(

kittaakos:
> So marking as beta can give users a hint on that

👍 

Can we agree that I should get rid of `@since` and use `@Beta` instead?

kthoms:
So API rule incompatibility is blocking this change? If so, please veto @cdietrich .

cdietrich:
No it’s new code. Simply mark it as beta

kittaakos:
> No it’s new code. Simply mark it as beta

This.

Thanks, guys! I apply the changes soon.

kittaakos:
I made the changes.

kittaakos:
If this PR is good to merge, please push the button. Thank you!

PR #880, opened 2018-09-23
[eclipse/xtext#1282] Upgrade to EMF 2.15

kthoms:
Using latest EMF artifacts from Maven Central


cdietrich:
I miss exclusions for bad transitive dependencies as we have it for jdt

kthoms:
Where would you expect them?

cdietrich:
There where the different emf plugins are used. See places for jdt. Let gradle plaint transitive dependencies. Make sure the plugins that should be used from platform are actually used

cdietrich:
Alternatively do this after the bom update (am not sure we find time to actually test this in time, besides me all of the Xtext team did nothing to actually try and test it or find better solution)

szarnekow:
This should be merged after the bom stuff.

cdietrich:
obsolete

PR #886, opened 2018-10-01
[539677] getValue should return null for null arg

kthoms:

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/SimpleAttributeResolver.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/SimpleAttributeResolver.java
@@ -73,7 +73,7 @@ public class SimpleAttributeResolver<K extends EObject, T> implements Function<K
 	}
 
 	public T getValue(final K object) {
-		return valueCache.get(object);
+		return object != null ? valueCache.get(object) : null;
 	}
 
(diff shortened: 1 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=539677

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=539677

cdietrich:
@szarnekow am not sure if this change is a good idea since it may cover user code errors. what do you think

kthoms:
All clients in the Xtext code base are safe to retrieve null. Clients would have to catch the NPE that they would have experienced. I consider this as improvement. But good to have other viewpoints.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/SimpleAttributeResolver.java
```diff
@@ -73,7 +73,7 @@ public T apply(EObject param) {
 	}
 
 	public T getValue(final K object) {
-		return valueCache.get(object);
+		return object != null ? valueCache.get(object) : null;
```
szarnekow: Why was null passed in here at all?

PR #889, opened 2018-10-02
[eclipse/xtext#1310] Set encoding for Test tasks

kthoms:
Test failures can happen on machines that do not define a LANG environment property


kthoms:
We'll try to fix all related test cases to ensure to work with the required encoding instead. Closing.

PR #894, opened 2018-10-08
Initial Diagnostics can never arrive to the client

dhuebner:


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -194,6 +196,12 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 		], []).thenApply [result]
 	}
 	
+	override void initialized(InitializedParams params) {
+		requestManager.runWrite([
```
cdietrich: why is the runWrite needed here?

dhuebner: now directly calling` initialized.complete()`

dhuebner:
@cdietrich @svenefftinge  Thanks for reviewing!

PR #895, opened 2018-10-08
Ignore didChangeWatchedFiles for open documents

dhuebner:


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -277,15 +278,20 @@ import org.eclipse.lsp4j.InitializedParams
...
 				} else {
-					dirtyFiles += toUri(fileEvent.uri)
+					dirtyFiles += key
 				}
-			}
-			workspaceManager.didChangeFiles(dirtyFiles, deletedFiles)
-		], [cancelIndicator , buildable | 
+			]
+			var Buildable buildable = []
+			if (!deletedFiles.isEmpty || !dirtyFiles.isEmpty)
```
cdietrich: switching order would be consistent

cdietrich:
@dhuebner could you test that with vscode.
see 

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/ServerTest.xtend / testIncrementalDeletion

could we add a testcase there with open vs non open

dhuebner:
@cdietrich 

> could you test that with vscode.

I don't know how to test something with vscode. Is there a description somewhere? 

> could we add a testcase there with open vs non open

Will look in to it. Hopefully there is api to simply open an editor

cdietrich:
(1) have a sample dsl with file (a) references files (b)
(2) have file (a) and (b) open
(3) delete (b) on disk - no errors
(4) close editor for (b) then (a) becomes red

dhuebner:
@cdietrich I have a theia app, would it also work as a test platform?

cdietrich:
what i dont know: how is this stuff implemented in different platforms.
what i want to know: do we make sure no stale files stay in index caused by this change.
we had problems with that before and i dont want it again

dhuebner:
> what i dont know: how is this stuff implemented in different platforms.

They all should follow the protocol. I will write a junit tests as described in 
https://github.com/eclipse/xtext-core/pull/895#issuecomment-428083793 and than test the PR in a theia app. That should be enough

> what i want to know: do we make sure no stale files stay in index caused by this change.

There are tests for that I think. org.eclipse.xtext.ide.tests.server.OpenDocumentTest . The new test with open/close/build will additionally cover the case.

> we had problems with that before and i don't want it again

If the problem was solved there should be tests for that, if not we add them.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/ServerTest.xtend
```diff
@@ -100,7 +100,34 @@ class ServerTest extends AbstractTestLangLanguageServerTest {
...
+		assertTrue("Bar.testlang references type Foo from Foo.testlang: expect no error",
+			diagnostics.get(referencingFileURI).empty)
+
+		fileURI.deleteFile
+		languageServer.getWorkspaceService.didChangeWatchedFiles(
+			new DidChangeWatchedFilesParams(#[new FileEvent(fileURI, FileChangeType.Deleted)])
+		)
+		assertTrue("delete file on disk: expect no error", diagnostics.get(referencingFileURI).empty)
+		close(fileURI);
+		assertFalse("close deleted file: expect error", diagnostics.get(referencingFileURI).empty)
```
cdietrich: :+1:

dhuebner: Build is passing now. should I merge?

Issue #885, opened 2018-10-01
Make LSP server shutdown configurable


kaikreuzer:
We (that is the Eclipse SmartHome project) recently upgraded to Xtext 2.14 and now ran into the issue that [our server is silently shutting down](https://github.com/eclipse/smarthome/issues/6291), if VS Code is disconnecting after having used LSP.
We tracked that down to https://github.com/eclipse/xtext-core/pull/474 and being a "feature" of an LSP server. While I see that this is a part of the specification and thus has to be supported, it is nonetheless an absolutely no-go to call `System.exit()` from within some library, especially within an OSGi container.
I guess our situation, where LSP is run as a service on an existing server (and not as a JVM process on its own), clearly shows that this is a pretty nasty behavior.

Imho, the clean way to implement this in an OSGi way would be to provide a hook for listening to exit requests and have a default implementation of such a listener, which cleanly shuts down the OSGi container, while giving consumers the chance to deactivate/replace this default implementation (e.g. through config admin). Wdyt?

kthoms:
I understand your concern, but as you already found out it is a feature. Your requirement is to shut down the OSGi container then, and a listener could be a way to do so. However, the Xtext core bundles are not aware of OSGi and don't have a dependency on that. And I do not think that one should be introduced. This means it must be somehow added from outside.

I could imagine to delegate the behaviour to a shutdown handle that could be overridden by a specific language. Or do you have an idea how to shut down the OSGi container without a dependency here?

kaikreuzer:
> Your requirement is to shut down the OSGi container then

No, not at all - at most, the LSP server bundles should stop their services and must not impact any other code running on the OSGi container. Btw, just for my understanding, what is the background of this "feature"? If the server is stopped upon closing the editor, how is it started when opening the editor again?

> This means it must be somehow added from outside.

Well, the concept of registering listeners (`addExitNotificationListener` method) isn't related to OSGi, so this "logic" can easily be drawn out of the code code. And the default listener implementation could e.g. be deactivated by a system property (to not have any OSGi dependency either).

Btw, in any case, the code should properly handle Security exceptions. When using a Java security manager, you currently get something like this in your log:
```
Oct 01, 2018 9:25:16 AM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint handleNotification
WARNING: Notification threw an exception: {
  "jsonrpc": "2.0",
  "method": "exit",
  "params": null
}
[...]

PR #896, opened 2018-10-09
[#885] make LSP server exit behaviour configurable

cdietrich:

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ILanguageServerShutdownAndExitHandler.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ILanguageServerShutdownAndExitHandler.java
@@ -1,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ide.server;
+
+import com.google.inject.ImplementedBy;
+
+/**
+ * @author dietrich - Initial contribution and API
+ * 
+ * @since 2.16
+ */
+@ImplementedBy(ILanguageServerShutdownAndExitHandler.DefaultImpl.class)
+public interface ILanguageServerShutdownAndExitHandler {
+
+	/**
+	 * Callback that is called when Language Servers <code>exit</code> is called.
+	 */
+	void exit();
+	
+	/**
+	 * Callback that is called when Language Servers <code>shutdown</code> is called.
+	 */
+	void shutdown();
+	
+	/**
+	 * Implementation for {@link ILanguageServerShutdownAndExitHandler} that does nothing.
+	 */
+	public static class NullImpl implements ILanguageServerShutdownAndExitHandler {
+
+		@Override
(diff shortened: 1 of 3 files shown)


cdietrich:
@kaikreuzer is this what you were thinking of?
@svenefftinge please review

kaikreuzer:
Yes, thanks, that should do the trick!
Just one question: If I am not mistaken, the termination is a two step process: shutdown & exit.
So would we need a similar mechanism for the shutdown step as well? As mentioned [here](https://github.com/eclipse/smarthome/issues/6291#issuecomment-425843106), I once had an issue to reconnect to the server after avoiding the exit call as it was replying that it is currently shutting down and won't allow any connections.

cdietrich:
since i dont have your usecase we would need more detailed input on that.

kaikreuzer:
I was hoping that you knew what exactly the LSP server is doing when receiving the [shutdown request](https://microsoft.github.io/language-server-protocol/specification#shutdown) and whether this leaves it in an unusable state or not.

cdietrich:
i have heavy doubts it leaves a recoverable states. it is a noop. (just sets a boolean)

kaikreuzer:
> i have heavy doubts it leaves a recoverable state

You mean you are pretty sure it leaves a recoverable state as nothing really happens upon such a request? Then all should be fine.

Unfortunately, I didn't capture the exact exception that I had seen at that time - maybe it didn't say that "shutdown has been called already", but that ["the server has already been initialized"](https://github.com/eclipse/xtext-core/blob/ceca3728be5b57e33a6727509128a482a483f0b1/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend#L138)? I'm not sure, just wanted to mention it that the re-connection should be tested.

cdietrich:
no i mean: shutdown is i a noop. code is not tested with other scenarios.
=> we require your testing/input

kaikreuzer:
> i am 100% sure a new try will fail

Why?

Since I merged my workaround with https://github.com/openhab/openhab-core/pull/409, I didn't get any negative feedback from the users - it all seems to work smoothly. So having shutdown a noop and exit ignored seems to be ok.

cdietrich:
i am not sure if e.g. the  shutdowncalled flag is properly reset

cdietrich:
and it will explode if you call initialize twice

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ILanguageServerExitHandler.java
```diff
@@ -0,0 +1,49 @@
...
+ * 
+ * @since 2.16
+ */
+@ImplementedBy(ILanguageServerExitHandler.DefaultImpl.class)
+public interface ILanguageServerExitHandler {
+
+	/**
+	 * Callback that is called when Language Servers <code>exit</code> is called.
+	 */
+	void exit(int status);
```
svenefftinge: Shouldn't we also add the `shutdown` method here and remove the argument?

cdietrich: will have a look

cdietrich:
@svenefftinge updated the PR. please review again

svenefftinge:
Ideally the hook would not be provided through DI but per configuration, so that when using the launcher we would not need to call `System.exit` but would simply stop listening and naturally terminate.

cdietrich:
but wont that be then up to LSP4j and the launchers there and default impl in the shutdown and exit methods?

Issue #864, opened 2018-09-11
Find a replacement for MergableManifest that works fine with java 9+


cdietrich:
- Find a replacement for MergableManifest that works fine with java 9+
- Mark the old one as deprecated

cdietrich:
cc  @ArneDeutsch

ArneDeutsch:
New implementation with high test code ceoverage is in place.

PR #913, opened 2018-10-17
#864: Rewrite MergeableManifest to MergeableManifest2 to avoid reflection

ArneDeutsch:
Adding extensive unit tests to ensure behaviour does not change.
Use only MergableMenufest2 internally.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest.java
@@ -33,7 +33,10 @@ import com.google.common.collect.Lists;
  * @author Sven Efftinge - Initial contribution and API
  * @author Jan Koehnlein - merging of parameterized entries
  * @since 2.2 moved to org.eclipse.xtext.util
+ * @deprecated Uses reflection which will not work well with future java
+ * versions. Use {@link MergeableManifest2} instead.
  */
+@Deprecated
 public class MergeableManifest extends Manifest {
 
 	public static final Attributes.Name BUNDLE_NAME = new Attributes.Name("Bundle-Name");
@@ -117,8 +120,8 @@ public class MergeableManifest extends Manifest {
 		}
 
 		/*
-		 * Copied from base class, but replaced call to make72Safe(buffer) with make512Safe(buffer)
-         * and does not write empty values
+		 * Copied from base class, but replaced call to make72Safe(buffer) with
+		 * make512Safe(buffer) and does not write empty values
 		 */
 		@SuppressWarnings("deprecation")
 		public void myWriteMain(DataOutputStream out) throws IOException {
@@ -159,8 +162,8 @@ public class MergeableManifest extends Manifest {
 		}
 
 		/*
-		 * Copied from base class, but omitted call to make72Safe(buffer)
-		 * and does not write empty values
+		 * Copied from base class, but omitted call to make72Safe(buffer) and
+		 * does not write empty values
 		 */
 		@SuppressWarnings("deprecation")
 		public void myWrite(DataOutputStream out) throws IOException {
@@ -214,7 +217,8 @@ public class MergeableManifest extends Manifest {
 	 * adds the qualified names to the require-bundle attribute, if not already
 	 * present.
 	 *
-	 * @param bundles - passing parameterized bundled (e.g. versions, etc.) is not supported
+	 * @param bundles - passing parameterized bundled (e.g. versions, etc.) is
+	 * not supported
 	 */
 	public void addRequiredBundles(Set<String> bundles) {
 		// TODO manage transitive dependencies
@@ -227,7 +231,8 @@ public class MergeableManifest extends Manifest {
 				bundleName = bundleName.substring(0, idx);
 			}
 		}
-		if (bundleName != null && bundles.contains(bundleName) || projectName != null && bundles.contains(projectName)) {
+		if (bundleName != null && bundles.contains(bundleName)
+				|| projectName != null && bundles.contains(projectName)) {
 			bundlesToMerge = new LinkedHashSet<String>(bundles);
 			bundlesToMerge.remove(bundleName);
 			bundlesToMerge.remove(projectName);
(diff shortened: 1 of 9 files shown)


org.eclipse.xtext.util/src/org/eclipse/xtext/util/Strings.java
```diff
@@ -576,4 +576,17 @@ public static String wordWrap(String string, int maxCharsPerLine) {
 		}
 		return document.toString();
 	}
+
+	/**
+	 * Repeat the stringToRepeat count times.
+	 * 
+	 * @since 2.16
+	 */
+	public static String repeat(String stringToRepeat, int count) {
```
cdietrich: why not use the one from guava?

org.eclipse.xtext.xtext.generator/xtend-gen/org/eclipse/xtext/xtext/generator/XtextGenerator.java
```diff
@@ -482,7 +482,7 @@ protected void generatePluginXmls() {
     }
   }
   
-  private static final Logger LOG = Logger.getLogger(XtextGenerator.class);
+  private final static Logger LOG = Logger.getLogger(XtextGenerator.class);
```
cdietrich: generated with a old xtend

org.eclipse.xtext.util/src/org/eclipse/xtext/util/Strings.java
ArneDeutsch: changed

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
```diff
@@ -0,0 +1,652 @@
...
+					continue;
+				writer.append(make512Safe(new StringBuffer(key + ": " + separateCommas(value)), newline));
+			}
+		}
+	}
+
+	private String separateCommas(String value) {
+		if (!value.contains(","))
+			return value;
+		String result = "";
```
szarnekow: Not necessarily import in this context, but in general: result should be of type StringBuilder since we append in a loop.

org.eclipse.xtext.xtext.generator/xtend-gen/org/eclipse/xtext/xtext/generator/model/ManifestAccess.java
```diff
@@ -283,7 +283,7 @@ public void initialize(final Injector injector) {
     injector.injectMembers(this);
   }
   
-  private static final Logger LOG = Logger.getLogger(ManifestAccess.class);
+  private final static Logger LOG = Logger.getLogger(ManifestAccess.class);
```
szarnekow: Outdated Xtend generator

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
```diff
@@ -0,0 +1,652 @@
...
+			result.append(" ");
+		result.append(toAppend).append(newline);
+		return toAppend;
+	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ((entries == null) ? 0 : entries.hashCode());
```
szarnekow: Why don't we consider the other field values in hashcode or equals?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
```diff
@@ -0,0 +1,652 @@
...
+		this.name = toCopy.name;
+		this.version = toCopy.version;
+		this.newline = toCopy.newline;
+		this.mainAttributes.putAll(toCopy.mainAttributes);
+		this.entries.putAll(toCopy.entries);
+		this.modified = false;
+	}
+
+	private void read(InputStream stream) throws IOException {
+		read(new BufferedReader(new InputStreamReader(stream)).lines().collect(Collectors.toList()));
```
szarnekow: Please document if the stream will be closed or not after an invocation of read(..)

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest.java
```diff
@@ -299,7 +304,7 @@ public boolean isModified() {
 	public void write(OutputStream out) throws IOException {
 		DataOutputStream dos = new DataOutputStream(out);
 		// Write out the main attributes for the manifest
-		((OrderAwareAttributes) getMainAttributes()).myWriteMain(dos);
+		getMainAttributes().myWriteMain(dos);
```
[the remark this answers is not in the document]
ArneDeutsch: I do understand the comment. The method "myWriteMain" is not new. Even if I agree the name is questionable.

org.eclipse.xtext.xtext.generator/xtend-gen/org/eclipse/xtext/xtext/generator/model/ManifestAccess.java
ArneDeutsch: Will regenerate with latest Xtend

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
ArneDeutsch: The old code hasn't done as well. Not sure if equals/hashCode is really needed for our code. I just wanted to change the semantic as minimal as possible. Of course two manifests with the same content but one dirty the other not are not considered equal this way ... what do you think? I would stay with current code but if you think otherwise I can easaly change. Removing the two methods or adding the remaining fields would be the options.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
ArneDeutsch: Will enhance documentation.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest.java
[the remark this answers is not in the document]
szarnekow: I see - sorry, I confused the old impl with the clean-room new impl.

Issue #918, opened 2018-10-19
Exception when executing MWE2 on Homeautomation Example in MergeableManifest2


ArneDeutsch:
The replacement of MergeableManifest developed in #864 does not yet work flawless. Creating the Homeautomation Example and executing the MWE2 workflow produces an exception.

> [ERROR]: GeneratorException: (Element: -UNKNOWN-; Reported by: XtextGenerator)
> 	 java.io.IOException: Line is to long 'Export-Package: org.eclipse.xtext.example.homeautomation.ruleEngine.util,org.eclipse.xtext.example.homeautomation.ruleEngine,org.eclipse.xtext.example.homeautomation.ruleEngine.impl,org.eclipse.xtext.example.homeautomation,org.eclipse.xtext.example.homeautomation.parser.antlr.lexer,org.eclipse.xtext.example.homeautomation.parser.antlr.internal,org.eclipse.xtext.example.homeautomation.validation,org.eclipse.xtext.example.homeautomation.jvmmodel,org.eclipse.xtext.example.homeautomation.scoping,org.eclipse.xtext.example.homeautomation.serializer,org.eclipse.xtext.example.homeautomation.parser.antlr,org.eclipse.xtext.example.homeautomation.services,org.eclipse.xtext.example.homeautomation.formatting2'
> 	at org.eclipse.xtext.util.MergeableManifest2.readHeader(MergeableManifest2.java:116)
> 	at org.eclipse.xtext.util.MergeableManifest2.read(MergeableManifest2.java:100)
> 	at org.eclipse.xtext.util.MergeableManifest2.read(MergeableManifest2.java:96)
> 	at org.eclipse.xtext.util.MergeableManifest2.<init>(MergeableManifest2.java:66)
> 	at org.eclipse.xtext.xtext.generator.XtextGenerator.mergeManifest(XtextGenerator.java:365)
> 	at org.eclipse.xtext.xtext.generator.XtextGenerator.generateManifests(XtextGenerator.java:340)
> 	at org.eclipse.xtext.xtext.generator.XtextGenerator.invokeInternal(XtextGenerator.java:245)
> 	at org.eclipse.emf.mwe.core.lib.AbstractWorkflowComponent.invoke(AbstractWorkflowComponent.java:132)
> 	at org.eclipse.emf.mwe.core.lib.Mwe2Bridge.invoke(Mwe2Bridge.java:34)
[...]

PR #919, opened 2018-10-19
#918: Fix MergeableManifest2 and add missing unit tests

ArneDeutsch:


org.eclipse.xtext.tests/src/org/eclipse/xtext/util/MergeableManifest2Test.java
```diff
@@ -971,6 +978,50 @@ public void hachCode_differ_becauseOfEntry() throws Exception {
 		assertNotEquals(manifest0.hashCode(), manifest1.hashCode());
 	}
 
+	@Test
+	public void readWrite_realisticManifest() throws Exception {
+		// @formatter:off
+		String content =
```
szarnekow: Should we try to read all Manifest.MF resources from the classpath and check if they yield errors?

cdietrich: would be at least a good manual test

PR #932, opened 2018-11-06
Performance Fix Cyclic Link Detection

### LINKED ISSUE #844: Integrate Performance optimization for cyclic linking detection
the cyclic linking detection for `org.eclipse.xtext.linking.lazy.LazyLinkingResource.getEObject(String, Triple<EObject, EReference, INode>)` is not optimal due hashset add/remove. the optimization we have in place in some projects should be generalizable so that we should include it in the default implementation

cdietrich:
Fixes #844

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
@@ -19,6 +19,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import javax.inject.Named;
+
 import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.TreeIterator;
 import org.eclipse.emf.common.util.URI;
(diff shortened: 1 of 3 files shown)


cdietrich:
@szarnekow should we go this way? use a different approach? leave it as is

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
```diff
@@ -95,6 +103,12 @@ private CyclicLinkingException(Object detailMessage, Triple<EObject,EReference,I
 
 	private boolean eagerLinking = false;
 
+	@Named(CYCLIC_LINKING_DECTECTION_COUNTER_LIMIT)
+	@Inject(optional=true)
+	private int cyclicLinkingDectectionCounterLimit = 100;
```
szarnekow: I'd make these protected sind `resolving` is protected, too.

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
```diff
@@ -287,7 +305,10 @@ protected EObject getEObject(String uriFragment, Triple<EObject, EReference, INo
 			createAndAddDiagnostic(triple, ex);
 			return null;
 		} finally {
-			resolving.remove(triple);
+			if (cyclicLinkingDetectionCounter > cyclicLinkingDectectionCounterLimit) {
+                resolving.remove(triple);
+            }
+            cyclicLinkingDetectionCounter--;
```
szarnekow: Indentation looks weird. Mixed tabs / spaces? Maybe reformat that method?

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
cdietrich: done

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
cdietrich: fixed

PR #934, opened 2018-11-06
[eclipse/xtext#1182] fixed tests with java 11 agains java 8 target

cdietrich:
[eclipse/xtext#1182] fixed tests with java 11 agains java 8 target

### DIFF
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.mavenTycho/org.xtext.example.mavenTycho.parent/org.xtext.example.mavenTycho.tests/pom.xml
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.mavenTycho/org.xtext.example.mavenTycho.parent/org.xtext.example.mavenTycho.tests/pom.xml
@@ -15,6 +15,25 @@
 				<groupId>org.eclipse.xtend</groupId>
 				<artifactId>xtend-maven-plugin</artifactId>
 			</plugin>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>target-platform-configuration</artifactId>
+				<configuration>
+					<dependency-resolution>
+						<extraRequirements>
+							<!-- to get the org.eclipse.osgi.compatibility.state plugin
+							if the target platform is Luna or later.
+							(backward compatible with kepler and previous versions)
+							see https://bugs.eclipse.org/bugs/show_bug.cgi?id=492149 -->
+							<requirement>
+								<type>eclipse-feature</type>
+								<id>org.eclipse.rcp</id>
+								<versionRange>0.0.0</versionRange>
+							</requirement>
+						</extraRequirements>
+					</dependency-resolution>
+				</configuration>
+			</plugin>
 		</plugins>
 	</build>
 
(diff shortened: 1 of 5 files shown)


cdietrich:
fixes
```
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.compare [22]
  Unresolved requirement: Require-Bundle: org.eclipse.ui.ide; bundle-version="[3.3.0,4.0.0)"
    -> Bundle-SymbolicName: org.eclipse.ui.ide; bundle-version="3.14.200.v20181003-0856"; singleton:="true"
       org.eclipse.ui.ide [26]
         Unresolved requirement: Require-Bundle: org.eclipse.ui; bundle-version="[3.106.0,4.0.0)"
           -> Bundle-SymbolicName: org.eclipse.ui; bundle-version="3.110.0.v20180828-1350"; singleton:="true"
              org.eclipse.ui [23]
                Unresolved requirement: Require-Bundle: org.eclipse.ui.workbench; bundle-version="[3.112.0,4.0.0)"; visibility:="reexport"
                  -> Bundle-SymbolicName: org.eclipse.ui.workbench; bundle-version="3.112.100.v20181005-0915"; singleton:="true"
                     org.eclipse.ui.workbench [106]
                       Unresolved requirement: Import-Package: javax.annotation```

LorenzoBettini:
On a side note, the comments in the POMs are misleading, from what I understand.
I added those comments when I added those sections in the ui.tests (for which, from what I understand, are still required). In headless tests the feature is required for the missing javax.annotation, isn't it?
@cdietrich @kthoms

LorenzoBettini:
I understand it is a temporary workaround, but since that's in the generated POMs (which will not be regenerated by clients when a new version of Xtext is released), the misleading comments will give no clues to the clients and they probably won't remove the workaround.

cdietrich:
feel free to provide a pr for master replacing the comment with something better

LorenzoBettini:
OK, but do you happen to know whether Tycho 1.4.0 (which fixes this problem) is released before Xtext 2.18.0?

cdietrich:
i dont know. i would expect to see the new tycho soon after as platform has published maven artifacts for 2019-03 but i dont see a release planned

cdietrich:
will ask

LorenzoBettini:
OK, if you have news about that please let me know and in case I can provide a PR

cdietrich:
from @mickaelistria on tycho-user
>The best we can say is "soon".
>I'll try to stage it later today (no guarantee) and share the staging URL
for testing. Then, if there is no objection, we'll try release it next week.
>I'll be taking care of this process for the 1st time ever as Jan is
currently not available to do it. So expect some glitches and be forgiving
>Some more news either within a few hours in the best case, or on Monday in
the hopefully worst case.

LorenzoBettini:
OK, so let's just wait for it to be released, there's no need to change the current code base :)

mickaelistria:
Unless you're planning a release of XText very soon, you can also configure the build to use Tycho 1.4.0-SNAPSHOT.

LorenzoBettini:
@cdietrich would it make sense to fix the comments for 2.17.1?
when is it scheduled?

cdietrich:
could you provide a pr?
will discuss with the team?

cdietrich:
we plan to release it today or tomorrow

cdietrich:
https://github.com/eclipse/xtext-core/tree/2.17_maintenance

LorenzoBettini:
mh... I'm unable to do that today I'm afraid... :(
maybe tomorrow, but only in the afternoon... I guess that's too late, isn't it?

cdietrich:
then i will do.

LorenzoBettini:
Thank you!

Issue #935, opened 2018-11-06
[Wizard] Consider workaround for: Surefire manifest jar classloading broken on latest Debian/Ubuntu Java8


cdietrich:
https://issues.apache.org/jira/browse/SUREFIRE-1588

```<plugin>
					<groupId>org.apache.maven.plugins</groupId>
					<artifactId>maven-surefire-plugin</artifactId>
					<version>2.22.1</version>
					<configuration>
          				<useSystemClassLoader>false</useSystemClassLoader>
        			</configuration>
					<dependencies>
						<dependency>
							<groupId>org.junit.platform</groupId>
							<artifactId>junit-platform-surefire-provider</artifactId>
							<version>1.2.0</version>
						</dependency>
						<dependency>
							<groupId>org.junit.jupiter</groupId>
							<artifactId>junit-jupiter-engine</artifactId>
							<version>5.1.0</version>
						</dependency>
					</dependencies>
				</plugin>
```

cdietrich:
Todo remove workaround when new surefire plugin is our

cdietrich:
maven-surefire-plugin 3.0 Final is not yet out

cdietrich:
move to 2.22

cdietrich:
surefire 3.0.0 seems to be out.

cdietrich:
of course none of the current ubuntus i have still have the problem.

LorenzoBettini:
I think that was fixed a few years ago...

cdietrich:
in ubuntu?

LorenzoBettini:
Yes, I switched to Arch a few months ago as my daily driver, but last year I was still using Ubuntu daily.

cdietrich:
replaced by https://github.com/eclipse/xtext/issues/2200

PR #936, opened 2018-11-06
[#935] workaround for surefirebug on latest ubuntu/debian/mint

cdietrich:
[#935] workaround for surefirebug on latest ubuntu/debian/mint


cdietrich:
that you cannot do that.  => we will stick with spcifiying plugin versions in wizard

LorenzoBettini:
We should consider providing a parent pom as well, that would make everything more transparent and maintainable... at least that's what parent poms are for ;)

cdietrich:
am not sure how this fits with other parent poms like sonatype oss or company specific parent poms

LorenzoBettini:
I'm not sure I understand your concern... if you refer to sonatype oss parent for deploying, IIRC, that's deprecated (I guess for allowing project/framework specific parent poms)... I was thinking of a parent pom like other projects do, e.g., spring boot, where main plugins are preconfigured.

For the moment, I guess that having a bom is already a big step forward

PR #949, opened 2018-11-09
[#950] Fixed regression in Strings.convertFromJavaString

szarnekow:
This fixes a regression introduced with Xtext 2.15. The entire logic for String escaping does perform better now and is easier to reuse. This is achieved by introducing the JavaStringConverter.

The commit does also contain a bunch of fixed warnings.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/JavaStringConverter.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/JavaStringConverter.java
@@ -21,15 +21,44 @@ public class JavaStringConverter {
 	 * Optionally handle unicode escape sequences, too. 
 	 */
 	public String convertFromJavaString(String string, boolean useUnicode) {
+		int firstEscapeSequence = string.indexOf('\\');
+		if (firstEscapeSequence == -1) {
+			return string;
+		}
 		int length = string.length();
 		StringBuilder result = new StringBuilder(length);
-		return convertFromJavaString(string, useUnicode, 0, result);
+		appendRegion(string, 0, firstEscapeSequence, result);
+		return convertFromJavaString(string, useUnicode, firstEscapeSequence, result);
+	}
+
+	protected void appendRegion(String string, int fromInclusive, int toExclusive, StringBuilder result) {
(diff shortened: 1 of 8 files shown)


szarnekow:
> Could you describe the regression? Maybe even in a separate issue and refer to that issue?

This test fails with 2.15 but runs successfully with 2.14:
https://github.com/eclipse/xtext-core/pull/949/files#diff-555f9c90389e4bd1912d8445e05af728R246

org.eclipse.xtext.util/src/org/eclipse/xtext/util/JavaStringConverter.java
```diff
@@ -0,0 +1,274 @@
...
+	protected int unescapeCharAndAppendTo(String string, boolean useUnicode, int index, StringBuilder result) {
+		char c = string.charAt(index++);
+		if (c == '\\') {
+			return doUnescapeCharAndAppendTo(string, useUnicode, index, result);
+		}
+		validateAndAppendChar(result, c);
+		return index;
+	}
+	
+	protected void validateAndAppendChar(StringBuilder result, char c) {
```
cdietrich: invert param order?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/JavaStringConverter.java
```diff
@@ -0,0 +1,274 @@
...
+		if (validate(result, c)) {
+			result.append(c);	
+		}
+	}
+	
+	protected boolean validate(StringBuilder result, char c) {
```
cdietrich: invert param order?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/JavaStringConverter.java
szarnekow: Done

org.eclipse.xtext.util/src/org/eclipse/xtext/util/JavaStringConverter.java
szarnekow: Done

Issue #953, opened 2018-11-12
Regression in MergableManifest2


szarnekow:
This test fails with the rewritten manifest impl:
```
@Test
    public void readWrite_realisticManifest_regression() throws Exception {
        // @formatter:off
        String content =
            "Manifest-Version: 1.0" + NL + 
            "Bundle-ManifestVersion: 2" + NL + 
            "Bundle-Name: Xtext Homeautomation Example - Runtime" + NL + 
            "Bundle-Vendor: Eclipse Xtext" + NL + 
            "Bundle-Version: 2.16.0.qualifier" + NL + 
            "Bundle-SymbolicName: org.eclipse.xtext.example.homeautomation; singleton:=true" + NL + 
            "Bundle-ActivationPolicy: lazy" + NL + 
            "Require-Bundle: org.eclipse.xtext," + NL + 
            " org.eclipse.xtext.xbase," + NL + 
            " org.eclipse.equinox.common;bundle-version=\"3.5.0\"," + NL + 
            " org.eclipse.emf.ecore," + NL + 
            " org.objectweb.asm;bundle-version=\"[6.2.1,6.3.0)\";resolution:=optional," + NL + 
            " org.eclipse.xtext.util," + NL + 
            " org.antlr.runtime;bundle-version=\"[3.2.0,3.2.1)\"," + NL + 
            " org.eclipse.xtend.lib;bundle-version=\"2.14.0\"," + NL + 
            " org.eclipse.emf.common," + NL + 
            " org.eclipse.xtext.common.types;bundle-version=\"2.14.0\";visibility:=reexport," + NL + 
            " org.eclipse.xtext.xbase.lib;bundle-version=\"2.14.0\"" + NL + 
            "Bundle-RequiredExecutionEnvironment: JavaSE-1.8" + NL + 
            "Export-Package: org.eclipse.xtext.example.homeautomation.ruleEngine.util," + NL + 
            " org.eclipse.xtext.example.homeautomation.ruleEngine," + NL + 
            " org.eclipse.xtext.example.homeautomation.ruleEngine.impl," + NL + 
            " org.eclipse.xtext.example.homeautomation," + NL + 
            " org.eclipse.xtext.example.homeautomation.parser.antlr.lexer," + NL + 
            " org.eclipse.xtext.example.homeautomation.parser.antlr.internal," + NL + 
            " org.eclipse.xtext.example.homeautomation.validation," + NL + 
[...]

PR #956, opened 2018-11-13
#953: Support setting of entries with attributes.

ArneDeutsch:
- Added test cases to cover setting of entries that have other
attributes then version alone.
- Fix MergeableManifest to support given test cases.
- Refactor existing test cases a bit to avoid uneccessary
"Collections.singleton"

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/util/MergeableManifest2Test.java
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/util/MergeableManifest2Test.java
@@ -17,7 +17,6 @@ import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
-import java.util.HashSet;
 import java.util.List;
 
 import org.junit.Test;
@@ -539,7 +538,7 @@ public class MergeableManifest2Test {
 	@Test
 	public void addRequiredBundles_missingAttribute_oneEmptyValue() throws Exception {
 		MergeableManifest2 manifest = newManifest("Manifest-Version: 1.0" + NL);
-		manifest.addRequiredBundles(Collections.singleton(""));
+		manifest.addRequiredBundles("");
 
 		assertFalse(manifest.isModified());
 		assertEquals("Manifest-Version: 1.0" + NL, write(manifest));
@@ -548,7 +547,7 @@ public class MergeableManifest2Test {
 	@Test
 	public void addRequiredBundles_missingAttribute_twoValues() throws Exception {
 		MergeableManifest2 manifest = newManifest("Manifest-Version: 1.0" + NL);
-		manifest.addRequiredBundles(new HashSet<>(java.util.Arrays.asList("a", "b")));
+		manifest.addRequiredBundles("a", "b");
 
 		assertTrue(manifest.isModified());
 		assertEquals("Manifest-Version: 1.0" + NL + "Require-Bundle: a," + NL + " b" + NL, write(manifest));
@@ -557,7 +556,7 @@ public class MergeableManifest2Test {
 	@Test
 	public void addRequiredBundles_oneValueBefore_addingOneValue() throws Exception {
 		MergeableManifest2 manifest = newManifest("Manifest-Version: 1.0" + NL + "Require-Bundle: a" + NL);
-		manifest.addRequiredBundles(Collections.singleton("b"));
+		manifest.addRequiredBundles("b");
 
 		assertTrue(manifest.isModified());
 		assertEquals("Manifest-Version: 1.0" + NL + "Require-Bundle: a," + NL + " b" + NL, write(manifest));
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.tests/src/org/eclipse/xtext/util/MergeableManifest2Test.java
```diff
@@ -1028,6 +1104,50 @@ public void readWrite_realisticManifest() throws Exception {
...
+            "Require-Bundle: org.eclipse.xtext," + NL + 
+            " org.eclipse.xtext.xbase," + NL + 
+            " org.eclipse.equinox.common;bundle-version=\"3.5.0\"," + NL + 
+            " org.eclipse.emf.ecore," + NL + 
+            " org.objectweb.asm;bundle-version=\"[6.2.1,6.3.0)\";resolution:=optional," + NL + 
+            " org.eclipse.xtext.util," + NL + 
+            " org.antlr.runtime;bundle-version=\"[3.2.0,3.2.1)\"," + NL + 
+            " org.eclipse.xtend.lib;bundle-version=\"2.14.0\"," + NL + 
+            " org.eclipse.emf.common," + NL + 
+            " org.eclipse.xtext.common.types;bundle-version=\"2.14.0\";visibility:=reexport," + NL +
```
szarnekow: Please add a test where bundle-version and visibility are present in a different order

ArneDeutsch: Added such a test.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
```diff
@@ -453,39 +490,47 @@ private void writeEntries(BufferedWriter writer) throws IOException {
...
 		String seperator = "";
+		for (String part : splitAtCharHonorQuoting(value, ',')) {
+			result += seperator + part;
+			seperator = "," + newline + " ";
+		}
+		return result;
+	}
+
+	private static List<String> splitAtCharHonorQuoting(String value, char c) {
+		if (!value.contains("" + c))
```
szarnekow: IndexOf

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
```diff
@@ -316,25 +318,36 @@ private static String mergeIntoCommaSeparatedList(String bundle, String oldBundl
 		if (oldBundles.isEmpty())
 			return bundle;
 		String result = "";
-		String[] oldBundleList = oldBundles.split(",");
-		String newBundleName = bundleName(bundle);
-		String newBundleVersion = bundleVersion(bundle);
 		String seperator = "";
```
szarnekow: Typo: separator

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
```diff
@@ -316,25 +318,36 @@ private static String mergeIntoCommaSeparatedList(String bundle, String oldBundl
 		if (oldBundles.isEmpty())
 			return bundle;
 		String result = "";
```
szarnekow: Even though it is not relevant here, but generally speaking a StringBuilder would be the better choice to create the result.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
```diff
@@ -343,11 +356,35 @@ private static String mergeIntoCommaSeparatedList(String bundle, String oldBundl
 	}
 
 	private static String bundleName(String bundle) {
-		return bundle.contains(";") ? bundle.split(";")[0] : bundle;
+		if (!bundle.contains(";"))
```
szarnekow: We do split this multiple times. Again: Not relevant for performance, but a result object that encapsulates the bunde name, suffix and version would be a cleaner solution. It could also expose a merge method and a reasonable toString()

```
static class RequiredBundle {
  String name, version, suffix // I don't like the name suffix
  RequiredBundle merge(String entry} { .. }
  String toManifest() {
     return ...
  }
}

```

cdietrich:
@ArneDeutsch @szarnekow what is the state of this pr?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
ArneDeutsch: Will change.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
ArneDeutsch: Will fix.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
ArneDeutsch: Will change.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
ArneDeutsch: Will extract classes "Bundle" and "BundleList".

ArneDeutsch:
All change requests handled.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
```diff
@@ -671,4 +645,130 @@ public boolean equals(Object obj) {
...
+						return part.substring(startIndex);
+					}
+				}
+			}
+			return null;
+		}
+
+		// Output is used by algorithm ... do not change for debugging purposes!
+		@Override
+		public String toString() {
```
szarnekow: Maybe add a dedicated method and let toString call that one and make the algo based on that method?

PR #957, opened 2018-11-13
[lsp] use last segment in document symbol name only

JanKoehnlein:
In an hierarchical outline, it doesn't make sense to show qualified names of all nodes by default.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolMapper.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolMapper.xtend
@@ -62,7 +62,7 @@ class DocumentSymbolMapper {
 	 * Provides {@link DocumentSymbol#detail detail} for a {@link DocumentSymbol document symbol}.
 	 * <p>
 	 * Always returns with an empty string by default.
-	 * 
+	 *
 	 * @see DocumentSymbol#detail
 	 */
 	@Beta
@@ -77,7 +77,7 @@ class DocumentSymbolMapper {
 
 	/**
 	 * Provides a human-readable name for the document symbol.
-	 * 
+	 *
 	 * @see DocumentSymbol#name
 	 */
 	@Beta
@@ -96,14 +96,14 @@ class DocumentSymbolMapper {
 		}
 
 		protected def String getName(QualifiedName qualifiedName) {
-			return qualifiedName?.toString;
+			return qualifiedName?.lastSegment;
 		}
 
 	}
 
 	/**
 	 * Provides the {@link SymbolKind symbol kind} information for the document symbol.
-	 * 
+	 *
 	 * @see DocumentSymbol#kind
 	 */
 	@Beta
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolMapper.xtend
```diff
@@ -96,14 +96,14 @@ class DocumentSymbolMapper {
 		}
 
 		protected def String getName(QualifiedName qualifiedName) {
-			return qualifiedName?.toString;
```
cdietrich: should we then use org.eclipse.xtext.ide.server.symbol.DocumentSymbolMapper.DocumentSymbolDetailsProvider.getDetails(EObject) 
to give the qualified name by default

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolMapper.xtend
```diff
@@ -96,14 +96,14 @@ class DocumentSymbolMapper {
...
+			return qualifiedName?.lastSegment;
```
kittaakos: Since the non-hierarchical document symbol service [delegates into the hierarchical one (and flattens the symbols after visiting them in pre-order traversal)](https://github.com/eclipse/xtext-core/blob/e3464552ade1f663065f1ae5bc8ef2040566b1a0/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.xtend#L185) we will show only the last segments for the non-hierarchical case too.

Let me know if this needs more clarification.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolMapper.xtend
```diff
@@ -62,7 +62,7 @@ class DocumentSymbolMapper {
 	 * Provides {@link DocumentSymbol#detail detail} for a {@link DocumentSymbol document symbol}.
 	 * <p>
 	 * Always returns with an empty string by default.
-	 *
```
kittaakos: [nitpicky-remark]: That extra whitespace was there because that is the default `Xtend` formatting output in Eclipse. Can we switch back to the default one?

screencast 2018-11-13 14-49-43

cdietrich: :+1:

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolMapper.xtend
cdietrich: per se it would be ok for me

JanKoehnlein:
Sorry, I thought it would only be used in the hierarchical service. For non-hierachical sumbols the FQN makes sense, so we better leave this as is.

PR #962, opened 2018-11-14
[ls] expose a bit more API to the subclasses of ServerLauncher

JanKoehnlein:


org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/server/ServerLauncher.java
```diff
@@ -71,6 +78,13 @@ public void start(final LaunchArgs it) {
     }
   }
   
+  /**
+   * @since 2.16
+   */
+  protected Launcher<LanguageClient> createLauncher(final LaunchArgs it) {
+    return Launcher.<LanguageClient>createLauncher(this.languageServer, LanguageClient.class, it.getIn(), it.getOut(), it.isValidate(), it.getTrace());
```
szarnekow: this.languageServer should probably use the getter? Otherwise it would be pointless to make it possible to override it.

JanKoehnlein:
Found my way to work around this. Sorry for the noise.

PR #965, opened 2018-11-20
Set reference kind when we know it should be set

tivervac:
The reference kind can already be set when the proposal is created at this place because we know the proposal is created for a CrossReference.

This avoids having to do more complex checking later on in the auto complete flow to set the kind.


cdietrich:
wonder if org.eclipse.xtext.xbase.ide.contentassist.XbaseIdeCrossrefProposalProvider needs to be changed as well

cdietrich:
this pr can be merged. can you please open a second one against extras?

tivervac:
Will do

PR #966, opened 2018-11-20
Use try read only

tivervac:
The suggested type for `readOnly()` is a `CancelableUnitOfWork` and several places ([here](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext/src/org/eclipse/xtext/resource/OutdatedStateManager.xtend#L63) and [here](https://github.com/eclipse/xtext-eclipse/blob/master/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocument.java#L506)) check if the `IUnitOfWork` is a `CancelableUnitOfWork`. It isn't because `tryReadOnly()` wraps the `IUnitOfWork` in a lambda.

This patch modifies `tryReadOnly()`, `tryPriorityReadOnly()` and `tryModify()` so that they wraps `work` in a `CancelableUnitOfWork` if work is a `CancelableUnitOfWork`.

A prettier way would be to add an overloaded `try...` that takes a `CancelableUnitOfWork` as argument instead of a `IUnitOfWork`, but that would require us to overload 9 methods.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/IReadAccess.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/IReadAccess.java
@@ -21,6 +21,7 @@ public interface IReadAccess<State> {
 	 * Gets a read-only copy of the State and executes {@code work} on it.
 	 * 
 	 * WARNING: the State passed to {@code work} can be null.
+	 * e.g. when reading from a read-only zip/jar entry
 	 * 
 	 * @param work Work to execute on the State
 	 * 
@@ -43,6 +44,11 @@ public interface IReadAccess<State> {
 		IUnitOfWork<Result, State> work,
 		Supplier<? extends Result> defaultResult
 	) {
+		// Some implementations rely on the type of {@code work}
+		if (work instanceof CancelableUnitOfWork<?, ?>) {
+			return readOnly(new WrappingCancelableUnitOfWork<>(defaultResult, work));
+		}
+
 		return readOnly((state) -> {
 			if (state == null) {
 				return defaultResult.get();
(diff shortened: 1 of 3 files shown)


cdietrich:
@tivervac is this needed for 2.16, is it ok to have this in 2.17 only?

tivervac:
@cdietrich Needed is a big word, but the old behaviour for tryReadOnly is not the expected behaviour. It also prevents migration from readOnly to tryReadOnly (same for tryPriorityReadOnly and tryModify).

cdietrich:
i am not sure if i want to get this in with the few days we have left for testing and the open problems we have

cdietrich:
i am not sure how this big the risk of this change is cause i was not involved in the previous changes.

cdietrich:
@szarnekow what do you think?

cdietrich:
if this PR goes into 2.16 then the eclipse one should go too?

tivervac:
@cdietrich The eclipse one can be delayed without a problem

org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/IReadAccess.java
```diff
@@ -126,6 +144,25 @@
 			IUnitOfWork<Result, State> work,
 			Supplier<? extends Result> defaultResult
 		) {
+			// Some implementations rely on the type of {@code work}
+			if (work instanceof CancelableUnitOfWork<?, ?>) {
+				return priorityReadOnly(new CancelableUnitOfWork<Result, State>() {
```
szarnekow: Can we extract the anonymous class into a named one since the code is the same for cases?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/concurrent/IWriteAccess.java
```diff
@@ -42,6 +44,25 @@
 		IUnitOfWork<Result, State> work,
 		Supplier<? extends Result> defaultResult
 	) {
+		// Some implementations rely on the type of {@code work}
+		if (work instanceof CancelableUnitOfWork<?, ?>) {
```
szarnekow: It could be reused here, too.

tivervac:
@szarnekow Done

tivervac:
Rebased and squashed, waiting for ci to reconfirm

cdietrich:
@tivervac build is still green

Issue #56, opened 2016-07-26
support @InjectWith on @Test methods


meysholdt:
In Xtext we have plenty of tests and plenty of test languages. 
Quite often a test language contains all use cases that a certain test class needs. This led to many quite large test languages, which is bad due to the heavy code footprint of each language.

Instead, it would be nicer if a test class could reuse languages from a larger selection of smaller test languages. For this, however, its necessary to choose a language per `@Test` method.

Our old infrastructure (AbstractXtextTest) did support this, too. 
See `with(SomeLang.class)` in [PartialParserTest](https://github.com/eclipse/xtext-core/blob/e1a4782403497dd27f73276c409049a3d3ecdcf9/org.eclipse.xtext.tests/src/org/eclipse/xtext/parser/PartialParserTest.java)

Conceptually it is no problem to use a language per method because the JUnit Runner by default creates a new instance of the test class for each execution of a method.

franzbecker:
I had a brief look into it and it seems like it's quite hard to accomplish using JUnit 4.12.

The reason is that in order to perform the injection we need the actual instance of the test. It is created in `XtextRunner.createTest()` which is missing the parameter which method it is created for. There is no other way to get hold of the instance and overwriting the entire `BlockJUnit4ClassRunner.methodBlock(FrameworkMethod)` would require a lot of copy&pasting as within it the private method `withRules(...)` is called.

JUnit 4.13 (whenever it might be released) will provide a hook for it (https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java):
```
    /**
     * Returns a new fixture to run a particular test {@code method} against.
     * Default implementation executes the no-argument {@link #createTest()} method.
     *
     * @since 4.13
     */
    protected Object createTest(FrameworkMethod method) throws Exception {
        return createTest();
    }
```

For now there might be hacky solutions that store either the created test instance or the current method in a thread-local variable ☠️. A better possibility would be to create an additional JUnit MethodRule (see this [discussion](https://github.com/junit-team/junit4/issues/793)) - which might also implement the TestRule interface. Personally I would prefer having a rule instead of a custom runner as this approach is more flexible.

szarnekow:
There are no plans to enhance the JUnit4 based test APIs.

PR #55, opened 2016-07-26
[LanguageServer] Fix HoverService

cdietrich:
Fixes the Hover Service inside the Language Server Impl to support Hovers for Cross References that are not self contained as well.

PR #969, opened 2018-11-21
[trace] Improved implementation for GeneratorNodeProcessor.Context

dhuebner:
Time spend for the performance test  

```
Lines 800_000 // Before x sec - after 4.6 sec
Lines  80_000 // Before 89 sec - after 0.43 sec
Lines  40_000 // Before 20 sec - after 0.32 sec
Lines  20_000 // Before 5 sec - after 0.25 sec
Lines  10_000 // Before 1.8 sec - after 0.25 sec
```

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/generator/trace/node/GeneratorNodePerformanceTest.xtend
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/generator/trace/node/GeneratorNodePerformanceTest.xtend
@@ -1,0 +1,47 @@
+/*******************************************************************************
+ * Copyright (c) 2017 TypeFox (https://typefox.io) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.generator.trace.node
+
+import java.util.concurrent.TimeUnit
+import org.eclipse.xtext.generator.trace.LocationData
+import org.eclipse.xtext.generator.trace.SourceRelativeURI
+import org.junit.Assert
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.Timeout
+
+/**
+ * @author Dennis Huebner - Initial contribution and API
+ */
+class GeneratorNodePerformanceTest {
+
+	extension GeneratorNodeExtensions exts = new GeneratorNodeExtensions()
+
+//	static val numberOfLines = 800_000 // Before x sec - after 4.6 sec
+//	static val numberOfLines = 80_000 // Before 89 sec - after 0.43 sec
+	static val numberOfLines = 40_000 // Before 20 sec - after 0.32 sec
+//	static val numberOfLines = 20_000 // Before 5 sec - after 0.25 sec
+//	static val numberOfLines = 10_000 // Before 1.8 sec - after 0.25 sec
+	@Rule
+	public val Timeout timeout = new Timeout(1000, TimeUnit.MILLISECONDS)
+
+	@Test def void testBasicCreationAndProcessingPerformance() {
+		val root = loc(0)
+		var node = root.trace.append('notindented').appendNewLine
+		for (i : 0 ..< numberOfLines) {
+			node.indent.trace(loc(1)).append('indented1').appendNewLine.indent.append('indented2').appendNewLine
+		}
+		node.appendNewLine.append('dedented')
+		val processor = new GeneratorNodeProcessor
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.tests/src/org/eclipse/xtext/generator/trace/node/GeneratorNodePerformanceTest.xtend
```diff
@@ -0,0 +1,47 @@
+/*******************************************************************************
+ * Copyright (c) 2017 TypeFox (https://typefox.io) and others.
```
cdietrich: Wrong year

dhuebner: yes

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/GeneratorNodeProcessor.xtend
```diff
@@ -42,38 +42,103 @@ class GeneratorNodeProcessor {
...
+			_currentIndents.push(node)
+			recalculateIdentLength
+		}
+		
+		def void decreaseIndents() {
+			_currentIndents.pop
+			recalculateIdentLength
+		}
+		
+		protected def recalculateIdentLength() {
```
dhuebner: Typo, should be `recalculateIndentLength`

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/GeneratorNodeProcessor.xtend
```diff
@@ -42,38 +42,103 @@ class GeneratorNodeProcessor {
 		}
 	}
 	
-	@Accessors protected static class Context {
```
cdietrich: do we need a getLines to stay api compatible or is this no api? (afaik package is still exported as xfriends/internal

dhuebner: I was thinking about the compatibility, but it seems like the GeneratorNodeProcessor API is not used very often and it's marked as internal.

cdietrich: yes thus the question if if you want to keep the other places of compatibility code

cdietrich:
@dhuebner do you know if we have tests that actually test the resulting code? i can find only TracingSugarTest. is that sufficient?

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/GeneratorNodeProcessor.xtend
dhuebner: This would mean to track the access to `getLines` members otherwise it will break the runtime compatibility. I think it's better to just replace the implementation. It's a protected Class in an internal package with experimental API.

cdietrich: ok

dhuebner:
@cdietrich  All the tests I found are in `org.eclipse.xtext.tests/org.eclipse.xtext.generator.trace.node package`

dhuebner:
> breakpoints in set/isPendingIndent never match

This is strange as it called in `def contentLength`

cdietrich:
Had a second look. Is called in GeneratorNodeTest.

kthoms:
@dhuebner I did not see that this PR was approved. Please do not merge PRs without approval of someone else! Also, we delete branches after merging. There are already many branches open.

https://github.com/eclipse/xtext/blob/master/QUALITY_GUIDELINES.md

dhuebner:
@kthoms The "all checks passed" was misleading. I thought @cdietrich did the review and approved.  Sorry for the inconvenience

kthoms:
That's just the IP & Build Check. All fine, it is a valuable change. Remember just for the next time.

cdietrich:
@dhuebner how much buffer does the test have: http://services.typefox.io/open-source/jenkins/job/xtext-core/job/master/57/

dhuebner:
@cdietrich Compared to my machine ~300%  (0.32 sec expected 1 sec Timeout) We can increase it to 2-3 sec

cdietrich:
lets see how the test behaves load was very heavy some minutes ago

dhuebner:
compare to prev build 57 was a bit slow:

  | #57 | 10 Minuten
-- | -- | --
  | #56 | 6 Minuten 9 Sekunden
  | #55 | 6 Minuten 58 Sekunden

Issue #981, opened 2018-12-17
[build] provide an easy to execute "create project" CLI tool


JanKoehnlein:
It's nice that we can create new Xtext projects from the command line, but setting up everything in order to run the respective Java class is still a bit tedious.

We should create at least some distro zip during the build with the required libs and scripts to assemble the classpath, locate the JRE, and launch the app.

JanKoehnlein:
Why is `args4j` an optional dependency of `org.eclipse.xtext.xtext.wizard`?

cdietrich:
Cause you don’t have the command line class but you can

JanKoehnlein:
The non-CLI usecase is Eclipse, isn't it? Does it really matter to add another jar of 159k then?

cdietrich:
There is no availability in orbit maybe. Or on default platform so that we would have to repackage it to our update site 
But why is it a problem?

JanKoehnlein:
Maybe we should refactor it to use something like apache.cli.

cdietrich:
am not sure if apache cli is even more heavyweight. so the problem is that you cannot tell the gradle application plugin to include optional dependencies. @oehme do you know a way to do so?

kthoms:
> Maybe we should refactor it to use something like apache.cli.

Same there, the dependency (commons-cli then) should be optional.

JanKoehnlein:
...or, the CLI class is just placed into the wrong project. 

I still don't quite get the weight argument. As far as Orbit is concerned, does that really make a difference when a dependency is optional? And do you really search for executables at Maven? How would that make consumption easier?

This was just a quick shot at making an pretty incomplete feature a bit more usable. Likely, still not complete enough. I am open for better solutions.

cdietrich:
well i think it was ment to be used from outside where you can provide the dependencies you need.

kthoms:
The publishing to Maven Central is done by [your publishing plugin](https://github.com/TypeFox/publishing.git). Please ask @spoenemann or @dhuebner what needs to be done to package the application to a zip and deploy it with an additional classifier coordinate (`cli` could be an appropriate classifier) to artifact coordinate [`org.eclipse.xtext:org.eclipse.xtext.xtext.wizard`](https://search.maven.org/search?q=a:org.eclipse.xtext.xtext.wizard).

cdietrich:
dependency is now non optional

szarnekow:
[...]

PR #982, opened 2018-12-17
GH-981 make project creator CLI tool easily consumable

JanKoehnlein:
Build the scripts and bundle the libs in a zip.

### DIFF
--- a/org.eclipse.xtext.xtext.wizard/build.gradle
+++ b/org.eclipse.xtext.xtext.wizard/build.gradle
@@ -3,7 +3,7 @@ description = 'Platform-independent component for creating new Xtext projects.'
 
 dependencies {
 	compile project(':org.eclipse.xtext')
-	optional "args4j:args4j:$versions.args4j"
+	compile "args4j:args4j:$versions.args4j"
 }
 
 sourceSets.main.resources.srcDir 'resources'
@@ -11,3 +11,10 @@ sourceSets.main.resources.srcDir 'resources'
 sourcesJar {
 	exclude '**/gradlew/**'
 }
+
+apply plugin: 'application'
+mainClassName = "org.eclipse.xtext.xtext.wizard.cli.CliProjectsCreatorMain"
+applicationName = 'createXtextProject'
+
+
+ 


org.eclipse.xtext.xtext.wizard/build.gradle
```diff
@@ -3,11 +3,18 @@ description = 'Platform-independent component for creating new Xtext projects.'
 
 dependencies {
 	compile project(':org.eclipse.xtext')
-	optional "args4j:args4j:$versions.args4j"
+	compile "args4j:args4j:$versions.args4j"
```
cdietrich: Am not sure if we want this

kthoms:
This will produce the application, but how would it be distributed?

org.eclipse.xtext.xtext.wizard/build.gradle
JanKoehnlein: It won't build an executable application otherwise.

JanKoehnlein:
Well, for me it'd be enough to download the ZIP from Jenkins. We could also push it to bintray or similar. What do you expect as distribution?

kthoms:
It should go to Maven Central like everything else, likely with packaging zip. I don't think we should introduce another distribution kind. And the process should be automated. From Maven perspective this would be an attached artifact, not sure what to do for Gradle here to make it work. 
The resulting zip should be picked up by the [xtext-snapshots](https://ci.eclipse.org/xtext/job/xtext-snapshots/) build job and be deployed.

kthoms:
> Well, for me it'd be enough to download the ZIP from Jenkins. 
This would be an undefined state and not usable from clients that would like to use a release version from a scripted build (e.g. to create a Yeoman plugin).

kthoms:
Could be something that could be configured for the Gradle [application plugin](https://docs.gradle.org/current/userguide/application_plugin.html), but I can't read it from the docs how to do that.

cdietrich:
dependency is now non optional on master. i dont think we should publish an additional zip to somewhere. i dont even know if the typefox publishing plugin could handle that

Issue #983, opened 2018-12-17
DiagnosticConverterImpl: Make message customizable


kthoms:
In a special use case we want to customize the message for produced issues that differ from the diagnostic's message (to be specific: the diagnostic's data carries an exception and the message should contain the exception type and even stack trace).

The creation of diagnostics from exceptions therefore should be customizable.

kthoms:
CC @fabtagon

cdietrich:
@kthoms anything left

kthoms:
DiagnosticConverterImpl contains now `getMessage()` methods, which can be overridden by a customized class.

PR #984, opened 2018-12-17
[#983] Make issue message customizable

kthoms:
Make creation of exception based diagnostics customizable

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/validation/CompositeEValidator.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/validation/CompositeEValidator.java
@@ -176,7 +176,10 @@ public class CompositeEValidator implements EValidator {
 		return result;
 	}
 
-	private Diagnostic createExceptionDiagnostic(String message, Object source, Throwable t) {
+	/**
+	 * @since 2.17
+	 */
+	protected Diagnostic createExceptionDiagnostic(String message, Object source, Throwable t) {
 		return new BasicDiagnostic(Diagnostic.ERROR, source.toString(), 0, message, new Object[] { t });
 
 	}


org.eclipse.xtext/src/org/eclipse/xtext/validation/DiagnosticConverterImpl.java
```diff
@@ -58,7 +58,7 @@ public void convertResourceDiagnostic(Diagnostic diagnostic, Severity severity,
 		issue.setSeverity(severity);
 		issue.setLineNumber(diagnostic.getLine());
 		issue.setColumn(diagnostic.getColumn());
-		issue.setMessage(diagnostic.getMessage());
```
cdietrich: why for message only and not for all other properties here?

szarnekow: We do have API to customize the diagnostic messages. Why is this necessary here?

kthoms: Could you give me a pointer? Might make this one obsolete. I did not came across when solving our requirement.

kthoms: diagnostic#data carries an Exception instance (actually it is promoting NPEs that are normally swallowed by `org.eclipse.xtext.validation.AbstractDeclarativeValidator.MethodWrapper.handleInvocationTargetException(Throwable, State)`).
The exception's class name and stack trace is put into the issue message (customer wants to see this in the editor).
So instead of "Error in EValidator" the exception is taken from the diagnostic's data.

cdietrich: So you want to bug the user with programming errors the dsl developer does?

kthoms: Me? No, the customer requires that. Yes, they want it that way.

cdietrich: Still don’t know why you need to change both places

kthoms: 1) First throws the exception => gets into the `data` of the diagnostic
2) The exception data object should be translated to the issue's message

cdietrich: Where is the place for 1

kthoms: custom MethodWrapper that does not swallow the NPE!

kthoms: *custom*!

kthoms: not here!

cdietrich: Yes but why can’t then the other thing produce a proper message

kthoms: Because it should _not_ produce a specific message, but just handle the NPE as any other runtime exception. And the converter here is responsible to _convert_ the diagnostic to an issue.

kthoms: are we still talking about a simple getter here?

cdietrich: so should that be the place to be protected?

kthoms: Might be an alternative.

kthoms:
Changed to customizable CompositeEValidator

PR #1000, opened 2019-01-22
[eclipse/xtext#1182] added support for java 11 as target

cdietrich:
[eclipse/xtext#1182] added support for java 11 as target


svenefftinge:
Is the jetty version bump somehow related to the Java 11 support?

cdietrich:
yes the old jetty does not work with java 11

PR #1002, opened 2019-01-25
Improve ResourceSetBasedAllContainersStateProvider.getResourc…

AndersDala:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=427770

Short:
    Implement solution mentioned in above BUG report.

Longer:
    Bug (?) can be triggered by e.g.:

    IResourceServiceProvider sp = IResourceServiceProvider.Registry.INSTANCE.getResourceServiceProvider(fileUri);
    IResourceSetProvider provider = sp.get(IResourceSetProvider.class);
    IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
    // ... file etc
    IProject project = file.getProject();
    // Use the provider to get the resource set for the given project
    // FIXME how to get the existing XtextResourceSet? Below just creates a new rs. :-(
    ResourceSet rs = provider.get(project);
    // ... run build to generate index used by namespace etc
    BuildRequest request = new BuildRequest();
    request.setResourceSet(rs);
    request.setBaseDir(UriUtil.createFolderURI(new File(baseDir)));
    request.setDirtyFiles(allFiles);
    // ... copy paste code ...
    indexState = incrementalBuilder.build(request, languages).getIndexState();

    Hints for doing this correctly/cleaner is highly appreciated!

    After the code above has been triggered doing a Project Clean will
    trigger method from title to be called with a CurrentDescriptions.

Note that this commit goes together with a commit in xtext-eclipse (see
https://github.com/eclipse/xtext-eclipse/pull/947/commits/8c30394d3bb904c2370d81afad9657575b361cd3).

Signed-off-by: Anders Dahlberg <anders.xb.dahlberg@ericsson.com>

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/IResourceDescriptions.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/IResourceDescriptions.java
@@ -13,6 +13,7 @@ import org.eclipse.emf.common.notify.Notifier;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.resource.ResourceSet;
 import org.eclipse.xtext.naming.QualifiedName;
 import org.eclipse.xtext.resource.impl.ResourceDescriptionsProvider;
 
@@ -42,6 +43,16 @@ public interface IResourceDescriptions extends ISelectable {
 	interface IContextAware extends IResourceDescriptions {
 
 		void setContext(Notifier ctx);
+	}
+	
+	/**
+	 * A resource set aware instance {@link IResourceDescriptions}.
+	 * 
+	 * @since 2.17
+	 */
+	interface IResourceSetAware extends IResourceDescriptions {
+
+		ResourceSet getResourceSet();
 	}
 
 	class NullImpl implements IResourceDescriptions {
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/ResourceSetBasedAllContainersStateProvider.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/ResourceSetBasedAllContainersStateProvider.java
@@ -11,8 +11,7 @@ import org.apache.log4j.Logger;
 import org.eclipse.emf.ecore.resource.ResourceSet;
 import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.xtext.resource.IResourceDescriptions;
-import org.eclipse.xtext.resource.impl.EagerResourceSetBasedResourceDescriptions;
-import org.eclipse.xtext.resource.impl.ResourceSetBasedResourceDescriptions;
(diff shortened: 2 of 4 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/resource/IResourceDescriptions.java
```diff
@@ -43,6 +44,16 @@
...
 	}
+	
+	/**
+	 * A resource set aware instance {@link IResourceDescriptions}.
+	 * 
+	 * @since 2.17
+	 */
+	interface IResourceSetAware extends IResourceDescriptions {
+
+		ResourceSet getResourceSet();
```
szarnekow: What do you think about a default method in IResourceDesriptions, e.g `default Optional<ResourceSet> getContextResourceSet() { returns Optional.absent() }` with a clarifying JavaDoc, that describes what the context resource set is and under which circumstances it is present or absent?

szarnekow: For existing impls that do have a getResourceSet, the getContextResourceSet should delegate to that one.

AndersDala: If using default, what is the reason for not doing a default getResourceSet() {  return null; } with similar javadoc? Just trying to understand the difference. :-)

szarnekow: Three things come to my mind:
- `getContextResourceSet` carries a slightly better semantics
- `Optional<ResourceSet>` makes it very explicit that this may be not available. 
- existing implementation may already have a method `getResourceSet` and it could potentially be undesirable to expose that one.

AndersDala: Thanks!
Should I change to using Optional<> et al. or is the current solution of an additional interfance easier to understand?

szarnekow: Let's keep the approach as is. The alternative with the default method is not a substantial improvement. Thank you for proving the patch!

org.eclipse.xtext/src/org/eclipse/xtext/resource/IResourceDescriptions.java
```diff
@@ -43,6 +44,16 @@
 
 		void setContext(Notifier ctx);
 	}
+	
+	/**
+	 * A resource set aware instance {@link IResourceDescriptions}.
+	 * 
+	 * @since 2.17
+	 */
+	interface IResourceSetAware extends IResourceDescriptions {
```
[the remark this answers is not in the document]
szarnekow: 👍 I think they deserve to co-exist.

cdietrich:
thx @AndersDala

org.eclipse.xtext/src/org/eclipse/xtext/resource/IResourceDescriptions.java
AndersDala: Np, thank you!

PR #1006, opened 2019-01-29
[#1009] Remove managed dependencies from build descriptors

kthoms:
- removed managed versions of
  - org.eclipse.emf.mwe2.launch
  - xtext-antlr-generator
  - org.junit:junit
  - org.junit.jupiter.api
  - xtext-dev-bom
- updated test expectation
- GradleBuildFile: Use single quoted dependencies for managed
dependencies (where version is null)


kthoms:
Depends on https://github.com/eclipse/xtext-lib/pull/156

org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/build.gradle
```diff
@@ -15,7 +15,7 @@ subprojects {
 	
 	apply plugin: 'java'
 	dependencies {
-		compile platform("org.eclipse.xtext:xtext-dev-bom:unspecified")
```
cdietrich: did you test that?

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/GradleBuildFile.xtend
```diff
@@ -27,7 +27,11 @@ class GradleBuildFile extends TextFile {
 						compile project(':�p.name�')
 					�ENDFOR�
 					�FOR dep : mavenDependencies�
-						�dep.scope.gradleNotation� "�dep.groupId�:�dep.artifactId�:�dep.version�"
+						�IF dep.version !== null�
+							�dep.scope.gradleNotation� "�dep.groupId�:�dep.artifactId�:�dep.version�"
```
cdietrich: wonder if we just add the if for the 3rd part.
would find it better readable

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
```diff
@@ -280,10 +279,10 @@ class RuntimeProjectDescriptor extends TestedProjectDescriptor {
 				}
 
 				dependencies {
-					mwe2 "org.eclipse.emf:org.eclipse.emf.mwe2.launch:�config.xtextVersion.mweVersion�"
```
cdietrich: should the values in the config then be marked as deprecated or are they used elsewhere?

org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/build.gradle
kthoms: for newly created projects, yes.

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/GradleBuildFile.xtend
kthoms: Did you recognize that there is a difference in single vs double quote also?

org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/build.gradle
cdietrich: :+1:

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/GradleBuildFile.xtend
cdietrich: yes

kthoms:
```
					«FOR dep : mavenDependencies»
						«dep.scope.gradleNotation» «IF dep.version !== null»"«dep.groupId»:«dep.artifactId»:«dep.version»"«ELSE»'«dep.groupId»:«dep.artifactId»'«ENDIF»
					«ENDFOR»
```
is actually less readable, or what did you mean?

cdietrich: then lets go your way

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
kthoms: good point. I'll check that. Likely they can be deprecated.

kthoms: It does not work to remove dependency versions from exec-maven-plugin :( So there versions stay mandatory.

kthoms: But we should pin xtext-antlr-generator to 2.1.1 and no open range. There is no other version and won't.

cdietrich: :+1:

kthoms:
I think we are done here for a first step. If there is potential to use managed dependencies for Tycho builds is another thing.

kthoms:
Which test projects did you mean? generateTestLanguages does not produce a change, the wizard expectations are updated.

cdietrich:
reference projects

kthoms:
ah ok, will do!

PR #1008, opened 2019-01-30
Make ResourceDescriptionStrategyFragment aware of super grammars

jschuhmacher:
such that the @Exported annotations of any super grammars are used as well. This way it behaves more like the ValidatorFragment2, where validations of super grammars are also taken into account.


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/index/ResourceDescriptionStrategyFragment.xtend
```diff
@@ -122,6 +126,9 @@ class ResourceDescriptionStrategyFragment extends AbstractInheritingFragment {
 						return createEObjectDescriptionsFor�exportedRule.type.classifier.name�(eObject, acceptor);
 					}
 				�ENDFOR�
+				�IF needsToInvokeSuperResourceDescriptionStrategy�
+				super.createEObjectDescriptions(eObject, acceptor);
```
szarnekow: Should be `return super.createEObjectDescriptions(..)` instead of `return true` in that case

szarnekow:
Thank you.

PR #1009, opened 2019-01-30
Fixes an IllegalStateException during MWE execution

jschuhmacher:
In some cases when a super grammar is used, MWE fails with an
IllegalStateException. The name used to check agains the names
administration is overwritten with the qualified name, leading to an
attempt to add the same "uniqueName" twice.

Signed-off-by: Jelle Schuhmacher <jelle.schuhmacher@altran.com>

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/xtext/RuleNames.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/xtext/RuleNames.java
@@ -146,13 +146,15 @@ public class RuleNames {
 			String uniqueName = name;
 			String antlrRuleName;
 			if (names.containsKey(name)) {
+				names.put(name, rule);
 				name = qualifiedName;
 				uniqueName = getInheritedUniqueName(rule, usedUniqueNames);
 				antlrRuleName = getInheritedAntlrRuleName(rule, usedAntlrNames);
 			} else {
+				names.put(name, rule);
 				antlrRuleName = getDefaultAntlrRuleName(rule);
 			}
-			names.put(name, rule);
+
 			if (!usedUniqueNames.add(uniqueName)) {
 				throw new IllegalStateException(uniqueName);
 			}


org.eclipse.xtext/src/org/eclipse/xtext/xtext/RuleNames.java
```diff
@@ -146,13 +146,15 @@ public RuleNames(Grammar grammar, boolean installAdapter) {
 			String uniqueName = name;
 			String antlrRuleName;
 			if (names.containsKey(name)) {
+				names.put(name, rule);
```
szarnekow: I do not understand the fix. Can you please provide a test scenario, that illustrates the problem?

szarnekow: In fact this looks wrong to me, since we put a second rule instance into the map by simple name. But the rule from the super grammar should not replace the rule from the sub-grammer (which is already in the map)

jschuhmacher: It started with the following stacktrace:
```
Caused by: java.lang.IllegalStateException: AlgorithmInterface
	at org.eclipse.xtext.xtext.RuleNames.<init>(RuleNames.java:157)
	at org.eclipse.xtext.xtext.RuleNames.getRuleNames(RuleNames.java:94)
	at org.eclipse.xtext.xtext.generator.XtextGeneratorLanguage.initialize(XtextGeneratorLanguage.java:296)
	at org.eclipse.xtext.xtext.generator.XtextGeneratorLanguage.initialize(XtextGeneratorLanguage.java:283)
	at org.eclipse.xtext.xtext.generator.StandardLanguage.initialize(StandardLanguage.java:189)
	at org.eclipse.xtext.xtext.generator.XtextGenerator.initialize(XtextGenerator.java:185)
	at org.eclipse.xtext.xtext.generator.XtextGenerator.checkConfigurationInternal(XtextGenerator.java:139)
	at org.eclipse.emf.mwe.core.lib.AbstractWorkflowComponent2.checkConfiguration(AbstractWorkflowComponent2.java:21)
	at org.eclipse.emf.mwe.core.lib.Mwe2Bridge.preInvoke(Mwe2Bridge.java:65)
	at org.eclipse.emf.mwe.core.lib.AbstractWorkflowComponent.preInvoke(AbstractWorkflowComponent.java:197)
	at org.eclipse.emf.mwe2.runtime.workflow.AbstractCompositeWorkflowComponent.preInvoke(AbstractCompositeWorkflowComponent.java:29)
	at org.eclipse.emf.mwe2.runtime.workflow.Workflow.run(Workflow.java:18)
	at org.eclipse.emf.mwe2.launch.runtime.Mwe2Runner.run(Mwe2Runner.java:102)
	... 4 more
```

The debugger showed that the value of `uniqueName` here is not qualified which lead to a problem later in the block:

```
if (!usedUniqueNames.add(uniqueName)) {
	throw new IllegalStateException(uniqueName);
}
```

So I thought the rule was meant to be overwritten in names.

Unfortunately the grammars are large, and I cannot share them. I'll try to figure out a minimal set that shows the problem.

cdietrich:
looks like https://bugs.eclipse.org/bugs/show_bug.cgi?id=486657

cdietrich:
@jschuhmacher @szarnekow what is the state of this pr?

szarnekow:
Without the repro, I'm slightly worried about the change.

jschuhmacher:
I've so far not been successful in creating a repro. Shall I close it for now and reopen when one is available?

PR #1010, opened 2019-01-30
Added RepeatedTest annotation to xtext.testing

szarnekow:
Can be used to run tests in a loop for n times

### DIFF
--- a/org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/RepeatedTest.java
+++ b/org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/RepeatedTest.java
@@ -1,0 +1,92 @@
+/*******************************************************************************
+ * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


kthoms:
This class makes use of JUnit 4 rules API. I know that there is already the Flaky rule in this package, but can we consider to move this new class to a "rules" package for clearer separation?

kthoms:
Separation of JUnit 4 only API. We have the same with the "extensions" package for JUnit 5 only extensions.

cdietrich:
id like it to keep as is

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/RepeatedTest.java
```diff
@@ -0,0 +1,94 @@
...
+	/**
+	 * Test rule used along with {@link RepeatTest} to let a test loop for a couple of rounds.
+	 */
+	public static class Rule implements TestRule {
+
+		private final boolean quiet;
+
+		/**
+		 * @param quiet
+		 *            set to true if log output is desired
```
szarnekow: Meh ... JavaDoc is bogus - will fix it immediatly

PR #1014, opened 2019-02-04
Fix ui tests

ArneDeutsch:
Needed to fix failing UI tests (Actually the JUnit 5 minimal version is 5.0.0 instead of 5.1.0 for eclipse oxygen).


kthoms:
Side note: Oxygen.3a release repo contains org.junit.jupiter.api=5.0.0.v20170910-2246

PR #1030, opened 2019-02-11
[performance] Improve isInstantiatableSubType

kthoms:
Check the case that c == superType, which is always the case when called
from findInstantiableCompatible, and occasionally for calls from
findSubTypeInEPackage

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinker.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinker.java
@@ -218,10 +218,9 @@ public class LazyLinker extends AbstractCleaningLinker {
 	}
 
 	protected EClass findInstantiableCompatible(EClass eType) {
-		if (!isInstantiatableSubType(eType, eType)) {
+		if (eType.isAbstract() || eType.isInterface()) {
 			// check local Package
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinker.java
```diff
@@ -277,7 +277,7 @@ protected EClass findSubTypeInEPackage(EPackage ePackage, EClass superType) {
 	}
 
 	private boolean isInstantiatableSubType(EClass c, EClass superType) {
-		return !c.isAbstract() && !c.isInterface() && EcoreUtil2.isAssignableFrom(superType, c);
+		return !c.isAbstract() && !c.isInterface() && (c == superType || EcoreUtil2.isAssignableFrom(superType, c));
```
cdietrich: why here and not at org.eclipse.xtext.EcoreUtil2.isAssignableFrom(EClass, EClass)

kthoms: right, would be the better place
I wonder if we avoid the call from findInstantiableCompatible at all, it's just unneccessary

kthoms:
changed to short-cut the identity check case and improved EcoreUtil2 with separate PR

PR #1031, opened 2019-02-11
isAssignableFrom: check both args for null & identity

kthoms:
Added covering test case

makeInvalid moved below single method that uses it & made private

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
@@ -403,10 +403,14 @@ public class EcoreUtil2 extends EcoreUtil {
 	 * Returns whether the given super type is the same as, or a super type of, some other class.
 	 * @param superType the super type
 	 * @param candidate the subtype
-	 * @return whether the super type is the same as, or a super type of, some other class.
+	 * @return whether the super type is the same as, or a super type of, some other class. Yields <code>null</code>
+	 * when either argument is <code>null</code>.
 	 */
 	public static boolean isAssignableFrom(EClass superType, EClass candidate) {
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
```diff
@@ -403,10 +403,14 @@ private static void collectAllSuperTypes(Set<EClass> collectedTypes, EClass eCla
...
-	 * @return whether the super type is the same as, or a super type of, some other class.
+	 * @return whether the super type is the same as, or a super type of, some other class. Yields <code>null</code>
+	 * when either argument is <code>null</code>.
 	 */
 	public static boolean isAssignableFrom(EClass superType, EClass candidate) {
-		return (candidate != null && (superType == EcorePackage.Literals.EOBJECT || superType.isSuperTypeOf(candidate)));
+		if (superType == null || candidate == null) {
+			return false;
+		}
+		return superType == candidate || superType == EcorePackage.Literals.EOBJECT || superType.isSuperTypeOf(candidate);
```
szarnekow: I wonder why this is an improvement over the immediate identity check in Eclass.isSuperTypeOf. Can you please elaborate?

kthoms: I changed that when the method was still called by LazyLinker#findInstantiableCompatible, which called that method very frequently with identical args. This just did avoid the additional hop to isSuperTypeOf, which also was the last check here.
In the meantime, since findInstantiableCompatible does not call that method anymore, the `superType == candidate` check can be removed and leave it up to isSuperTypeOf again. The additional null check & test are still valuable.
Shall I create a PR that removes `superType == candidate`?

szarnekow: No, I'm still fine with the change. I was just curious if there was anything that you could measure / observe.

Issue #1029, opened 2019-02-11
Prefer ThreadLocal.remove over ThreadLocal.set(null) in AbstractDeclarativeValidator


szarnekow:
The latter will keep an entry in the thread map thus keeping a strong reference to the validation thread.

We should also try to reduce the number of calls to `state.get()`. Occasionally it is possible to keep the result in a local variable and reuse it later on.

See `AbstractDeclarativeValidator.MethodWrapper.invoke`
See #1021

szarnekow:
Thank you, @epoell

PR #1033, opened 2019-02-14
Solve Issue 1029

epoell:
[Issue 1029](https://github.com/eclipse/xtext-core/issues/1029)
Replaced ThreadLocal state.set(null) with state.remove(). And tried to decrement the calls of state.get() by keeping it in a variable if possible.
Might need a closer look, if this works. And sorry for the wrong number (1021 instead of 1029) in the commit! I got mixed up.

Signed-off-by: Eva Poell <epoell@itemis.com>


cdietrich:
testing is currently blocked by jcenter having no current mwe metadata unfortunately

epoell:
Is there something I can do about that, or may I just wait patiently?

cdietrich:
no we have to be patient

szarnekow:
I triggered a build

Issue #1034, opened 2019-02-14
Missing validation: Entry rule cannot be a fragment


szarnekow:
The Grammar workflow breaks if the first rule is a fragment. There is a missing validation for that. Should be a warning by default.

cdietrich:
@szarnekow for non parser rules it is already an error?
so should it be here the same?

szarnekow:
I’m not sure. Error by default is sane, but I’m not certain whether it should be configurable.

cdietrich:
we have this rule but terminals does not stick to it :)

szarnekow:
Challenge is, that it’s mostly the mwe2 config, that dictates what is valid or not

loradd:
Hi everybody, I would like to start contributing. Can I work on this issue?

kthoms:
@loradd That‘ll be awesome! Please ask when you need assistance.

szarnekow:
Sorry but @epoell is already on it

miklossy:
@loradd How about working on a similar issue, such as [#410479](https://bugs.eclipse.org/bugs/show_bug.cgi?id=410479) ?

loradd:
@miklossy I can work on [#410479](https://bugs.eclipse.org/bugs/show_bug.cgi?id=410479). Thank you!

szarnekow:
Fix was merged into master. Thank you @epoell

PR #1041, opened 2019-02-19
[#1034] Added Validation that entry rule is not a fragment

epoell:
Signed-off-by: Eva Poell <epoell@itemis.com>

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/ValidationIssue1034Test.xtend
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/ValidationIssue1034Test.xtend
@@ -1,0 +1,82 @@
+/*******************************************************************************
+ * Copyright (c) 2019 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.xtext
+
+import com.google.inject.Inject
+import org.eclipse.xtext.Grammar
+import org.eclipse.xtext.XtextStandaloneSetup
+import org.eclipse.xtext.testing.InjectWith
+import org.eclipse.xtext.testing.XtextRunner
+import org.eclipse.xtext.testing.util.ParseHelper
+import org.eclipse.xtext.testing.validation.ValidationTestHelper
+import org.eclipse.xtext.tests.AbstractXtextTests
+import org.eclipse.xtext.tests.XtextInjectorProvider
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+import static org.eclipse.xtext.XtextPackage.Literals.PARSER_RULE
+import static org.eclipse.xtext.xtext.XtextConfigurableIssueCodes.INVALID_FRAGMENT_AS_FIRST_RULE
+
+/**
+ * @author Eva Poell - Initial contribution and API
+ */
+@RunWith(XtextRunner)
+@InjectWith(XtextInjectorProvider)
+class ValidationIssue1034Test extends AbstractXtextTests {
+
+	@Inject extension ParseHelper<Grammar>
+	@Inject extension ValidationTestHelper
+
+	@Before
+	override setUp() throws Exception {
+		super.setUp();
+		with(XtextStandaloneSetup)
+	}
+
+	@Test
+	def void check_FirstRuleIsNotFragment() {
+		'''grammar com.space.Test with org.eclipse.xtext.common.Terminals
+		generate testDSL "http://space.com/test"
+		Model:
+			"type" type=ID;'''.parse.assertNoIssues
+	}
+
+	@Test
+	def void check_FirstRuleIsFragment() {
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/ValidationIssue1034Test.xtend
```diff
@@ -0,0 +1,82 @@
...
+@InjectWith(XtextInjectorProvider)
+class ValidationIssue1034Test extends AbstractXtextTests {
+
+	@Inject extension ParseHelper<Grammar>
+	@Inject extension ValidationTestHelper
+
+	@Before
+	override setUp() throws Exception {
+		super.setUp();
+		with(XtextStandaloneSetup)
```
szarnekow: Is the with-invocation necessary?

epoell: It seems to me like it.
The tests fail without that line the tests fail.

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextConfigurableIssueCodes.java
```diff
@@ -67,6 +68,8 @@ protected void initialize(IAcceptor<PreferenceKey> acceptor) {
 		acceptor.accept(create(BIDIRECTIONAL_REFERENCE, SeverityConverter.SEVERITY_WARNING));
 		acceptor.accept(create(DISCOURAGED_RULE_NAME, SeverityConverter.SEVERITY_WARNING));
 		acceptor.accept(create(USAGE_OF_DEPRECATED_RULE, SeverityConverter.SEVERITY_WARNING));
+
```
kthoms: nitpick: why empty line here?

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextConfigurableIssueCodes.java
```diff
@@ -47,6 +47,7 @@
 	public static final String INVALID_ANNOTAION = ISSUE_CODE_PREFIX + "InvalidAnnotation";
 	public static final String USAGE_OF_DEPRECATED_RULE = ISSUE_CODE_PREFIX + "UsageOfDeprecatedRule";
 	
+	public static final String INVALID_FRAGMENT_AS_FIRST_RULE = ISSUE_CODE_PREFIX + "InvalidFragmentFirstRule";
```
kthoms: nitpick: why empty line before?

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
```diff
@@ -258,6 +258,13 @@ else if (GrammarUtil.isDatatypeRule((ParserRule) firstRule)) {
 					firstRule,
 					XtextPackage.Literals.ABSTRACT_RULE__NAME, 
 					ValidationMessageAcceptor.INSIGNIFICANT_INDEX);
+		} 
+		else if (((ParserRule) firstRule).isFragment()) {
```
kthoms: nitpick: we usually place the `else` after the closing braces of the previous block

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextConfigurableIssueCodes.java
epoell: I assumed the file was organised in semantic blocks and figured, that my new error code did not fit into the block above...
Would you prefer, to drop that empty line?

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextConfigurableIssueCodes.java
epoell: Same as above.
Would you prefer to drop that empty line?

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
epoell: Me too, I wanted to imitate the rest of the method.
Would you prefer it on the same line?

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextConfigurableIssueCodes.java
kthoms: Sorry, I did not see that this was already the case above in the file. Imitating the file's existing style is right. But when I see this correctly, the blocks are "errors" and "warnings". So this should go the "errors" section.

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextValidator.java
kthoms: This method is unfortunately an example where it is handled differently than in the rest of the file. So I'd say please change the one above also to match the file's common formatting.

org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextConfigurableIssueCodes.java
epoell: Thanks for the hint! I will change that

Issue #1043, opened 2019-02-20
[ide] Bogus serialization when reassign container 


dhuebner:
See org.eclipse.xtext.ide.tests.serializer.ChangeSerializerTest.testMoveElement()

```
#22 {
	child1
	children1 {
		child2 jumper
	}
}
```
Adding `jumper` as sibling of `child1` doesn't work.

dhuebner:
We should also check  eclipse/xtext#2565

szarnekow:
I'm not sure about the impact of your changes, but please bear in mind that we are already close to 2019-3 - so maybe this should be postponed to 2.18 to allow for more involving testing?

dhuebner:
I know, therefore I didn't assigned a milestone.
I don't think we can use the 2.17 in our project anyway.
Would be nice to have a bugfix release for an older version aka LTS

szarnekow:
Makes sense. I don't that anybody would object having a maintenance release.

cdietrich:
@dhuebner is this fixed now?

dhuebner:
@cdietrich Yes

PR #1044, opened 2019-02-21
[ide] Bogus serialization when reassign container #1043

dhuebner:
All in one commit 

Main changes are in:
AbstractEObjectRegion.java
PartialSerializer.java


org.eclipse.xtext/src/org/eclipse/xtext/formatting2/regionaccess/internal/AbstractEObjectRegion.java
```diff
@@ -84,7 +87,7 @@ public EStructuralFeature getContainingFeature() {
 		if (feature == null) {
 			return null;
 		}
-		return semanticElement.eClass().getEStructuralFeature(feature);
+		return semanticElement.eContainer().eClass().getEStructuralFeature(feature);
```
szarnekow: It looks like this code is working solely on the AST level. Why can't we use EObject.eContainingFeature() ?

dhuebner: In case where AST and the regionaccess gets out of sync, we prefer to return `null` instead of the probably wrong containment feature.

szarnekow: In that case, a guard could make sense, to check if the containingFeature and the feature name from the grammar are the same. What do you think?

Smth like semanticElement.eContainingFeature.name == feature?

cdietrich:
@szarnekow @dhuebner what is the state of this pr?

dhuebner:
@cdietrich This bug cost a lot of time to find and fix, so I would love to see it merged ASAP.

dhuebner:
@cdietrich Could you merge the PR?

cdietrich:
@dhuebner my assumption is that you are still committer. comitters can and should merge their pr once they are approved

PR #1045, opened 2019-02-21
Prevent NPE in InternalNodeModelUtils

tivervac:


org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/InternalNodeModelUtils.java
```diff
@@ -140,7 +140,8 @@ private static LineAndColumn getLineAndColumnOfLineBreak(String text, int[] line
 
 	private static int[] getLineBreakOffsets(INode rootNode) {
 		if (rootNode instanceof RootNode) {
-			return ((RootNode) rootNode).basicGetLineBreakOffsets();
+			int[] lineBreakOffsets = ((RootNode) rootNode).basicGetLineBreakOffsets();
```
szarnekow: Under which cicrumstances does this happen?

tivervac: We have a class that extends `EObjectDiagnosticImpl`

Its hashcode is as follows:
```
	override hashCode() {
		toString().hashCode
	}
```

This `hashCode` gets called by a `HashSet.contains()`. The `toString` then calls `getLine()` which eventually ends up in the fixed case.

Sadly, I cannot reproduce the issue, I only have a stacktrace

tivervac: Said stacktrace (happens during scoping):
```
    at java.util.Arrays.binarySearch(Arrays.java:1864)
    at org.eclipse.xtext.nodemodel.impl.InternalNodeModelUtils.getLineAndColumn(InternalNodeModelUtils.java:62)
    at org.eclipse.xtext.nodemodel.impl.InternalNodeModelUtils.getLineAndColumn(InternalNodeModelUtils.java:49)
    at org.eclipse.xtext.nodemodel.impl.AbstractNode.basicGetLineOfOffset(AbstractNode.java:175)
    at org.eclipse.xtext.nodemodel.impl.AbstractNode.getStartLine(AbstractNode.java:183)
    at org.eclipse.xtext.diagnostics.AbstractDiagnostic.getLine(AbstractDiagnostic.java:58)
    at org.eclipse.xtext.diagnostics.AbstractDiagnostic.toString(AbstractDiagnostic.java:84)
    at com.sigasi.LinkingErrorDiagnostic.toString()(LinkingErrorDiagnostic.java:40)
    at com.sigasi.LinkingErrorDiagnostic.hashCode()(LinkingErrorDiagnostic.java:17)
    at java.util.HashMap.hash(HashMap.java:338)
    at java.util.HashMap.containsKey(HashMap.java:595)
    at java.util.HashSet.contains(HashSet.java:203)
```

szarnekow: Do you use any kind of artificial root nodes / nodes in the tree?

tivervac: AFAIK we don't

tivervac: Looking at it a bit closer we're very confused about the matter as well. Building is single threaded so scoping can only run after the entire node model was made, no races should be possible by design. Also, right before invoking the `Arrays.binarySearch` we checked that the `text.length > 0`. Which means that `rootNode.completeContent` was available, while `rootNode.lineBreakOffsets` wasn't. As those are set together in `RootNode.basicSetCompleteContent(String)` and races shouldn't be possible, this is very odd behaviour. It's also the first time I see this stacktrace so it obviously doesn't happen often.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/InternalNodeModelUtils.java
```diff
@@ -140,7 +140,8 @@ private static LineAndColumn getLineAndColumnOfLineBreak(String text, int[] line
...
+			return (lineBreakOffsets != null) ? lineBreakOffsets : new int[] {};
```
szarnekow: Should we fallback to the else case under these circumstances?

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/InternalNodeModelUtils.java
szarnekow: I'm ok with hardening the code here. It would still be interesting to learn, how this could be null.

tivervac: @lievenlemiengre What do you think? Do we fix it here, or try to fix it earlier?

lievenlemiengre: Somehow an invariant is broken here, we can remedy the symptom but the cause is unknown.

I wouldn't push this fix, we can see if we can add some sanity checks in our own code to get us closer to the cause.

szarnekow: Appreciated :)

szarnekow:
Can we discard this?

tivervac:
Yes

PR #1046, opened 2019-02-21
Actually delete LSP test folder after test

tivervac:
Delete on exit doesn't work if the folder is not empty. Since we don't
remove the individual files the folder doesn't get deleted.


org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -106,11 +108,14 @@ abstract class AbstractLanguageServerTest implements Endpoint {
 		languageServer.supportedMethods()
 
 		// create workingdir
-		root = new File(new File("").absoluteFile, "/test-data/test-project")
-		if (!root.mkdirs) {
-			Files.cleanFolder(root, null, true, false)
+		root = new File(new File("").absoluteFile, TEST_PROJECT_PATH)
+	}
+
+	@After
```
szarnekow: Missing import. Also missing `@AfterEach`

PR #1051, opened 2019-02-26
GH-1047: Implemented `textDocument/prepareRename`

### LINKED ISSUE #1047: [LS] Support `textDocument/prepareRename`
The [`textDocument/prepareRename`](https://microsoft.github.io/language-server-protocol/specification#textDocument_prepareRename) language feature can be used to check whether the [`textDocument/rename`](https://microsoft.github.io/language-server-protocol/specification#textDocument_rename) can be executed or not.

The problem:
 - Let's have `this.is.a.fqn.MyTypeDeclaration`.
 - Let's have a selection: `this.is<|>.a.fqn.MyTypeDeclaration`.
 - Press `Rename Symbol` in the monaco editor: `is` will be proposed for the rename, rename to `foo`.
 - The `RenameService` [looks up the `EObject`](https://github.com/eclipse/xtext-core/blob/4f2f3f70228e5c668368299dc3784c54c71d2e7b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService.xtend#L56), finds `MyTypeDeclaration`, and renames it. So we will end up `this.is.a.fqn.foo` instead of `this.foo.a.fqn.MyTypeDeclaration`.
 - Instead of perfoming the `Rename Symbol`, we should not allow it by default.

A good analogy for this issue is when one tries to rename the package declaration in Java. The JDT LS is not complete either when trying to rename the package, but most of the uses-cases are covered; it tells the user `Renaming this element is not supported` in VS Code.

kittaakos:
- `textDocument/prepareRename` is automatically executed
when the client declares `prepareSupport`.
 - Exposed the `InitializeResult` on the LS access API.
 - From now on, the refactoring issue acceptor does not throw
an exception when adding issues with `ERROR` or `FATAL` severity.
 - Extended the test language to support FQNs via package declarations.

Closes #1047.


org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/serializer/TestLanguageReferenceUpdater.xtend
```diff
@@ -0,0 +1,56 @@
...
+import org.eclipse.xtext.scoping.IScopeProvider
+import org.eclipse.xtext.serializer.tokens.SerializerScopeProviderBinding
+
+/**
+ * Customized reference updater to handle FQN renaming gracefully.
+ * 
+ * <p>
+ * When renaming {@code Foo} to {@code Bar}, it follows the default logic,
+ * but when renaming {@code my.type.Foo} to {@code Bar}, it will result
+ * in {@code my.Type.Bar} instead of {@code Bar}.
```
kittaakos: `my.Type.Bar` -> `my.type.Bar`.

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/serializer/TestLanguageReferenceUpdater.xtend
```diff
@@ -0,0 +1,56 @@
...
+	@Inject
+	IQualifiedNameConverter nameConverter;
+
+	@Inject
+	@SerializerScopeProviderBinding
+	IScopeProvider scopeProvider;
+
+	override void updateReference(ITextRegionDiffBuilder rewriter, IUpdatableReference upd) {
+		val updatable = upd;
+		if (rewriter.isModified(updatable.getReferenceRegion())) {
```
kittaakos: TODO: Xtendify this copied code.

kthoms:
I'm afraid this comes too late for 2.17. Today is M3 and this means feature freeze. Please do not merge before March 6th.

kittaakos:
> I'm afraid this comes too late for 2.17.

👍 Thanks for the note. That's not a problem. I will update it to `2.18`.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService.xtend
```diff
@@ -42,8 +54,17 @@ class RenameService implements IRenameService {
...
+	Function<EObject, String> attributeResolver = SimpleAttributeResolver.newResolver(String, 'name')
+	
 	override rename(WorkspaceManager workspaceManager, RenameParams renameParams, CancelIndicator cancelIndicator) {
-		val uri = renameParams.textDocument.uri.toUri
+		val textDocument = renameParams.textDocument
+		val uri = textDocument.uri.toUri
+		val positionParams = new TextDocumentPositionParams(new TextDocumentIdentifier(textDocument.uri), renameParams.position)
+		val rangeToRename = prepareRename(workspaceManager, positionParams, cancelIndicator)
+		if (rangeToRename === null|| rangeToRename.getLeft === null || !rangeToRename.getLeft.containsPosition(renameParams.position)) {
+			return null
```
JanKoehnlein: Should we also add an issue to the issueAcceptor?

kittaakos: This was done here: https://github.com/eclipse/xtext-core/pull/1051/files#diff-2a69e2e40efb5ed1e922cb1546ab8cecR104

kittaakos:
`2.17.0` is out; this PR is ready for the review.

cdietrich:
@kittaakos can you rebase this on eclipse/master?

kittaakos:
I have rebased this from the `master` (8e274f3daf75edad43a16339837f83e145445d25).

cdietrich:
@kittaakos look like this need a rebase with jans recent changes

cdietrich:
@kittaakos  please rebase and resolve the merge issues

kittaakos:
> @kittaakos please rebase and resolve the merge issues

Yeah, I am on it, I think this will be ready for the review today.

kittaakos:
I have adjusted this PR to the new `IRenameService2`'s API/behavior. Please review.

cdietrich:
@kittaakos should the `prepareSupport` flag set somewhere

kittaakos:
> @kittaakos should the `prepareSupport` flag set somewhere

Good point, I am glad you ask it. What do you think? I wanted to introduce a protected boolean method on the `RenameService2` with default `true`. If one configures it to `false`, then the `prepareRename` would not run before `rename`. Are you aware of any `prepareSupport` in the LSP, by the way, or you want it for Xtext only?

cdietrich:
@kittaakos i have no idea how this is used in practice. i just stumbled on it in the spec. am not sure how we deal with such (client) capabilities elsewhere

kittaakos:
> i just stumbled on it in the spec.

This is what I mean, I have overlooked it. `prepareSupport` is a client capability, nice. Thanks for checking it for me. I will update the PR in short.

cdietrich:
@kittaakos see also

```
/**
	 * The server provides rename support. RenameOptions may only be
	 * specified if the client states that it supports
	 * `prepareSupport` in its initial `initialize` request.
	 */
	renameProvider?: boolean | RenameOptions;
```
for the server side

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
```diff
@@ -95,13 +86,13 @@ class RenameService2 implements IRenameService2 {
...
 				val position = options.renameParams.position
-				var EObject element;
+				var EObject element
 				try {
 					element = xtextResource.getElementAtOffset(context.document, position)
 				} catch (IndexOutOfBoundsException exc) {
 					issueAcceptor.add(FATAL, '''Invalid document �position.toPositionFragment(uri)�''')
 				}
-				if (issueAcceptor.maximumSeverity !== FATAL && (element === null || element.eIsProxy)) {
+				if (issueAcceptor.maximumSeverity !== FATAL && element === null || element.eIsProxy) {
```
JanKoehnlein: Doesn't that now throw an NPE when element is `null` and the severity is not FATAL?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
```diff
@@ -83,25 +114,96 @@ class RenameService2 implements IRenameService2 {
...
+
 	protected def isIdentifier(ILeafNode leafNode) {
 		return (leafNode.grammarElement instanceof RuleCall || leafNode.grammarElement instanceof CrossReference)
-			&& !tokenUtil.isWhitespaceOrCommentNode(leafNode) 
+			&& !tokenUtil.isWhitespaceOrCommentNode(leafNode)
+	}
+
+	override prepareRename(PrepareRenameOptions options) {
+		val uri = options.params.textDocument.uri
+		options.languageServerAccess.doRead(uri) [ context |
```
JanKoehnlein: When called from `rename`, should we make sure to use a single `doRead()` transaction? The builder could kick in and change the model in between.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
kittaakos: Funny, I could have asked the same; it's your code. So what was your intention? First, I change it because it looked odd, then I reverted.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
JanKoehnlein: - If it's already fatal, don't add another error (it's likely a follow up)
- If not, and the element could not be detected (null) or is a unresolvable line (proxy, guarded by the former check), put accept the error
Am I missing something?

kittaakos: OK, it makes sense; but overriding clients can still mess it up easily. I will do it in one `doRead` for the default case.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
JanKoehnlein: - If it's already fatal, don't add another error (it's likely a follow up)
- If not, and the element could not be detected (null) or is a unresolvable line (proxy, guarded by the former check), put accept the error
    Am I missing something?

kittaakos: Also, why to check `issueAcceptor.maximumSeverity !== FATAL` at all? If it has `ERROR` or `FATAL`, it already failed so it does not matter. Otherwise, our intention is to fail here. Thanks!

JanKoehnlein: I see. It may have been a bad idea to throw an exception in the issueAcceptor on ERRORS and WARNINGS. Clients should not have to rely on such assumptions. Feel free to refactor.

The entire thing was inspired by Eclipse LTK, where you can collect all kinds of errors and warnings on refactoring and report them to the user. LSP does not provide any proper error channel, and the entire thing got mixed up in order to have a simpler control flow.

kittaakos: OK, I will change the issue acceptor.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
kittaakos: [Done](https://github.com/eclipse/xtext-core/pull/1051/files#diff-f0264dd6c439d644cd998f039eba6d21R81).

kittaakos:
I consider this PR to be ready for the review.

cdietrich:
@JanKoehnlein please review

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
```diff
@@ -21,48 +32,74 @@ import org.eclipse.xtext.ide.refactoring.RenameChange
...
+				val identifier = new TextDocumentIdentifier(textDocument.uri)
+				val position = options.renameParams.position
+				val positionParams = new TextDocumentPositionParams(identifier, position)
+				val resource = context.resource
+				val document = context.document
+				val cancelIndicator = options.cancelIndicator
+
+				val prepareRenameResult = doPrepareRename(resource, document, positionParams, cancelIndicator)
+				if (!mayPerformRename(prepareRenameResult, options.renameParams)) {
+					return null
```
JanKoehnlein: Shouldn't we report an issue here?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
```diff
@@ -77,31 +114,137 @@ class RenameService2 implements IRenameService2 {
 			} else {
 				issueAcceptor.add(FATAL, 'Loaded resource is not an XtextResource', context.resource.URI)
 			}
+			issueAcceptor.checkSeverity
 			return workspaceEdit
+		].exceptionally [ exception |
+			val rootCause = Throwables.getRootCause(exception);
+			if (rootCause instanceof FileNotFoundException) {
+				if (shouldPrepareRename) {
+					return null
```
JanKoehnlein: Report issue?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
```diff
@@ -77,31 +114,137 @@ class RenameService2 implements IRenameService2 {
...
+
 	protected def isIdentifier(ILeafNode leafNode) {
 		return (leafNode.grammarElement instanceof RuleCall || leafNode.grammarElement instanceof CrossReference)
-			&& !tokenUtil.isWhitespaceOrCommentNode(leafNode) 
+			&& !tokenUtil.isWhitespaceOrCommentNode(leafNode)
+	}
+
+	override prepareRename(PrepareRenameOptions options) {
+		val uri = options.params.textDocument.uri
+		val shouldPrepareRename = options.languageServerAccess.shouldPrepareRename
```
JanKoehnlein: If that is false, shouldn't we skip all the rest of this method?

kittaakos: No, you can invoke `textDocument/prepareRename` without calling `textDocument/rename`.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
kittaakos: I did not want to. It is already traced, one can glance into the logs. From the LSP's perspective, it does not add any value if we "reporting" the issues.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
kittaakos: Yes, you're right, but if `shouldPrepareRename` is `false`, we still have to throw an exception if the resource does not exist. It is not clear from the spec, whether calling `textDocument/prepareRename` itself is a valid call or it is always an implicit precheck before `textDocument/rename`.

kittaakos: It is [done](https://github.com/eclipse/xtext-core/commit/6bb48ebd57781ecb175934e3bab73ab1745d87e8#diff-f0264dd6c439d644cd998f039eba6d21R160).

kittaakos:
Thanks for the review, @JanKoehnlein, and for looking into LSP, @cdietrich 👏

cdietrich:
thanks @kittaakos

PR #1060, opened 2019-03-04
[eclipse/xtext-core#1059] updated to lsp4j 0.7.0

cdietrich:


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolMapper.xtend
```diff
@@ -48,7 +48,11 @@ class DocumentSymbolMapper {
 	 */
 	def DocumentSymbol toDocumentSymbol(EObject object) {
 		return new DocumentSymbol => [
-			name = nameProvider.getName(object);
+			val pName = nameProvider.getName(object) {
+				if (pName !== null) {
+					name = pName;
+				}
+			}
```
spoenemann: This code looks strange. What are the curly braces around the `if` statement doing?

cdietrich: arrrg. will correct

cdietrich: corrected

kittaakos: @cdietrich, can you please explain why do we need the `null` check here?

cdietrich: cause of this fancy new check:
https://github.com/eclipse/lsp4j/blob/58bd73af0b7b62610e626aaaf1a625c1d57ccb1d/org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/DocumentSymbol.java#L107

cdietrich: will add a check for the others too

kittaakos: I see. :+1: Thank you! Quite an inconvenient new lsp4j feature.

kittaakos: Then we need the `null` check for almost all other setters, don't we? Why do we have this new fail fast "feature" in lsp4j, @spoenemann?

cdietrich: done

kittaakos: Well, I think this is an LSP4J issue, we use it in JDT too, perhaps other projects as well, we cannot just change the runtime behavior from `0.6.0` to `0.7.0`.

cdietrich: yes and no. the old impl was to use the stuff created here as "schmierzettel" and sort out the bad ones later. see org.eclipse.xtext.ide.server.symbol.HierarchicalDocumentSymbolService.isValid(DocumentSymbol)

cdietrich: the same is still used with my change. it avoid the illegalarguments only

spoenemann: Here's the origin issue: https://github.com/eclipse/lsp4j/issues/284

cdietrich: @kittaakos how to proceed? go with how it is now and open a lsp4j issue?

kittaakos: I would either revert the behavior change in LSP4J or accept it and add the `null` checks to the client code, but opening a new LSP4J ticket does not make much sense to me. What do you think, @spoenemann? Is it urgent to decide on this, @cdietrich?

cdietrich: well i dont know when and if people might want to update lsp4j with xtext

spoenemann: But do we really need all those null guards? Have you confirmed that `nameProvider`, `kindProvider` and `rangeProvider` can actually return null values? If yes, isn't that a bug in those providers? In those cases we could as well just let the exception bubble up so an error response message is sent to the client.

cdietrich: @spoenemann i added null guards to the places that currently can return null

cdietrich: an as i said: the current impl uses schmiezettel logik and throws out the improper objects later

Issue #923, opened 2018-10-26
[LSP] API of ICodeActionService has no access to WorkspaceManager


cdietrich:
the current API work ICodeActionService looks like

`def List<Either<Command, CodeAction>> getCodeActions(Document document, XtextResource resource, CodeActionParams params, CancelIndicator indicator)`

as you can see the implementer of code actions has no access to the `WorkspaceManager`

lets compare it with the `IRenameService` API

`def WorkspaceEdit rename(WorkspaceManager workspaceManager, RenameParams renameParams, CancelIndicator cancelIndicator)`

One usecase for CodeActions is Quickfixes. Lets asume we have a model that looks like an enhanced Greeting example and a model like this

```
Hello xtext!
Hello Quickfix from xtext!
```

Then lets assume we want to offer a (one pass) code action that fixes the lower case `xtext` to upper case `Xtext`. that would be a perfect fit to use the change serializer as the Rename Service does, unfortunately i need a `WorkspaceManager` to do so.

(see attached file)

@svenefftinge @kittaakos @dhuebner what do you think? do i overlook something? should we change the api? should we add a ICodeActionServiceExtension?

cdietrich:
[MyDslCodeActionService.xtend.txt](https://github.com/eclipse/xtext-core/files/2518283/MyDslCodeActionService.xtend.txt)

cdietrich:
org.eclipse.xtext.ide.tests.testlanguage.ide.server.CodeActionService.recordWorkspaceEdit(Document, XtextResource, IModification<Resource>) tricks around that by not using manager at all (and e.g. ignoring dirty state)

dhuebner:
@cdietrich Adding access to the resources did make sense at least from the point we added CodeAction as return type. I'm not quite sure that it should be the `WorkspaceManager` class because the only method being called is `doRead(URI uri, (Document, XtextResource)=>T work)` For me it  looks more like a kind of `IResourceAccess` is needed. However I would prefer an extension interface.

cdietrich:
hmm the workspace manager is already part of the change serializer api

svenefftinge:
We also have `ILanguageServerAccess` which also provides the underlying Document. 
Maybe we should add an `ILspDocumentAccess` that only has the `doRead` method and have ILanguageServerAccess extend it.

svenefftinge:
[...]

PR #1068, opened 2019-03-11
GH-1067: Exposed the Xtext index to the LS access.

### LINKED ISSUE #1067: [LS] As an LS extension developer, I would like to access the Xtext index
As an LS extension developer (`ILanguageServerExtension`), I can load any Xtext resource given by its URI via the `ILanguageServerAccess`. I would like to access the Xtext index, and read its content, in such a convenient way. The index can be retrieved from the `WorkspaceManager` but injecting it into my LS extension is not an option, I would get different instances.

I propose extending the `ILanguageServerAccess` API with something like this:
```java
def CompletableFuture<Void> doReadIndex(Consumer<IResourceDescription> consumer)
```

kittaakos:
From now on, it is possible to read the content of the Xtext
index from a LS access.

Closes #1067.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ILanguageServerAccess.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ILanguageServerAccess.xtend
@@ -18,6 +18,7 @@ import org.eclipse.lsp4j.services.LanguageClient
 import org.eclipse.xtend.lib.annotations.Data
 import org.eclipse.xtext.ide.serializer.IChangeSerializer
 import org.eclipse.xtext.resource.IResourceDescription
+import org.eclipse.xtext.resource.IResourceDescriptions
 import org.eclipse.xtext.util.CancelIndicator
 
 /**
@@ -35,11 +36,24 @@ interface ILanguageServerAccess {
 		CancelIndicator cancelChecker
 	}
 
+	@Data
+	static class IndexContext {
+		IResourceDescriptions index
+		CancelIndicator cancelChecker
+	}
+
 	/**
 	 * provides read access to a fully resolved resource and Document.
 	 */
 	def <T> CompletableFuture<T> doRead(String uri, Function<Context, T> function)
-	
+
+	/**
+	 * Provides read access to the Xtext index.
+	 * 
+	 * @since 2.18
+	 */
+	def <T> CompletableFuture<T> doReadIndex(Function<? super IndexContext, ? extends T> function)
+
 	static interface IBuildListener {
 		def void afterBuild(List<IResourceDescription.Delta> deltas)
 	}
@@ -62,9 +76,9 @@ interface ILanguageServerAccess {
 	 * In order not to mess up the originals, the resp. models should be loaded into a
 	 * new resource set which this method provides.
 	 * 
-	 * @param uri a file URI used to detect the project to configure the scoüpe of the resource set.
+	 * @param uri a file URI used to detect the project to configure the scope of the resource set.
 	 * @return a new empty resource set, configured with the project the <code>uri</code>
-	 *   belongs to and the {@link ResourceDescriptionsProvider.LIVE_SCOPE} in order to
+	 *   belongs to and the {@link ResourceDescriptionsProvider.LIVE_SCOPE} in order to
 	 *   reflect model changes immediately.
 	 * @since 2.18
 	 */
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
@@ -712,12 +712,12 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 	ILanguageServerAccess access = new ILanguageServerAccess () {
 		
 		override <T> doRead(String uri, Function<Context, T> function) {
-				requestManager.runRead [ cancelIndicator |
-					workspaceManager.doRead(uri.toUri) [ document, resource |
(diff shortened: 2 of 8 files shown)


cdietrich:
@kittaakos how does this semantically differ from WorkspaceSymbolService?

svenefftinge:
You can get the `IResourceDescriptions` through
`resourceDescriptionsProvider.getResourceDescriptions(context.getResource().getResourceSet())`
Would that be good enough for your use case?

kittaakos:
> Would that be good enough for your use case?

I think no. I do not have any URIs, so I do not have the `Context` object.

kittaakos:
> how does this semantically differ from WorkspaceSymbolService?

I would like to collect and return with my custom objects and not the `SymbolInformation`.

cdietrich:
but aint that a workaround e.g. around the same problem https://github.com/eclipse/xtext-core/issues/923 has

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -706,7 +708,11 @@ import org.eclipse.xtext.findReferences.IReferenceFinder.IResourceAccess
 		override getLanguageClient() {
 			client
 		}
-		
+
+		override doReadIndex(Consumer<IResourceDescription> consumer) {
+			workspaceManager.index.allResourceDescriptions.forEach(consumer)
```
svenefftinge: This should run within `requestManager.runRead [ cancelIndicator | ... ]`.
Also it would be good to provide access to the full `IResourceDescriptions` and return a `CompletableFuture<T>`. So it is symmetric.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -706,7 +707,13 @@ import org.eclipse.xtext.findReferences.IReferenceFinder.IResourceAccess
 		override getLanguageClient() {
 			client
 		}
-		
+
+		override <T> doReadIndex(Function<IResourceDescriptions, T> function) {
+			requestManager.runRead[ cancelIndicator |
+				return function.apply(workspaceManager.index)
```
svenefftinge: we also need to pass the cancelIndicator.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ILanguageServerAccess.xtend
```diff
@@ -32,11 +33,24 @@ interface ILanguageServerAccess {
...
 	 */
 	def <T> CompletableFuture<T> doRead(String uri, Function<Context, T> function)
-	
+
+	/**
+	 * Provides read access to the Xtext index.
+	 * 
+	 * @since 2.18
+	 */
+	def <T> CompletableFuture<T> doReadIndex(Function<IndexContext, T> function)
```
szarnekow: This should be `Function<? super IndexContent, ? extends T>`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ILanguageServerAccess.xtend
```diff
@@ -32,11 +33,24 @@ interface ILanguageServerAccess {
 		CancelIndicator cancelChecker
 	}
 
+	@Data
+	static class IndexContext {
+		IResourceDescriptions index
+		CancelIndicator cancelChecker
```
szarnekow: Why `cancelChecker` instead of `cancelIndicator`?

szarnekow: I see  - since it's called cancelChecker above.

cdietrich:
@kittaakos this pr needs to be rebased.

kittaakos:
> @kittaakos this pr needs to be rebased.

Ready for the review.

cdietrich:
@szarnekow what do you think?

org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/server/ILanguageServerAccess.java
```diff
@@ -148,9 +220,9 @@ public CancelIndicator getCancelChecker() {
    * In order not to mess up the originals, the resp. models should be loaded into a
    * new resource set which this method provides.
    * 
-   * @param uri a file URI used to detect the project to configure the sco�pe of the resource set.
+   * @param uri a file URI used to detect the project to configure the sco�pe of the resource set.
```
szarnekow: Encoding problems?

org.eclipse.xtext.ide/xtend-gen/org/eclipse/xtext/ide/server/ILanguageServerAccess.java
```diff
@@ -148,9 +220,9 @@ public CancelIndicator getCancelChecker() {
...
    * @return a new empty resource set, configured with the project the <code>uri</code>
-   *   belongs to and the {@link ResourceDescriptionsProvider.LIVE_SCOPE}�in order to
+   *   belongs to and the {@link ResourceDescriptionsProvider.LIVE_SCOPE}�in order to
```
szarnekow: Encoding problems

kittaakos:
> Encoding looks broken at two places.

Thanks for spotting it, I have fixed them. Please stop using <kbd>Alt</kbd>+<kbd>Space</kbd> in the source code.

cdietrich:
thx @kittaakos

PR #1078, opened 2019-03-14
[LS] overhauled rename service

### LINKED ISSUE #916: [LSP] Improve re-name target matching in RenameService
```
override rename(CancelIndicator cancelIndicator<|>) {
}
```

Having cursor at position `<|>` (e.g. in VSCode wenn double clicking `cancelIndicator`) will trigger rename for the function and not the parameter `cancelIndicator`

### LINKED ISSUE #917: [LSP] Use object definition services when renaming a cross reference to an other language
Currently RenameService uses the the services for the language that triggers the rename refactoring.

Language A
```
type MyType
```

Language B
```
def myFun(MyType t) {}
``` 
When triggering re-name MyType at `MyType t` the Language A should be used do the refactoring.

### LINKED ISSUE #1072: [LSP] support versioned documents in WS edits returned by rename
The rename service currently returns a workspace edit that doesn't take document versions into account. It is therefore prone to race errors when the document changes between the calculation of the edit and its application.

We should use `TextDocumentEdit`s with `VerisonedTextDocumentIdentifier`when the client supports this (`WorkspaceClientCapabilities#workspaceEdit#documentChanges`).

JanKoehnlein:
- delegate to rename language of the declaration of the renamed target
- allow to trigger rename with caret right after the identifier
- support versioned documents

Closes #917, closes #916 and closes #1072

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
@@ -102,6 +102,7 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 import com.google.common.collect.ImmutableMultimap
 import com.google.common.collect.ImmutableMap
 import org.eclipse.xtext.findReferences.IReferenceFinder.IResourceAccess
+import org.eclipse.xtext.ide.server.rename.IRenameServiceExtension
 
 /**
  * @author Sven Efftinge - Initial contribution and API
@@ -607,14 +608,19 @@ import org.eclipse.xtext.findReferences.IReferenceFinder.IResourceAccess
 		throw new UnsupportedOperationException("TODO: auto-generated method stub")
 	}
 
-	override rename(RenameParams params) {
+	override rename(RenameParams renameParams) {
 		return requestManager.runRead[ cancelIndicator |
-			val uri = params.textDocument.uri.toUri
+			val uri = renameParams.textDocument.uri.toUri
 			val resourceServiceProvider = uri.resourceServiceProvider
 			val renameService = resourceServiceProvider?.get(IRenameService)
 			if (renameService === null)
 				return new WorkspaceEdit
-			renameService.rename(workspaceManager, params, cancelIndicator)
+			if (renameService instanceof IRenameServiceExtension) {
+				val options = new IRenameServiceExtension.Options(params?.capabilities?.workspace?.workspaceEdit?.documentChanges === Boolean.TRUE)
+				renameService.rename(workspaceManager, renameParams, options, cancelIndicator)
+			} else {
+				renameService.rename(workspaceManager, renameParams, cancelIndicator)
+			}
 		]
 	}
 	
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
@@ -14,10 +14,14 @@ import org.eclipse.emf.common.util.URI
 import org.eclipse.emf.ecore.resource.Resource
 import org.eclipse.emf.ecore.xmi.XMLResource
 import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.TextDocumentEdit
 import org.eclipse.lsp4j.TextEdit
+import org.eclipse.lsp4j.VersionedTextDocumentIdentifier
 import org.eclipse.lsp4j.WorkspaceEdit
+import org.eclipse.lsp4j.jsonrpc.messages.Either
 import org.eclipse.xtext.ide.serializer.IEmfResourceChange
 import org.eclipse.xtext.ide.serializer.ITextDocumentChange
+import org.eclipse.xtext.ide.server.Document
 import org.eclipse.xtext.ide.server.WorkspaceManager
 import org.eclipse.xtext.parser.IEncodingProvider
 import org.eclipse.xtext.resource.IResourceServiceProvider
@@ -31,21 +35,35 @@ class ChangeConverter implements IAcceptor<IEmfResourceChange> {
 
 	static class Factory {
 
-		@Inject IResourceServiceProvider.Registry registry
+		@Inject protected IResourceServiceProvider.Registry registry
 
+		/**
+		 * @deprecated use {@link #create(WorkspaceManager, WorkspaceEdit, boolean} instead.
+		 */
+		@Deprecated
 		def ChangeConverter create(WorkspaceManager workspaceManager, WorkspaceEdit edit) {
-			new ChangeConverter(workspaceManager, registry, edit)
+			new ChangeConverter(workspaceManager, registry, edit, null)
+		}
+
+		def ChangeConverter create(WorkspaceManager workspaceManager, WorkspaceEdit edit, IRenameServiceExtension.Options options) {
+			new ChangeConverter(workspaceManager, registry, edit, options)
 		}
 	}
 
 	val WorkspaceManager workspaceManager
 	val IResourceServiceProvider.Registry registry
 	val WorkspaceEdit edit
+	val IRenameServiceExtension.Options options
 	
-	protected new(WorkspaceManager workspaceManager, IResourceServiceProvider.Registry registry, WorkspaceEdit edit) {
+	protected new(WorkspaceManager workspaceManager, IResourceServiceProvider.Registry registry, WorkspaceEdit edit, IRenameServiceExtension.Options options) {
 		this.workspaceManager = workspaceManager
 		this.registry = registry
 		this.edit = edit
+		this.options = options
+		if (options?.clientSupportsVerisonedDocuments)
+			this.edit.documentChanges = newArrayList 
+		else
+			this.edit.changes = newLinkedHashMap
 	}
 
 	override accept(IEmfResourceChange change) {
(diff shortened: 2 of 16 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService.xtend
```diff
@@ -9,59 +9,86 @@ package org.eclipse.xtext.ide.server.rename
...
+	/**
+	 * @deprecated use {@link #rename(WorkspaceManager, RenameParams, Options, CancelIndicator)} 
+	 *    instead.
+	 */
+	@Deprecated
 	override rename(WorkspaceManager workspaceManager, RenameParams renameParams, CancelIndicator cancelIndicator) {
+		rename(workspaceManager, renameParams, new Options(false), cancelIndicator)
+	}
+	
+	override rename(WorkspaceManager workspaceManager, RenameParams renameParams, Options options, CancelIndicator cancelIndicator) {
```
dhuebner: I'm not sure if need the @since tag here too

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -607,14 +608,19 @@ import org.eclipse.xtext.findReferences.IReferenceFinder.IResourceAccess
...
-			val uri = params.textDocument.uri.toUri
+			val uri = renameParams.textDocument.uri.toUri
 			val resourceServiceProvider = uri.resourceServiceProvider
 			val renameService = resourceServiceProvider?.get(IRenameService)
 			if (renameService === null)
 				return new WorkspaceEdit
-			renameService.rename(workspaceManager, params, cancelIndicator)
+			if (renameService instanceof IRenameServiceExtension) {
+				val options = new IRenameServiceExtension.Options(params?.capabilities?.workspace?.workspaceEdit?.documentChanges === Boolean.TRUE)
+				renameService.rename(workspaceManager, renameParams, options, cancelIndicator)
```
cdietrich: i still wonder if we can get rid of workspace manager here. e.g. codeaction does not have access to it too.
i also wonder if and how we should maintain api here since lsp/lsp4j changes theirs for each release

cdietrich: @szarnekow @kthoms what do you think?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
```diff
@@ -31,21 +35,35 @@ class ChangeConverter implements IAcceptor<IEmfResourceChange> {
 
 	static class Factory {
 
-		@Inject IResourceServiceProvider.Registry registry
+		@Inject protected IResourceServiceProvider.Registry registry
 
+		/**
+		 * @deprecated use {@link #create(WorkspaceManager, WorkspaceEdit, boolean}�instead.
```
szarnekow: Looks like the link points to a method that does not exist

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
```diff
@@ -31,21 +35,35 @@ class ChangeConverter implements IAcceptor<IEmfResourceChange> {
...
 		}
 	}
 
 	val WorkspaceManager workspaceManager
 	val IResourceServiceProvider.Registry registry
 	val WorkspaceEdit edit
+	val IRenameServiceExtension.Options options
 	
-	protected new(WorkspaceManager workspaceManager, IResourceServiceProvider.Registry registry, WorkspaceEdit edit) {
+	protected new(WorkspaceManager workspaceManager, IResourceServiceProvider.Registry registry, WorkspaceEdit edit, IRenameServiceExtension.Options options) {
```
szarnekow: Could it be that clients may have specialized the ChangeConverter and the constructor still would need to callable as is.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
```diff
@@ -31,21 +35,35 @@ class ChangeConverter implements IAcceptor<IEmfResourceChange> {
...
 		this.workspaceManager = workspaceManager
 		this.registry = registry
 		this.edit = edit
+		this.options = options
+		if (options?.clientSupportsVerisonedDocuments)
```
szarnekow: typo `clientSupportsVerisonedDocuments `

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
```diff
@@ -88,12 +106,24 @@ class ChangeConverter implements IAcceptor<IEmfResourceChange> {
...
 				]
-				addTextEdit(uri, textEdits)
+				addTextEdit(uri, document, textEdits)
 			]
 		}
 	}
 	
-	protected def addTextEdit(URI uri, TextEdit... textEdit) {
-		edit.changes.put(uri.toString, textEdit)
+	protected def addTextEdit(URI theUri, Document document, TextEdit... textEdits) {
```
szarnekow: Again: is subclassing the ChangeConverter a use case?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
szarnekow: Iff there was a customized IRenameService, it would no longer be called?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService.xtend
```diff
@@ -9,59 +9,86 @@ package org.eclipse.xtext.ide.server.rename
...
 	
+	@Inject IResourceServiceProvider.Registry serviceProviderRegistry
+	
+	@Inject TokenUtil tokenUtil
+	
+	/**
+	 * @deprecated use {@link #rename(WorkspaceManager, RenameParams, Options, CancelIndicator)} 
+	 *    instead.
+	 */
+	@Deprecated
```
szarnekow: Hmm since this method is no longer called, marking it deprecated appears to be not useful.

Maybe there's a better solution compared to the extension interface? Probably default methods are an option.

Alternatively we explicitly do not care about backwards compatibility and do deliberate binary incompatible changes.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
JanKoehnlein: Yes, that's where I stopped to be API compatible cause I think this is highly unlikely.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
JanKoehnlein: @cdietrich 

@szarnekow why? The LS still injects the old interface and only calls the new method if it's present. Did I miss something?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
JanKoehnlein: Fixed

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
JanKoehnlein: Again: I don't think so.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService.xtend
JanKoehnlein: It is called when the `IRenameServiceExtension` hasn't been implemented.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService.xtend
JanKoehnlein: Fixed

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter.xtend
JanKoehnlein: Fixed

JanKoehnlein:
Thanks for the reviews. I have addressed all your comments and I am going to merge this now. 

For future API compatibility, I propose to add `option` parameters to all services and declare only these service interfaces API. It is going to be much easier to just evolve the `option`'s types API compatibly when necessary. We could also use them parameter to pass things like the `WorkspaceManager` when it's needed. We could discuss that here or on Gitter.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService.xtend
szarnekow: Specializations are likely based on the default impl and its API. Since that one implements the extension interface, none of the old methods on it are a) called or b) available anymore.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
szarnekow: My feeling is that you missed that the old impl implements the original interface but no framework code will call it anymore.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService.xtend
JanKoehnlein: Got it. Good catch.

PR #1080, opened 2019-03-17
added missing getter/setter for operationCanceledManager

cdietrich:
added missing getter/setter for operationCanceledManager


cdietrich:
@szarnekow  non of the changes were annotated. do you think they should be added to the class and other methods too?

szarnekow:
Looks good to me as is.

PR #1088, opened 2019-03-22
PR: Incorrect handling of tycho.testArgLine in the generated POMs

### LINKED ISSUE #1087: Incorrect handling of tycho.testArgLine in the generated POMs
I think the way tycho.testArgLine is used in the generated POMs is not completely correct.

In the parent POM the property is set to empty

```
<tycho.testArgLine></tycho.testArgLine>
```

and used in the general configuration of tycho-surefire in the pluginManagement section

```
<plugin>
	<groupId>org.eclipse.tycho</groupId>
	<artifactId>tycho-surefire-plugin</artifactId>
	<version>${tychoVersion}</version>
	<configuration>
		<!-- THE FOLLOWING LINE MUST NOT BE BROKEN BY AUTOFORMATTING -->
		<argLine>${tycho.testArgLine} ${platformSystemProperties} ${systemProperties} ${moduleProperties}</argLine>
	</configuration>
</plugin>
```

I seem to understand that the property is then set in the TestProjectDescriptor in case at least Java 9 is used:

https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TestProjectDescriptor.xtend#L82

To specify `--add-modules=ALL-SYSTEM` but that's useless, since in the parent POM there's already a profile that sets another property (`moduleProperties`) with `--add-modules=ALL-SYSTEM` :

https://github.com/eclipse/xtext-core/blob/cc6ab5469fc0172cef4765688722fafc59604447/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend#L565

The real problem is that `tycho.testArgLine` is typically set by Jacoco for enabling code coverage. However, if the property is set to empty (as it is now in the parent's properties section), Jacoco will have no way of setting the property in time so that it can be used when running tests, resulting in no coverage. This problem is known, but usually it is quite hard to detect when tycho.testArgLine is set to empty by default (see, e.g., https://github.com/LorenzoBettini/tycho-multiproject-jacoco-report-example/pull/2).

I'd suggest not to specify `tycho.testArgLine` at all in the parent POM. Instead, another property could be used, e.g., `additionalTestArguments`, set to empty by default and used in the tycho-surefire configuration in pluginManagement. Since Jacoco is typically configured in a separate profile, in that profile one could set

```
<additionalTestArguments>${tycho.testArgLine}</additionalTestArguments>
```

and everything works automatically with the current configuration of tycho-surefire.

I think the authors of these generated POM are @kthoms (for the parent) and @cdietrich (for the  POM of the test project, which is already useless right now).

I can provide a PR, if you think that makes sense.

LorenzoBettini:
Closes #1087

### DIFF
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/pom.xml
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/pom.xml
@@ -17,7 +17,7 @@
 		<platformSystemProperties></platformSystemProperties>
 		<moduleProperties></moduleProperties>
 		<systemProperties></systemProperties>
-		<tycho.testArgLine></tycho.testArgLine>
+		<additionalTestArguments></additionalTestArguments>
 	</properties>
 	<modules>
 		<module>org.xtext.example.full</module>
@@ -273,7 +273,7 @@
 					<version>${tycho-version}</version>
 					<configuration>
 						<!-- THE FOLLOWING LINE MUST NOT BE BROKEN BY AUTOFORMATTING -->
-						<argLine>${tycho.testArgLine} ${platformSystemProperties} ${systemProperties} ${moduleProperties}</argLine>
+						<argLine>${platformSystemProperties} ${systemProperties} ${moduleProperties} ${additionalTestArguments}</argLine>
 						<failIfNoTests>false</failIfNoTests>
 						<useUIHarness>false</useUIHarness>
 					</configuration>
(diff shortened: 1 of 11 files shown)


LorenzoBettini:
@cdietrich just for curiosity, why isn't there an automatic Jenkins build for the PR?

cdietrich:
there is and was. https://ci.eclipse.org/xtext/job/xtext-core/job/lb_issue1087/

LorenzoBettini:
but that's the job for the branch... I meant that there's no job for the PR

cdietrich:
cause <insert curses here> infrastructure is slow like hell and we cannot afford building twice

PR #1084, opened 2019-03-19
[ls] further rename fix

JanKoehnlein:
Sorry for the noise. The boolean condition to detect the offset of the element to be renamed was wrong.

PR #1091, opened 2019-03-25
[ls] fixing yet another race in rename

JanKoehnlein:
Well, as the title suggests, there was another race condition in rename.


cdietrich:
@JanKoehnlein when triggering https://ci.eclipse.org/xtext/job/xtext-snapshots/1004/consoleFull please trigger ubrella before

JanKoehnlein:
@cdietrich Oh, yes, sure. I kind of lost the overview what is built where and when.

kittaakos:
It looks good to me. I have checked all the `access.doRead` calls from the https://github.com/eclipse/xtext-core/pull/1084 PR and it all has the `get` with this PR. 👍

However, there is no need to close the `ByteArrayOutputStream` in the `finally`. See: https://github.com/eclipse/xtext-core/pull/1091/files#diff-1be23fbb80ac9198019b46aba1ba6896R84

PR #1092, opened 2019-03-26
#923 provide live resource sets to code action service

### LINKED ISSUE #923: [LSP] API of ICodeActionService has no access to WorkspaceManager

JanKoehnlein:
Fixes #923

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/ICodeActionService.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/ICodeActionService.xtend
@@ -12,15 +12,38 @@ import org.eclipse.lsp4j.CodeAction
 import org.eclipse.lsp4j.CodeActionParams
 import org.eclipse.lsp4j.Command
 import org.eclipse.lsp4j.jsonrpc.messages.Either
+import org.eclipse.xtend.lib.annotations.Accessors
 import org.eclipse.xtext.ide.server.Document
+import org.eclipse.xtext.ide.server.ILanguageServerAccess
 import org.eclipse.xtext.resource.XtextResource
 import org.eclipse.xtext.util.CancelIndicator
 
 /**
  * @author Sven Efftinge - Initial contribution and API
+ * @deprecated use {@link ICodeActionService2} instead
  */
+@Deprecated
 interface ICodeActionService {
 	
 	def List<Either<Command, CodeAction>> getCodeActions(Document document, XtextResource resource, CodeActionParams params, CancelIndicator indicator)
-	
+		
 }
+
+/**
+ * @author Jan Koehnlein
+ * @since 2.18
+ */
+ interface ICodeActionService2 {
+	
+	def List<Either<Command, CodeAction>> getCodeActions(Options options)
+		
+	@Accessors
+	class Options {
+		Document document
+		XtextResource resource
+		ILanguageServerAccess languageServerAccess
+		CodeActionParams codeActionParams
+		CancelIndicator cancelIndicator
+	}
+}
+ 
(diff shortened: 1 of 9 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/ICodeActionService.xtend
```diff
@@ -15,12 +15,26 @@ import org.eclipse.lsp4j.jsonrpc.messages.Either
...
+
+/**
+ * @author Jan Koehnlein
+ * @since 2.18
+ */
+ interface ICodeActionService2 {
 	
-}
\ No newline at end of file
+	def List<Either<Command, CodeAction>> getCodeActions(Document document, XtextResource resource, ILanguageServerAccess access, CodeActionParams params, CancelIndicator indicator)
```
kittaakos: One question; since we introduce a new service and deprecate existing APIs, wouldn't it be better to have something like:

```java
def List<Either<Command, CodeAction>> getCodeActions(CodeActionOptions options)
```

where `CodeActionOptions` is
```java
@Data
class CodeActionOptions {
  val Document document;
  val XtextResource resource;
  val ILanguageServerAccess access;
  val CodeActionParams params;
  val CancelIndicator cancelIndicator;
}
```

otherwise, we might see `ICodeActionService3` popping up in the near future. Thoughts?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/ICodeActionService.xtend
```diff
@@ -12,15 +12,38 @@ import org.eclipse.lsp4j.CodeAction
...
+ * @author Jan Koehnlein
+ * @since 2.18
+ */
+ interface ICodeActionService2 {
 	
-}
\ No newline at end of file
+	def List<Either<Command, CodeAction>> getCodeActions(Options options)
+		
+	@Accessors
```
kittaakos: Do we want to make this mutable, or did you use `@Accessor` instead of `@Data` for the easier instantiation in Xtend (`new Option() => []`)?

JanKoehnlein: The latter. These are quite a lot of params, a bit to many for a constructor IMHO.

kittaakos: OK, thanks. I am OK with both, just wanted to understand your intention.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/ICodeActionService.xtend
cdietrich: should we do this for rename too?

JanKoehnlein: Already happened https://github.com/eclipse/xtext-core/pull/1091

PR #1103, opened 2019-04-02
fixed bad comment in test plugin

cdietrich:


LorenzoBettini:
I was late also for the review, sorry ;)

Issue #1076, opened 2019-03-13
[Wizard] Duplicating Managed Version


cdietrich:
The Wizard creates for the maven lsp usecase the following into the pom of the ide project

```
<dependencies>
		<dependency>
			<groupId>log4j</groupId>
			<artifactId>log4j</artifactId>
			<version>1.2.16</version>
		</dependency>
		<dependency>
			<groupId>org.eclipse.lsp4j</groupId>
			<artifactId>org.eclipse.lsp4j</artifactId>
			<version>0.6.0</version>
		</dependency>
		<dependency>
			<groupId>org.ow2.asm</groupId>
			<artifactId>asm</artifactId>
			<version>7.0</version>
		</dependency>
		<dependency>
			<groupId>org.ow2.asm</groupId>
			<artifactId>asm-commons</artifactId>
			<version>7.0</version>
		</dependency>
		<dependency>
			<groupId>org.ow2.asm</groupId>
			<artifactId>asm-tree</artifactId>
			<version>7.0</version>
		</dependency>
	</dependencies>
```

which gives

```
Description	Resource	Path	Location	Type
Duplicating managed version 0.6.0 for org.eclipse.lsp4j	pom.xml	/org.xtext.example.mydsl.ide	line 21	Maven pom Loading Problem
Duplicating managed version 7.0 for asm	pom.xml	/org.xtext.example.mydsl.ide	line 26	Maven pom Loading Problem
Duplicating managed version 7.0 for asm-commons	pom.xml	/org.xtext.example.mydsl.ide	line 31	Maven pom Loading Problem
Duplicating managed version 7.0 for asm-tree	pom.xml	/org.xtext.example.mydsl.ide	line 36	Maven pom Loading Problem
Overriding managed version 1.2.17 for log4j	pom.xml	/org.xtext.example.mydsl.ide	line 16	Maven pom Loading Problem

```

=> we shall investigate what we can remove

kthoms:
This only occurs with wizard configuration
- Eclipse plugin enabled
- Build System Maven
- Language Server enabled

cdietrich:
fixed in 2.18

cdietrich:
@kthoms is it possible this got lost somehow?

cdietrich:
was just an old bin folder :(

PR #1109, opened 2019-04-02
[#1076] Remove BOM managed versions

kthoms:
Dependency versions are obsolete since they are managed by the BOM.


kthoms:
Updated test expectation

PR #1110, opened 2019-04-02
[eclipse/xtext#1403] Improve progress monitor utilization

kthoms:
Use a SubMonitor in applyModifications & endRecordChanges to report fine
grained progress. Further this does not rely on knowledge of the number
of passed ticks from the surrounding monitor.


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
```diff
@@ -80,6 +81,7 @@ else if (context instanceof Resource)
 			else if (context instanceof ResourceSet) {
 				throw new IllegalStateException("Not supported");
 			}
+			progress.worked(1);
```
szarnekow: Should we use progress.split(..) instead?

kthoms: Makes sense. Updated, please check again.

PR #1100, opened 2019-04-01
Occurrence highlighting of nested references

mark-christiaens:
#1090

Signed-off-by: Mark Christiaens <mark.christiaens@sigasi.com>

PR #1119, opened 2019-04-08
[testlanguages] removed superfluous entry in 'Export-Package'

sailingKieler:
follow-up of #1100


szarnekow:
Will the package be added again when the testlanguages are regenerated?

cdietrich:
no its not

Issue #888, opened 2018-10-02
[LSP] Ignore didChangeWatchedFiles for open documents


svenefftinge:
The language server should ignore file change events for documents that are opened in an editor.

svenefftinge:
Yes, it should be ignored as well.

cdietrich:
Sure? What do clients send on delete? Just make sure the server recognizes deletes

svenefftinge:
When a document is open it overlays the file version. Only when it gets closed the language server should update with the contents from the file system again.

cdietrich:
@dhuebner is this fixed now?

dhuebner:
@cdietrich Yes

cdietrich:
thx @dhuebner

PR #1122, opened 2019-04-11
[lsp] always re-process URIs from cancelled builds

### LINKED ISSUE #1121: [lsp] better handle canceled builds
When I edit files in an auto-saving editor (like Theia), `didChange` notifications are usually followed by `didChangeWatchedFiles` notifications. The latter cancel the build, but only look at their own changes to decide whether a new build should be issued. The changes from the prior `didChange` queued on the `BuildManager` are ignored.

JanKoehnlein:
Fixes #1121

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
@@ -74,7 +74,6 @@ import org.eclipse.lsp4j.services.WorkspaceService
 import org.eclipse.xtend.lib.annotations.Accessors
 import org.eclipse.xtend.lib.annotations.FinalFieldsConstructor
 import org.eclipse.xtext.findReferences.IReferenceFinder.IResourceAccess
-import org.eclipse.xtext.ide.server.BuildManager.Buildable
 import org.eclipse.xtext.ide.server.ILanguageServerAccess.IBuildListener
 import org.eclipse.xtext.ide.server.codeActions.ICodeActionService
 import org.eclipse.xtext.ide.server.codeActions.ICodeActionService2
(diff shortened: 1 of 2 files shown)


kittaakos:
Do you remember why [was the change](https://github.com/eclipse/xtext-core/commit/b25320cd3b5d26417bc3dd0c180adb97ade9af64#diff-0fc0eb4aab841adf6e1c198a00c6e95dR291), @dhuebner?

cdietrich:
i wonder if we can somehow test this

cdietrich:
@kittaakos do you mean the filtering for open documents?
that was https://github.com/eclipse/xtext-core/issues/888
but i dont understand why it was done with nested buildables.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -288,10 +287,7 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 					dirtyFiles += key
 				}
 			]
-			var Buildable buildable = []
```
cdietrich: @JanKoehnlein how does this actually differ? or is it somehow required to actually do something on the buildable?

cdietrich: (i guess the submit)

JanKoehnlein: In my case, both lists (`deletedFiles` and `dirtyFiles`) were empty, likely because the `dirtyFile` was pruned as being open, so the empty `Buildable` does nothing. As opposed to that, `workspaceManager#didChangeFiles` calls `submit` which returns the `Buildable` that actually calls `internalBuild`.

JanKoehnlein:
@kittaakos: @dhuebner 's change was to avoid duplicate builds of open files. That should still be the case, but we have to re-trigger the build if it has been canceled.

kittaakos:
@JanKoehnlein, thanks; it was clear. I was asking about the `if` statement. Seem my link above.

JanKoehnlein:
@cdietrich Don't know how to test it. To trigger the condition, we'd have to call `didChange`, then wait at least unitl the build has started, but not as long that it could finish to fire `didChangedWatchedFile`. That timing seems to be impossible to implement in a test. If you have any better ideas let me know

cdietrich:
ok. then let is at is if you have tested it manually.

JanKoehnlein:
I tested it manually for a couple of minutes. The effect (missing updates of validation markers) was no longer reproducible.

PR #1123, opened 2019-04-11
Performance improvement of array interning in NodeModelBuilder

lievenlemiengre:
More efficient method to intern the array of grammar elements during the building of the node model.

In my benchmarks parsing performance improves between 15% and 20%, the overhead of the interning becomes negligible and GC pressure is reduced (<10%).

Signed-off-by: Lieven Lemiengre lieven.lemiengre@sigasi.com

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
@@ -7,10 +7,8 @@
  *******************************************************************************/
 package org.eclipse.xtext.nodemodel.impl;
 
-import java.util.AbstractList;
-import java.util.Arrays;
 import java.util.Map;
-import java.util.RandomAccess;
+import java.util.Objects;
 
 import org.eclipse.emf.common.notify.Adapter;
 import org.eclipse.emf.ecore.EObject;
@@ -21,8 +19,8 @@ import org.eclipse.xtext.nodemodel.ILeafNode;
 import org.eclipse.xtext.nodemodel.INode;
 import org.eclipse.xtext.nodemodel.SyntaxErrorMessage;
 
-import com.google.common.collect.Interner;
 import com.google.common.collect.Maps;
+import com.google.common.collect.ObjectArrays;
 
 /**
  * A stateful (!) builder that provides call back methods for clients who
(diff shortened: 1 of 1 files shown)


cdietrich:
Will Meeres that with some bigger scenarios too

cdietrich:
@lievenlemiengre can you please sign off the commit

lievenlemiengre:
@cdietrich done :)

cdietrich:
after this change i see better cpu times but drastically worse memory behaviour. even looks like a leak. will investigate

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -32,54 +30,58 @@
...
-		@Override
-		public /* @Nullable */ T[] intern(/* @Nullable */ T[] sample) {
-			ArrayAsList<T> key = new ArrayAsList<T>(sample);
-			T[] canonical = map.get(key);
-			if (canonical == null) {
-				canonical = sample;
-				map.put(key, canonical);
-			}
-	        return canonical;
+		Map<InternKey, EObject[]> interningMap = Maps.newHashMap();
```
kthoms: This causes a leak. The ArrayInterner instance holds a strong referenced map.
You may get an inspiration of weak interning from EMF. Look at org.eclipse.emf.common.util.WeakInterningHashSet and its usage.

lievenlemiengre: No, this map is disposed together with the NodeModelBuilder after the resource was parsed.

szarnekow:
It's probably not a leak, but the ArrayInterner used to intern the arrays accross _all_ resources whereas we now have the object pool per resource.

szarnekow:
> It's probably not a leak, but the ArrayInterner used to intern the arrays accross _all_ resources whereas we now have the object pool per resource.

I probably hit return too fast. I'll double check.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -32,54 +30,58 @@
...
+					return ObjectArrays.concat(myGrammarElement, (EObject[]) childGrammarElement);
+				}
 			}
 
-			@Override public int hashCode() {
+			@Override
+			public int hashCode() {
 				if (hashCode == -1)
-					hashCode = Arrays.hashCode(array);
+					hashCode = Objects.hash(myGrammarElement, childGrammarElement);
```
szarnekow: Here we allocate an unnecessary array.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -32,54 +30,58 @@
...
 
-			@Override public boolean equals(/* @Nullable */ Object o) {
-				if (this == o)
-					return true;
-				return o instanceof ArrayAsList<?> && Arrays.equals(array, ((ArrayAsList<?>) o).array);
+			@Override
+			public boolean equals(Object o) {
+				if(o instanceof InternKey) {
+					InternKey interned = (InternKey) o;
+					return Objects.equals(this.myGrammarElement, interned.myGrammarElement)
```
szarnekow: All things considered, identity comparison should be possible here for the grammar element and the childGrammarElement

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -32,54 +30,58 @@
...
-			}
-
-			@Override /* @Nullable */ public T get(int index) {
-				return array[index];
+			
+			EObject[] joinElements() {
+				if(childGrammarElement instanceof EObject) {
+					return new EObject[] {myGrammarElement, (EObject) childGrammarElement};
+				} else {
+					return ObjectArrays.concat(myGrammarElement, (EObject[]) childGrammarElement);
```
szarnekow: This uses refection internally. I'm slightly in favor of System.arrayCopy

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
lievenlemiengre: The performance of hashCode really isn't an issue, and I'm pretty sure that hotspot has no problem folding a statically sized array, even in C1. I'm not boxing primitives either so.. I don't think this is worth it.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
lievenlemiengre: I'll change this

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
szarnekow: What I learn from @cdietrich and his project: never underestimate the cost of hashcode computation ;) 
I'm fine with keeping this as is, as long as it does not show up in any relevant measurements.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
lievenlemiengre: I couldn't measure any performance difference, I've looked at 'Array.newInstance' before and it's intrinsified (also the reason why it's often overrepresented in yourkit profiling results)

lievenlemiengre:
> What if we make the array interner a Guice singleton an injectable. We could use a ConcurrentHashMap and thereby indeed share the arrays across all resources. I wouldn't expect a memory leak due to his. I can imagine that we'll save a few more CPU cycles and also spare some memory by doing so.

I've tried this & didn't see a big difference, it could be because my testsuite doesn't properly test this usecase. I'm not convinced that the extra complexity is worth it.

This interner already has crazy hit rates, 95-98% even for relatively small files.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
szarnekow: Ok, same argumentation as below: If it doesn't show up somewhere, it's good enough :)

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -203,13 +205,8 @@ public ICompositeNode compressAndReturnParent(ICompositeNode compositeNode) {
...
 				EObject myGrammarElement = casted.getGrammarElement();
 				Object childGrammarElement = firstChild.basicGetGrammarElement();
-				EObject[] list = null;
-				if (childGrammarElement instanceof EObject) {
-					list = new EObject[] {myGrammarElement, (EObject) childGrammarElement};
-				} else {
-					list = newEObjectArray(myGrammarElement, (EObject[]) childGrammarElement);
-				}
-				casted.basicSetGrammarElement(cachedFoldedGrammarElements.intern(list));
+				EObject[] grammarElements = cachedFoldedGrammarElements.joinAndIntern(myGrammarElement, childGrammarElement);
```
cdietrich: i wonder if there i a way to restructure this (or the old version)
in a way so that it could be customized in a subclass without 
package private troubles

szarnekow: It wouldn't harm to introduce new protected methods / fields in this class.

cdietrich:
thx @lievenlemiengre

PR #1128, opened 2019-04-12
[lsp] expose some more API

JanKoehnlein:
- make it easier to extend `ProjectManager`
- mark another unused method


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.xtend
```diff
@@ -198,6 +198,10 @@ import org.eclipse.xtext.workspace.IWorkspaceConfig
 		return didChangeFiles(#[uri], newArrayList)
 	}
 
+	/**
+	 * @deprecated this method is no longer called
+	 */
+    @Deprecated
```
szarnekow: Indentation

JanKoehnlein: Fixed

JanKoehnlein:
Thanks a lot for reviewing.

PR #1130, opened 2019-04-15
[refactoring] improved progress reporting in `ChangeSerializer`

sailingKieler:
related to https://github.com/eclipse/xtext-eclipse/issues/1053

this PR improves the progress reporting in `ChangeSerializer`.
The current implementation didn't distinguish the iterations in the second loop of `applyModifications(...)`, which cause the most workload.
Besides, cancelation should be definitively checked there!

I removed the reporting from the first loop, since that simply traverses a list and add elements to another list.

I furthermore improved the progress reporting in `endRecordChanges`.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
@@ -67,9 +67,11 @@ public class ChangeSerializer implements IChangeSerializer {
 		modifications.add(Tuples.create(context, modification));
 	}
 
+	/** some additional effort required in 'endRecordChanges(IAcceptor, SubMonitor)' */
+	private static final int additionalEffortInEndRecordChanges = 5; 
+
 	@Override
 	public void applyModifications(IAcceptor<IEmfResourceChange> changeAcceptor) {
-		SubMonitor progress = SubMonitor.convert(monitor, "Preparing Text Changes...", modifications.size()*2);
 		Set<Resource> resources = Sets.newLinkedHashSet();
 		for (Pair<Notifier, IModification<? extends Notifier>> p : modifications) {
 			Notifier context = p.getFirst();
@@ -80,17 +82,34 @@ public class ChangeSerializer implements IChangeSerializer {
 			else if (context instanceof ResourceSet) {
 				throw new IllegalStateException("Not supported");
 			}
-			progress.split(1);
 		}
+		
+		// in the following effort calculation, the amount of resources is doubled for the sake of reflecting
+		//  the usually higher effort of indexing the contents of a resource compared to applying a modification
+		final int weightedPrepareAndApplyEffort = 2 * resources.size() + modifications.size();
+		
+		// here, the amount of resources is doubled for the sake of reflecting the still unknown amount of
+		//  related resources to be updated later on, too; the additional effort is a rough approximation of
+		//  the effort required to determine the (affected) related resources
+		final int weightedCreateTextChangesEffort = 2 * resources.size() + additionalEffortInEndRecordChanges;
+		
+		final SubMonitor monitorInAll = SubMonitor.convert(this.monitor,
+				weightedPrepareAndApplyEffort + weightedCreateTextChangesEffort);
+		final SubMonitor monitorPrepareAndApply = monitorInAll.split(weightedPrepareAndApplyEffort, SubMonitor.SUPPRESS_NONE);
+		monitorPrepareAndApply.beginTask("Preparing and applying file changes...", weightedPrepareAndApplyEffort);
+		
 		for (Resource res : resources) {
+			monitorPrepareAndApply.split(2);
 			// TODO: use the exact context
 			beginRecordChanges(res);
 		}
 		for (Pair<Notifier, IModification<? extends Notifier>> entry : modifications) {
+			monitorPrepareAndApply.split(1);
 			apply(entry.getFirst(), entry.getSecond());
-			progress.split(1);
 		}
-		endRecordChanges(changeAcceptor);
+		final SubMonitor monitorCreateTextChanges = monitorInAll.split(weightedCreateTextChangesEffort, SubMonitor.SUPPRESS_NONE);
+		monitorCreateTextChanges.setWorkRemaining(weightedCreateTextChangesEffort);
+		endRecordChanges(changeAcceptor, monitorCreateTextChanges);
 	}
 
 	@SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


sailingKieler:
A successful build ran [here](https://ci.eclipse.org/xtext/job/xtext-core/job/cs-eclipse-1053/).

sailingKieler:
@meysholdt @kthoms @cdietrich any objections on merging this PR? Need it to get the new test coming with https://github.com/eclipse/xtext-eclipse/pull/1055 satisfied.

cdietrich:
@sailingKieler i would prefer it if you temporarily change the target file in eclipse
to get the stuff tested there.
this is the way we call do it

sailingKieler:
> @sailingKieler i would prefer it if you temporarily change the target file in eclipse
> to get the stuff tested there.
> this is the way we call do it

So I shall commit a change in the target definition on that branch and drop it afterwards?

cdietrich:
yes this is the usual way to go.
if you do force pushes or squash before merge then the history should not be affected

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
```diff
@@ -67,9 +67,11 @@
 		modifications.add(Tuples.create(context, modification));
 	}
 
+	/** some additional effort required in 'endRecordChanges(IAcceptor, SubMonitor)' */
+	static final int additionalEffortInEndRecordChanges = 5;
```
szarnekow: nitpic: should this be a private field?

sailingKieler: You got me. Worked too much in Xtend in the past.

sailingKieler:
https://ci.eclipse.org/xtext/job/xtext-eclipse/job/cs-1053refactoring/3/ 🎉

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
```diff
@@ -80,17 +82,34 @@ else if (context instanceof Resource)
...
+		
+		// here, the amount of resources is doubled for the sake of reflecting the still unknown amount of
+		//  related resources to be updated later on, too; the additional effort is a rough approximation of
+		//  the effort required to determine the (affected) related resources
+		final int weightedCreateTextChangesEffort = 2 * resources.size() + additionalEffortInEndRecordChanges;
+		
+		final SubMonitor monitorInAll = SubMonitor.convert(this.monitor,
+				weightedPrepareAndApplyEffort + weightedCreateTextChangesEffort);
+		final SubMonitor monitorPrepareAndApply = monitorInAll.split(weightedPrepareAndApplyEffort, SubMonitor.SUPPRESS_NONE);
+		monitorPrepareAndApply.beginTask("Preparing and applying file changes...", weightedPrepareAndApplyEffort);
```
kthoms: It is quite unusual to call `beginTask` on a SubMonitor. Instead of `split()` you might convert the monitor again:
```
final SubMonitor monitorPrepareAndApply = SubMonitor.convert(monitorInAll, "Preparing and applying file changes...", weightedPrepareAndApplyEffort);
```

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
```diff
@@ -80,17 +82,34 @@ else if (context instanceof Resource)
...
 		}
+		
+		// in the following effort calculation, the amount of resources is doubled for the sake of reflecting
+		//  the usually higher effort of indexing the contents of a resource compared to applying a modification
+		final int weightedPrepareAndApplyEffort = 2 * resources.size() + modifications.size();
+		
+		// here, the amount of resources is doubled for the sake of reflecting the still unknown amount of
+		//  related resources to be updated later on, too; the additional effort is a rough approximation of
+		//  the effort required to determine the (affected) related resources
+		final int weightedCreateTextChangesEffort = 2 * resources.size() + additionalEffortInEndRecordChanges;
```
kthoms: good docs :)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
```diff
@@ -153,14 +181,16 @@ protected void endRecordChanges(IAcceptor<IEmfResourceChange> changeAcceptor) {
 				updaters.add(updater);
 			}
 		}
-		SubMonitor progress = SubMonitor.convert(monitor, "Creating Text Changes...", updaters.size());
```
kthoms: why this? I think it is better to use another SubMonitor here.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
```diff
@@ -140,9 +159,18 @@ protected RecordingResourceUpdater createResourceUpdater(Resource resource) {
...
 
 	protected void endRecordChanges(IAcceptor<IEmfResourceChange> changeAcceptor) {
+		endRecordChanges(changeAcceptor, null);
+	}
+
+	protected void endRecordChanges(IAcceptor<IEmfResourceChange> changeAcceptor, SubMonitor monitor) {
 		if (updaters.isEmpty()) {
 			return;
 		}
+		if (monitor == null) {
```
kthoms: I think you should use a dedicated SubMonitor here. Whatever gets passed in, you should convert either that or use this.monitor. But convert that to a `progress` SubMonitor. The type for `monitor` could be IProgressMonitor, no need to be more specific there.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
```diff
@@ -153,14 +181,16 @@ protected void endRecordChanges(IAcceptor<IEmfResourceChange> changeAcceptor) {
...
+		monitor.beginTask("Creating text changes...", updaters.size());
 		for (ResourceUpdater updater : updaters) {
+			monitor.split(1);
 			updater.applyChange(deltas, changeAcceptor);
-			progress.split(1);
 		}
 		for (ResourceUpdater updater : updaters) {
 			updater.unload();
 		}
+		monitor.done();
```
kthoms: When you use a dedicated SubMonitor within the method, `done` does not need to be called.

kthoms:
Obviously I was not fast enough for the review :(

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
cdietrich: @kthoms @szarnekow @sailingKieler we should think about this point again and maybe correct the type in the signature

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
sailingKieler: I prepare a test concerning the progress reporting, and I implemented it s.t. my expectations are met, but I'll have a look.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
sailingKieler: If you have a look into the impl of `split()`, you'll notice that split just "reserves"  work in the parent monitor, but doesn't report it via `worked()`. In stead it reports the work of the formerly split child monitor and its children. So in order to get the last piece of work reported, too, `done()` is called here. According to https://www.eclipse.org/articles/Article-Progress-Monitors/article.html, Section `4.0 Responsibilities of callers and callees` this is valid.

PR #1133, opened 2019-04-17
[refactoring] a bit of progress monitor rework in `ChangeSerializer`

sailingKieler:
A follow-up on #1130, replacing the calls of `SubMonitor.beginTask(...)` by `SubMonitor.convert(...)` while keeping my desired reporting behaviour.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
@@ -95,8 +95,9 @@ public class ChangeSerializer implements IChangeSerializer {
 		
 		final SubMonitor monitorInAll = SubMonitor.convert(this.monitor,
 				weightedPrepareAndApplyEffort + weightedCreateTextChangesEffort);
-		final SubMonitor monitorPrepareAndApply = monitorInAll.split(weightedPrepareAndApplyEffort, SubMonitor.SUPPRESS_NONE);
-		monitorPrepareAndApply.beginTask("Preparing and applying file changes...", weightedPrepareAndApplyEffort);
+		final SubMonitor monitorPrepareAndApply = SubMonitor.convert(
+				monitorInAll.split(weightedPrepareAndApplyEffort, SubMonitor.SUPPRESS_NONE),
+				"Preparing and applying file changes...", weightedPrepareAndApplyEffort);
 		
 		for (Resource res : resources) {
 			monitorPrepareAndApply.split(2);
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
```diff
@@ -95,8 +95,9 @@ else if (context instanceof ResourceSet) {
 		
 		final SubMonitor monitorInAll = SubMonitor.convert(this.monitor,
 				weightedPrepareAndApplyEffort + weightedCreateTextChangesEffort);
-		final SubMonitor monitorPrepareAndApply = monitorInAll.split(weightedPrepareAndApplyEffort, SubMonitor.SUPPRESS_NONE);
-		monitorPrepareAndApply.beginTask("Preparing and applying file changes...", weightedPrepareAndApplyEffort);
+		final SubMonitor monitorPrepareAndApply = SubMonitor.convert(
+				monitorInAll.split(weightedPrepareAndApplyEffort, SubMonitor.SUPPRESS_NONE),
```
kthoms: why do you split here again? should be enough to convert monitorInAll, or what am I missing?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
```diff
@@ -182,15 +185,15 @@ protected void endRecordChanges(IAcceptor<IEmfResourceChange> changeAcceptor, Su
...
 		for (ResourceUpdater updater : updaters) {
-			monitor.split(1);
+			subMonitor.split(1);
 			updater.applyChange(deltas, changeAcceptor);
 		}
 		for (ResourceUpdater updater : updaters) {
 			updater.unload();
 		}
-		monitor.done();
+		subMonitor.done();
```
kthoms: this is now obsolete. it is guaranteed that all ticks in this submon are consumed, and anyway it is not necessary to call done then.

kthoms: but I'm fine with leaving it to make this more explicit. But I'd prefer to declare a second variable for the subprogress only and call on the "main" subMonitor done here then, i.e. leave the line as is here but assign the result of convert in l.188 to another variable.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
sailingKieler: I need to split with `suppressflags` set to `SUPPRESS_NONE`, as the default setting made by `split(int)` is `SUPPRESS_BEGINTASK`, which would prevent reporting the task description `"Preparing and applying file changes..."` to the root monitor because of the following check in `SubMonitor.beginTaskImpl()`:
```
if ((flags & SUPPRESS_BEGINTASK) == 0 && name != null)
```
But I want it to be shown in the UI.

This behavior is tested by https://github.com/eclipse/xtext-eclipse/blob/387fbf08c8f34c5f50cae05229be59a8ec004949/org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/refactoring/ProgressReportingTest.xtend (xtext-eclipse)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ChangeSerializer.java
sailingKieler: > this is now obsolete. it is guaranteed that all ticks in this submon are consumed, and anyway it is not necessary to call done then.

I don't think so, as subclasses (e.g. https://github.com/eclipse/xtext-eclipse/blob/387fbf08c8f34c5f50cae05229be59a8ec004949/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/edit/ChangeSerializerTextEditComposer.java#L65) call this method via the original API `endRecordChanges(IAcceptor<IEmfResourceChange>)` with `this.monitor` potentially being set to anything.

As said in #1130, according to https://www.eclipse.org/articles/Article-Progress-Monitors/article.html, Section `4.0 Responsibilities of callers and callees` this is valid.

> But I'd prefer to declare a second variable for the subprogress only

In theory I agree. The implementation, however, returns just the provided object in that case so I skipped the additional variable.

kthoms: TLDR; : Leave as is.

> may call this method via the original API endRecordChanges(IAcceptor<IEmfResourceChange>) with this.monitor potentially being set to anything.

Right. That's the reason why you create a SubMonitor here, and you could call `done()` on that are just leave it out. You are _not_ operating on `this.monitor` here. `this.monitor` could be anything, correct, and that's why it is always a good choice to create a SubMonitor first and operate only on that in a particular method.

> In theory I agree. The implementation, however, returns just the provided object in that case so I skipped the additional variable.

Right. Since you know that it is already a SubMonitor and you know a SubMonitor converted returns the same instance, a new variable is not necessary. The assignment is also not necessary then.

sailingKieler: > The assignment is also not necessary then.

Indeed. :-)

kthoms:
Thanks, Christian!

PR #1144, opened 2019-05-08
Make DiagnosticTreeIterator return a new iterator every time

tivervac:
DiagnosticTreeIterator always returns the same iterator when you call
iterator(). This means it can't be iterated twice.

The Iterable interface does not force you to return a new Iterator every time but this is common practice. If this change doesn't get accepted because of this reason, it would be nice to at least have a comment on iterator().

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/validation/DiagnosticTreeIterableTest.java
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/validation/DiagnosticTreeIterableTest.java
@@ -1,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Sigasi N.V. (http://www.sigasi.com) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.tests/src/org/eclipse/xtext/validation/DiagnosticTreeIterableTest.java
```diff
@@ -0,0 +1,25 @@
+package org.eclipse.xtext.validation;
```
szarnekow: Missing copyright header here, too.

tivervac:
This is indeed very confusing

org.eclipse.xtext.tests/src/org/eclipse/xtext/validation/DiagnosticTreeIterableTest.java
tivervac: Done

cdietrich:
at least here it is not api

cdietrich:
@szarnekow @tivervac ping. what is the state. how shall we continue

tivervac:
I'm waiting for a decision on the "Remove Iterable" matter

szarnekow:
I'm not passionate about this corner case. I think this is ready to be merged.

PR #1148, opened 2019-05-14
Improved performance of validation for keyword heavy languages

szarnekow:


szarnekow:
> Maybe noteworthy for release notes?

I don't think so, given that only weird languages would be affected.

PR #1157, opened 2019-05-28
[lsp] fixed Document.applyEdit for multiple textEdits

### LINKED ISSUE #1156: [lsp] IOOBE in Document.applyEdit
Steps to reproduce:
- in Theia, delete multiple characters at the end of a line using backspace
- trigger undo
=> an IOOBE is thrown and swallowed in Document.applyEdit

JanKoehnlein:
Fixes #1156

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
@@ -11,6 +11,7 @@ import org.eclipse.lsp4j.Position
 import org.eclipse.lsp4j.TextEdit
 import org.eclipse.xtend.lib.annotations.Data
 import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.TextDocumentContentChangeEvent
 
 /**
  * @author Sven Efftinge - Initial contribution and API
@@ -124,6 +125,39 @@ import org.eclipse.lsp4j.Range
     		return this.contents.substring(start, end)
     }
 
+    /**
+     * As opposed to {@link TextEdit}[] the positions in the edits of a 
+     * {@link DidChangeTextDocumentParams} refer to the state after applying the preceding edits. See 
+     * https://microsoft.github.io/language-server-protocol/specification#textedit-1 and 
+     * https://github.com/microsoft/vscode/issues/23173#issuecomment-289378160 for details.
+     * 
+     * @return a new document with an incremented version and the text document changes applied.
+     * @since 2.18
+     */
+    def Document applyTextDocumentChanges(Iterable<? extends TextDocumentContentChangeEvent> changes) {
+        var currentDocument = this
+        val newVersion = if (currentDocument.version !== null)
+                currentDocument.version + 1
+            else
+                null
+        for (change : changes) {
+            val newContent = if (change.range === null) {
+                    change.text
+                } else {
+                    val start = currentDocument.getOffSet(change.range.start)
+                    val end = currentDocument.getOffSet(change.range.end)
+                    currentDocument.contents.substring(0, start) + change.text + currentDocument.contents.substring(end)
+                }
+            currentDocument = new Document(newVersion, newContent, printSourceOnError)
+        }
+        return currentDocument
+    }
+
+    /**
+     * Only use for testing.
+     * 
+     * All positions in the {@link TextEdit}s refer to the same original document (this).
+     */
     def Document applyChanges(Iterable<? extends TextEdit> changes) {
         var newContent = contents
         for (change : changes) {
@@ -137,7 +171,7 @@ import org.eclipse.lsp4j.Range
         }
         return new Document(if (version !== null) version + 1 else null, newContent)
     }
-    
+
     /**
      * @since 2.15
      */
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/DocumentTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/DocumentTest.xtend
@@ -14,6 +14,7 @@ import org.eclipse.xtext.ide.server.Document
 import org.junit.Test
 
 import static org.junit.Assert.*
+import org.eclipse.lsp4j.TextDocumentContentChangeEvent
 
 /**
  * @author efftinge - Initial contribution and API
@@ -63,7 +64,7 @@ class DocumentTest {
             assertEquals('''
                 hello world
                 bar
-            '''.normalize, applyChanges(#[
+            '''.normalize, applyTextDocumentChanges(#[
                 change(position(1,0), position(2,0), "")
             ]).contents)
         ]
@@ -79,7 +80,7 @@ class DocumentTest {
                 hello world
                 future
                 bar
-            '''.normalize, applyChanges(#[
+            '''.normalize, applyTextDocumentChanges(#[
                 change(position(1,1), position(1,3), "uture")
             ]).contents)
         ]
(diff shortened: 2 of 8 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
```diff
@@ -119,31 +119,36 @@ import org.eclipse.lsp4j.Range
...
+                } else {
+                    val start = currentDocument.getOffSet(change.range.start)
+                    val end = currentDocument.getOffSet(change.range.end)
+                    currentDocument.contents.substring(0, start) + change.newText + currentDocument.contents.substring(end)
+                }
+            val newVersion = if (currentDocument.version !== null)
+                    currentDocument.version + 1
+                else
+                    null
+            currentDocument = new Document(newVersion, newContent)
```
szarnekow: Here we are forgetting the value of printSourceOnError

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
```diff
@@ -119,31 +119,36 @@ import org.eclipse.lsp4j.Range
...
     def String getSubstring(Range range) {
-    		val start = getOffSet(range.start)
-    		val end = getOffSet(range.end)
-    		return this.contents.substring(start, end)
+          val start = getOffSet(range.start)
+          val end = getOffSet(range.end)
+          return this.contents.substring(start, end)
     }
 
     def Document applyChanges(Iterable<? extends TextEdit> changes) {
```
szarnekow: Could you please add documentation that explicitly states the relation of the offsets / ranges in the iterable of TextEdits, e.g. that the second text edit has a range that relates to the result after applying the first edit?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
```diff
@@ -119,31 +119,36 @@ import org.eclipse.lsp4j.Range
...
+        var currentDocument = this
         for (change : changes) {
-            if (change.range === null) {
-                newContent = change.newText
-            } else {
-                val start = getOffSet(change.range.start)
-                val end = getOffSet(change.range.end)
-                newContent = newContent.substring(0, start) + change.newText + newContent.substring(end)
-            }
+            val newContent = if (change.range === null) {
```
szarnekow: I wonder if this is correct, given the spec of the lsp:
```
TextEdit[]
Complex text manipulations are described with an array of TextEdit’s, representing a single change to the document.

All text edits ranges refer to positions in the original document. Text edits ranges must never overlap, that means no part of the original document must be manipulated by more than one edit. However, it is possible that multiple edits have the same start position: multiple inserts, or any number of inserts followed by a single remove or replace edit. If multiple inserts have the same position, the order in the array defines the order in which the inserted strings appear in the resulting text.
```

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
szarnekow: Also it appears the we should create only a single new version (see above)

cdietrich:
i wonder how the requests look like that theia sends

JanKoehnlein:
> I've the impression that the fix is not in line with the spec of LSP.

Maybe not with the spec. I've tried in VSCode and I get exactly the same requests and behavior as in Theia. So it seems to be Monaco, which is kind of the reference client for LSP.

JanKoehnlein:
> i wonder how the requests look like that theia sends

Theia (and VSCode) send a didChangeRequest for the undo with individual TextEdits for each character that have increasing character positions, like in the test in this commit.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
JanKoehnlein: Fixed

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
JanKoehnlein: Done

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
cdietrich: TextDocumentContentChangeEvent do seem different from TextEdits.
i find it bogus that the semantics is different here

JanKoehnlein:
I am not sure what to do here. Seems that LSP's spec and VS Code/Monaco behave differently here.
Adding @akosyakov to the conversation.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.xtend
cdietrich: => @JanKoehnlein do you know why we use the same code for text edits and contentChanges

JanKoehnlein:
BTW, thanks a lot for the fast reviews.

cdietrich:
as said if the comment i think the problem is that the specs differs between text edits and TextDocumentContentChangeEvent
but we use the same code to apply them

cdietrich:
here is what vscode sends

undo
```
{
  "jsonrpc": "2.0",
  "method": "textDocument/didChange",
  "params": {
    "textDocument": {
      "version": 5,
      "uri": "file:///data/git/xtext-languageserver-example/demo/b.mydsl"
    },
    "contentChanges": [
      {
        "range": {
          "start": {
            "line": 1,
            "character": 23
          },
          "end": {
            "line": 1,
            "character": 23
          }
        },
        "rangeLength": 0,
        "text": "x"
      },
      {
        "range": {
          "start": {
            "line": 1,
            "character": 24
          },
          "end": {
            "line": 1,
            "character": 24
          }
        },
        "rangeLength": 0,
        "text": "x"
      }
    ]
  }
}
```

typing

```
{
  "jsonrpc": "2.0",
  "method": "textDocument/didChange",
  "params": {
    "textDocument": {
      "version": 6,
      "uri": "file:///data/git/xtext-languageserver-example/demo/b.mydsl"
    },
    "contentChanges": [
      {
        "range": {
          "start": {
            "line": 1,
            "character": 25
          },
          "end": {
            "line": 1,
            "character": 25
          }
        },
        "rangeLength": 0,
        "text": "y"
      }
    ]
  }
}
```

dont know if there are text edits sent/applied on server side (with the different semantics)

JanKoehnlein:
@cdietrich I see. I'll revert and add a separate impl then for `TextDocumentContentChangeEvent`

cdietrich:
yes it seems the conversion we do e.g. at org.eclipse.xtext.ide.server.LanguageServerImpl.didChange(DidChangeTextDocumentParams)
is a bit optimistic. 
i dont know if there are other places we do it too

JanKoehnlein:
The semantics of `TextDocumentChangeEvent[]` (see https://github.com/microsoft/vscode/issues/23173#issuecomment-289378160) is different from `TextEdit[]` (see LSP spec as cited by @szarnekow).

We should not have implemented `Document#applyEdit(TextEdit[])` cause the server never uses `TextEdits`. We only do in tests. Anyway, it's API now, so I leave it and add a new method.

cdietrich:
please also mark the old method as deprecated

JanKoehnlein:
Great! Thanks a lot for your reviews and constructive comments.
Any idea why the ECA check is pending? Am I allowed to merge anyway?

cdietrich:
you have to hit the revalidate button

PR #1165, opened 2019-06-07
[eclipse/xtext#1481] Delete org/eclipse/[xtext|xtend] from local .m2

Hejado:


cdietrich:
thanks @Hejado

PR #1175, opened 2019-06-18
[eclipse/xtext#1431] Enhance container support

kthoms:
Disable Gradle daemon, since a new container for each build is
spawned and the daemon can't be reused anyway.

Remove CPU resource constraint, since default is sufficient.


Hejado:
LGTM

Issue #1178, opened 2019-06-21
[CBI] Build takes very long and tests timeout


cdietrich:
The Gradle Build on CBI takes forever (40 mins) some tests fail due timeouts
https://ci-staging.eclipse.org/xtext/job/xtext-core/job/master/
this is much slower as the CI which is already relatively slow
(compared to a 5 year old thinkpad)

am not sure if the memory is insufficient with ` -Dorg.gradle.daemon=false`
and we need to specify `-Xmx` too.
or the CPU is to few

or CBI is so crappy in general (then we should give up, we will never have any builds)

cdietrich:
cc @Hejado

cdietrich:
https://ci-staging.eclipse.org/xtext/job/xtext-lib/job/master/ is much slower as well

cdietrich:
we also dont specfiy memory for gradle anywhere e.g. via

```gradle.properties
org.gradle.jvmargs=-Xmx2G
```

cdietrich:
cc @fredg02 this is one example of problems we have with CBI

cdietrich:
maybe 1 cpu is too less: https://ci-staging.eclipse.org/xtext/job/xtext-core/job/cd_noGradleCrapOnStaging/

Hejado:
@cdietrich You might not even have 1 cpu, as the default is set 0.25, if I remember correctly. I'd be interested in a comparison of build times with the following scenarios:

* no cpu settings
* cpu setting to 0.25
* cpu setting to 1
* cpu setting to 2

Unfortunately, I cannot yet test this myself :-(

cdietrich:
- with no settings: 40 mins (test is failing / see above)
- with 0.25: 49 mins (test is failing - no idea why this is even worse than with default)
- with 0.5: 31 mins (test is still failing)
- with 1: 31 mins
- with 2: 9 mins (test is green)
=> maybe the default is even worse than 0.25 or there are other factors i dont see.
=> will restart 0.25 and 1 and see if these times are stable

cdietrich:
@Hejado 
how do the jenkinsfiles get picked up.
why is https://ci-staging.eclipse.org/xtext/job/xtext-core/job/cd_noGradleCrapOnStaging025/2/console now 11 mins?

cdietrich:
the log says

```
resources:
      limits:
        memory: "2Gi"
        cpu: "1"
      requests:
        memory: "2Gi"
        cpu: "1"
```

but the code says
[...]

PR #1181, opened 2019-06-24
[eclipse/xtext-core#1178] Use one 1 cpu in every pod

Hejado:


cdietrich:
thx @Hejado

PR #1205, opened 2019-07-15
[https://github.com/eclipse/xtext-eclipse/issues/678]

miklossy:
- Avoid infinite loop (editor freeze) on FailedPredicateException.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/internal/Lexer.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/internal/Lexer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2014, 2019 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/internal/Lexer.java
```diff
@@ -25,17 +29,53 @@ protected Lexer(CharStream input) {
...
+			/*
+			 * Avoid infinite loop (editor freeze) on {@link FailedPredicateException}
+			 */
+			catch (NoViableAltException | FailedPredicateException e) {
+				reportError(e);
+				recover(e); // throw out current char and try again
+			}
+			catch (RecognitionException re) {
+				reportError(re);
+				// match() routine has already called recover()
```
szarnekow: I do not understand the comment.

miklossy: The comment comes basically from the super implementation. The only modification I made on the super implementation is the extension of the catch block by  `| FailedPredicateException e`. So I would leave the comment as is to be able to get a readable diff between the super and the customized imlementation of the `nextToken()` method.

miklossy:
Thanks for the feedback @szarnekow . I also provided a [fix](https://github.com/eclipse/xtext-eclipse/pull/1138/commits/a7179f94c46c373f256889bcd1d004446f3efac7) for the `org.eclipse.xtext.ui.editor.contentassist.antlr.internal.Lexer`, too. Even that class is already deprecated, I think it worth fixing that as well.

PR #1207, opened 2019-07-16
[eclipse/xtext#1515] fine-tune when slack gets notified

Hejado:

### DIFF
--- a/CBI.Jenkinsfile
+++ b/CBI.Jenkinsfile
@@ -114,7 +114,7 @@ spec:
         }
       }
     }
-    changed {
+    cleanup {
       script {
         def envName = ''
         if (env.JENKINS_URL.contains('ci.eclipse.org/xtext')) {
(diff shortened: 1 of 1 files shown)


CBI.Jenkinsfile
```diff
@@ -123,19 +123,29 @@ spec:
...
-           if (currentBuild.previousBuild != null && currentBuild.previousBuild.result != 'SUCCESS') {
-             curResult = 'FIXED'
-           }
+        def curResult = currentBuild.currentResult
+        def lastResult = 'NONE'
+        if (currentBuild.previousBuild != null) {
+          lastResult = currentBuild.previousBuild.result
+        }
+        if (lastResult == 'NONE') {
+          curResult = "NEW: ${curResult}"
```
cdietrich: What will this do when failing at the first time

Hejado: This part is only executed in the first build. If the first build is red, you will get a "NEW: FAILURE".

PR #1209, opened 2019-07-17
Extend IProjectConfig and related interfaces

mmews-n4:
While implementing LSP support for our Xtext language based on xtext.ide packages, we found some issues. Some of those we would like to address in this PR, regarding IProjectConfig and ISourceFolder:
* missing output folders in IProjectConfig (actually, we only need one, but others might need more)
* excluding some files within source folders (happens e.g. when source and output folders are the same, see e.g. TypeScript)

Notes:
* I am not sure whether the new interface `IOutputFolder` is the best choice. Actually, I would prefer to rename `ISourceFolder`, but I guess this wouldn't be a good idea for xtext API continuity.
* The method `ResourceStorageFacade#computeOutputPath(StorageAwareResource)` sounds like it does something with output paths, although it looks like it is computing a relative path only. Please let me know whether I should adjust some other place to use the method `getOutputFolder` (or alike) somewhere. Right now, it is used nowhere in the xtext code base.
* Mind the related PR in repo `xtext-eclipse`.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ProjectManager.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ProjectManager.xtend
@@ -76,8 +76,8 @@ class ProjectManager {
     def Result doInitialBuild(CancelIndicator cancelIndicator) {
         val uris = newArrayList
         projectConfig.sourceFolders.forEach [
-            fileSystemScanner.scan(path) [uris += it]
-        ] 
+        	uris += it.getAllResources(fileSystemScanner)
+        ]
         return doBuild(uris, emptyList, emptyList, cancelIndicator)        
     } 
 
--- a/org.eclipse.xtext/src/org/eclipse/xtext/workspace/ISourceFolder.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/workspace/ISourceFolder.java
@@ -7,9 +7,19 @@
  *******************************************************************************/
 package org.eclipse.xtext.workspace;
 
-import org.eclipse.emf.common.util.URI;
+import java.util.ArrayList;
+import java.util.List;
 
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.xtext.util.IAcceptor;
+import org.eclipse.xtext.util.IFileSystemScanner;
+import org.eclipse.xtext.util.UriUtil;
+
+/**
+ * Describes directories that contain source files of a project.
+ */
 public interface ISourceFolder {
+
 	/**
 	 * The logical name of the source folder, like "src/main/java"
 	 */
@@ -20,4 +30,23 @@ public interface ISourceFolder {
 	 * separator. It will never be null.
 	 */
 	URI getPath();
+
+	/**
+	 * @param uri
+	 *            to check
+	 * @return true iff the given {@link URI} is a sub element of the {@link URI} of this {@link ISourceFolder}.
+	 */
+	default boolean contains(URI uri) {
+		URI path = getPath();
+		return UriUtil.isPrefixOf(path, uri);
+	}
+
+	/** @return a list of all URIs that are passed to the acceptor of {@link #scan(URI, IAcceptor)} */
+	default List<URI> getAllResources(IFileSystemScanner scanner) {
+		List<URI> uris = new ArrayList<>();
+		scanner.scan(getPath(), uri -> uris.add(uri));
+
+		return uris;
+	}
+
 }
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/SourceFolderCustomImplTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/SourceFolderCustomImplTest.xtend
@@ -1,0 +1,117 @@
+/*******************************************************************************
+ * Copyright (c) 2019 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.server
+
+import com.google.inject.AbstractModule
+import com.google.inject.Inject
+import com.google.inject.Module
+import com.google.inject.Scopes
+import org.eclipse.emf.common.util.URI
+import org.eclipse.xtext.ide.server.IWorkspaceConfigFactory
+import org.eclipse.xtext.ide.server.ProjectWorkspaceConfigFactory
+import org.eclipse.xtext.ide.server.WorkspaceManager
+import org.eclipse.xtext.util.IFileSystemScanner
+import org.eclipse.xtext.util.Modules2
+import org.eclipse.xtext.workspace.FileProjectConfig
+import org.eclipse.xtext.workspace.FileSourceFolder
+import org.eclipse.xtext.workspace.WorkspaceConfig
+import org.junit.Test
+
+import static org.junit.Assert.*
+
+/**
+ * Tests customized version of default methods ISourceFolder#contains(URI) and ISourceFolder#getAllResources(IFileSystemScanner)
+ */
+class SourceFolderCustomImplTest extends AbstractTestLangLanguageServerTest {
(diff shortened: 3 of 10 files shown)


cdietrich:
there are test failures currently

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ProjectManager.xtend
```diff
@@ -42,7 +41,6 @@ class ProjectManager {
     @Inject protected IncrementalBuilder incrementalBuilder
     @Inject protected Provider<XtextResourceSet> resourceSetProvider
     @Inject protected IResourceServiceProvider.Registry languagesRegistry
-    @Inject protected IFileSystemScanner fileSystemScanner
```
kittaakos: The removal is a breaking API change. Can we leave it there even if it is unused?

org.eclipse.xtext/src/org/eclipse/xtext/workspace/IOutputFolder.java
```diff
@@ -0,0 +1,15 @@
+/*******************************************************************************
+ * Copyright (c) 2015 itemis AG (http://www.itemis.eu) and others.
```
kittaakos: 2015 -> 2019

org.eclipse.xtext/src/org/eclipse/xtext/workspace/IOutputFolder.java
```diff
@@ -0,0 +1,15 @@
...
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.workspace;
+
+/**
+ * Describes directories that contain output files of a project. 
+ */
+public interface IOutputFolder extends ISourceFolder {
```
cdietrich: what is the idea behind a output folder beeing a source folder

cdietrich: something like javasourcelanguage?

mmews-n4: Just to reuse all methods. Like noted above, I am not that satisfied with this solution... 

Maybe instead there should be a super interface of both `ISourceFolder` and `IOutputFolder`, and this super interface would declare all methods etc (currently in `ISourceFolder`), and these sub interfaces would be empty.

org.eclipse.xtext/src/org/eclipse/xtext/workspace/ISourceFolder.java
```diff
@@ -20,4 +32,28 @@
 	 * separator. It will never be null.
 	 */
 	URI getPath();
+
+	/**
+	 * @param uri
+	 *            to check
+	 * @return true iff the given {@link URI} is a sub element of the {@link URI} of this {@link ISourceFolder}.
+	 */
+	default boolean contains(URI uri) {
```
kittaakos: They are unused. Since it is super easy to mimic this functionality with extension methods; I am not sure we need this as an API.

mmews-n4: > They

The method `contains` is used twice: `FileProjectConfig.xtend` and `EclipseProjectConfig.xtend`.

> Since it is super easy to mimic this functionality with extension methods

When possible, we use Java instead of Xtend.

mmews-n4:
>> e.g. when source and output folders are the same, see e.g. TypeScript
> I would call it a bad project structure.

- Agreed, but that's the world of some languages, which we cannot change.
- Second, consider a project file like a `package.json` or other that is located in the root directory. This project file would be part of the sources, hence, the root directory should be a source folder that contains only this project file.

>> Right now, it is used nowhere in the xtext code base.
> Out of curiosity; why do you need it then?

Because from our source files, we generate output files. Like Java, TypeScript, and many other languages do.

> I will look into this later today, but I would like to understand what kind of problem are your trying to solve with this?

We want to be more precise with input files (source files) and output files (generated by compiler, transpiler, ...). More precise means: exclude some files from source folders (by overwriting the new default methods), and define output folders, which are not concerned at the moment.

mmews-n4:
> there are test failures currently.

Thanks, I'll look into this.

kittaakos:
> Because from our source files, we generate output files. Like Java, TypeScript, and many other languages do.

There are a few projects already generating outputs with the LS, cannot you use the [`OutputConfiguration`](https://github.com/eclipse/xtext-core/blob/3cc4a3459ede6c8519f119ace63070700245bb68/org.eclipse.xtext/src/org/eclipse/xtext/generator/OutputConfiguration.java#L24)?

org.eclipse.xtext/src/org/eclipse/xtext/workspace/ISourceFolder.java
```diff
@@ -7,9 +7,21 @@
...
+import org.eclipse.xtext.util.IAcceptor;
+import org.eclipse.xtext.util.IFileSystemScanner;
+import org.eclipse.xtext.util.UriUtil;
 
+/**
+ * Describes directories that contain source files of a project.
+ */
 public interface ISourceFolder {
+
+	static final IFileSystemScanner FILE_SYSTEM_SCANNER = new IFileSystemScanner.JavaIoFileSystemScanner();
```
szarnekow: Personally I don't like this static field.

org.eclipse.xtext/src/org/eclipse/xtext/workspace/ISourceFolder.java
```diff
@@ -20,4 +32,28 @@
...
+	default List<URI> getAllResources() {
+		List<URI> uris = new LinkedList<>();
+		scan(uri -> uris.add(uri));
+
+		return uris;
+	}
+
+	/** Scans this source folder and passes all results to the acceptor. */
+	default void scan(IAcceptor<URI> acceptor) {
+		FILE_SYSTEM_SCANNER.scan(getPath(), acceptor);
```
szarnekow: Probably something like `SkippableAcceptor` with `skipped=true` which is the default. Clients can then still use another default?

org.eclipse.xtext/src/org/eclipse/xtext/workspace/ISourceFolder.java
```diff
@@ -20,4 +32,28 @@
...
+	 * @return true iff the given {@link URI} is a sub element of the {@link URI} of this {@link ISourceFolder}.
+	 */
+	default boolean contains(URI uri) {
+		URI path = getPath();
+		return UriUtil.isPrefixOf(path, uri);
+	}
+
+	/** @return a list of all URIs that are passed to the acceptor of {@link #scan(URI, IAcceptor)} */
+	default List<URI> getAllResources() {
+		List<URI> uris = new LinkedList<>();
```
szarnekow: Please use ArrayList instead. In general it is the better choice compared to LinkedList. If things are used as a queue, ArrayDeque is superior to LinkedList.

mmews-n4:
In the last two commits, I removed the output folder concept completely from the IProjectConfig. Also I think the test problems are fixed now. Please check for any other issues - Thank you.

kittaakos:
You have customized the [`ProjectManager#doInitialBuild`](https://github.com/eclipse/xtext-core/pull/1209/files#diff-a4f64d2a9ca4676ecc435c9e4cb33227R79) and [`FileProjectConfig#findSourceFolderContaining`](https://github.com/eclipse/xtext-core/pull/1209/files#diff-5fba8e11799798d80336dd0763fb6beaR65), any chance you can provide a test case as a proof it was bogus previously?

mmews-n4:
> customized

Semantically it is the same. Instead of _customized_ I'd say _refactored_.

kittaakos:
> Semantically it is the same. Instead of _customized_ I'd say _refactored_.

Awesome, thanks for the correction.

mmews-n4:
@kittaakos Thanks. Your comment before made me think, because actually there is some new functionality. Hence, my last commit adds a test class to check that new functionality.

kittaakos:
> adds a test

Nice! Can you please squash your commits? And then we are done here.

mmews-n4:
@kittaakos Done.

mmews-n4:
So, this should be it. The last change splits the test class for the purpose of improving understandability (before, there was a static boolean to switch logic, which is now removed).

Issue #922, opened 2018-10-21
Bad Positions for Diagnostics at end of line when including newline


cdietrich:
from the spec

> If you want to specify a range that contains a line including the line ending character(s) then use an end position denoting the start of the next line.

sample model

```
...
Hello

```

currently sent message
```
{
  "jsonrpc": "2.0",
  "method": "textDocument/publishDiagnostics",
  "params": {
    "uri": "file:///Users/dietrich/git/xtext-languageserver-example/demo/b.mydsl",
    "diagnostics": [
      {
        "range": {
          "start": {
            "line": 12,
            "character": 5
          },
          "end": {
            "line": 12,
            "character": 7
          }
        },
        "severity": 1,
        "code": "org.eclipse.xtext.diagnostics.Diagnostic.Syntax",
        "message": "mismatched input \u0027\u003cEOF\u003e\u0027 expecting RULE_ID"
      }
    ]
  }
}
```

according to the spec:

```
{
    start: { line: 5, character: 23 },
    end : { line 6, character : 0 }
}
```

cdietrich:
when using in [Intellij](https://github.com/gtache/intellij-lsp) this leads to

```
java.lang.StringIndexOutOfBoundsException: String index out of range: 7
	at java.lang.String.substring(String.java:1963)
	at com.github.gtache.lsp.utils.DocumentUtils$.LSPPosToOffset(DocumentUtils.scala:73)
	at com.github.gtache.lsp.editor.EditorEventManager.$anonfun$diagnostics$5(EditorEventManager.scala:527)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:945)
	at com.github.gtache.lsp.utils.ApplicationUtils$.computableReadAction(ApplicationUtils.scala:26)
	at com.github.gtache.lsp.editor.EditorEventManager.$anonfun$diagnostics$4(EditorEventManager.scala:527)
	at com.github.gtache.lsp.editor.EditorEventManager.$anonfun$diagnostics$4$adapted(EditorEventManager.scala:520)
	at scala.collection.Iterator.foreach(Iterator.scala:929)
	at scala.collection.Iterator.foreach$(Iterator.scala:929)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1417)
	at scala.collection.IterableLike.foreach(IterableLike.scala:71)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:70)
[...]

PR #1211, opened 2019-07-23
[#922] Calculate correct position (end) for issues

cdietrich:
[#922] Calculate correct position (end) for issues

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/ValidatorTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/ValidatorTest.xtend
@@ -1,0 +1,100 @@
+/*******************************************************************************
+ * Copyright (c) 2019 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.server
+
+import org.eclipse.xtext.ide.tests.testlanguage.validation.TestLanguageValidator
+import org.junit.Test
+
+import static org.junit.Assert.*
+
+import static extension org.eclipse.xtext.util.Strings.*
+import org.eclipse.xtext.diagnostics.Diagnostic
+
+/**
+ * @author Christian Dietrich - Initial contribution and API
+ */
+class ValidatorTest extends AbstractTestLangLanguageServerTest {
+
+	@Test
+	def void testMultilineDiagnostic() {
+		'MyType1.testlang'.writeFile('''
+			type Multiline_Demo {
+			}
+		'''.toUnixLineSeparator)
+		initialize
+		val problems = diagnostics.entrySet.head.value
+		assertEquals("problems found:\n" + problems.join("\n"), 1, problems.size)
+		val problem = problems.head
+		assertEquals("Test Validation to mark the whole type", problem.message)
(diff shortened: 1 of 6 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
```diff
@@ -320,13 +319,19 @@ import static org.eclipse.xtext.diagnostics.Severity.*
 		initialized.thenAccept([
 			val diagnostics = new PublishDiagnosticsParams => [
 				it.uri = toUriString(uri)
-				it.diagnostics = issues.filter[severity !== IGNORE].map[toDiagnostic].toList
+				if (issues.isEmpty) {
+					diagnostics = #[]
+				} else {
+					diagnostics = workspaceManager.doRead(uri) [document, resource|
+						issues.filter[severity !== IGNORE].map[toDiagnostic(document, it)].toList
```
cdietrich: any idea how to do this better?

szarnekow: Should we have correct methods on the issue itself, eg `getRegion():ITextRegionWithLineInformation` incl a default method that has at least the currently (broken) semantics as it was implemented here? Note that line number could either start with 0 or 1 - we may have to convert here accordingly.

cdietrich: do we have some api  that helps to build that information in an efficient manner?

cdietrich: => we would have to change IssueLocation and LocationData too (we copy over the diags 3 times)

szarnekow: > API

NodeModelUtils.getLineAndColumn(INode, int) should be very efficient.

> IssueLocation / IssueImpl

Yes, unfortunately :(

szarnekow: This cleanup would make a good first issue I suppose. So I'd be ok with the solution that is implemented here. Should we create a follow up ticket?

cdietrich: i will have a look what can be done. mayby adding an issueextension to issue instead of a default interface?

szarnekow: If that simplifies things, I'd be perfectly fine with that, too.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/ValidatorTest.xtend
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * @author Christian Dietrich - Initial contribution and API
+ */
+class ValidatorTest extends AbstractTestLangLanguageServerTest {
+
+	@Test
+	def void testMultilineDiagnostic() {
+		'MyType1.testlang'.writeFile('''
+			type Multiline_Demo {
+			}
```
szarnekow: Could you please add another test that has more than one char in the endline but different start- and end lines?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.xtend
szarnekow: There are quite a few inconsistencies / mismatches between docs and impl. The refactoring was attempted in PR #1212 but postponed.

Issue #1231, opened 2019-08-29
Partial Serialzer does not work for unassigned transient attributes


cdietrich:
We implemented a move refactoring using the `IResourceRelocationStrategy` extension point. if we move a single file this works perfectly fine. 

but if we move the declaration and the referencing file at once, then `org.eclipse.xtext.ide.serializer.impl.RelatedResourcesProvider.getRelatedResources(Collection<IResourceSnapshot>)` will remove the related resource from `result ` as both the declaration and the reference are contained in `removed` as both files were moved.

@meysholdt do you remember why we do the sorting out of added/removed.

so if we dont consider moved resources as related who / which code should take care about updating the related resources?

or should `org.eclipse.xtext.ide.serializer.impl.RelatedResourcesProvider.getRelatedResources(Collection<IResourceSnapshot>)` add the new resources to the related resources list too and not the old ones only?

**Update:** as analysis has show the root cause is: Partial Serialzer does not work for unassigned transient attributes

cdietrich:
here is a test showing a similar issue. the workaround that works in our codebase (dont do the remove) does not work there

```
@Test
	def void testMoveToNewPackage2() {
		val fs = new InMemoryURIHandler()
		fs += "inmemory:/file1.fileawaretestlanguage" -> '''
			package pkg1
			
			element Foo {
			}
		'''
		fs += "inmemory:/file2.fileawaretestlanguage" -> '''
			package pkg2
			
			element Bar {
				ref pkg1.Foo
			}
		'''

		val rs = fs.createResourceSet
		val model = rs.contents("inmemory:/file1.fileawaretestlanguage", PackageDeclaration)
		val model2 = rs.contents("inmemory:/file2.fileawaretestlanguage", PackageDeclaration)
		println(model.eResource.errors)
		println(model2.eResource.errors)

		val serializer = serializerProvider.get()
		serializer.addModification(model.eResource)[model.eResource.URI = org.eclipse.emf.common.util.URI.createURI("inmemory:/file1new.fileawaretestlanguage")]
		serializer.addModification(model, [model.name = "newpackage"])
		serializer.addModification(model2.eResource)[model2.eResource.URI = org.eclipse.emf.common.util.URI.createURI("inmemory:/file2new.fileawaretestlanguage")]
		serializer.addModification(model2, [model2.name = "newpackage"])
		Assert.assertEquals(2, model.eResource.resourceSet.resources.size)
		serializer.endRecordChangesToTextDocuments === '''
[...]

PR #1232, opened 2019-08-29
Fix Partial Serializer + Unassigened Transient issue

cdietrich:
Fix Partial Serializer + Unassigened Transient issue 
#1231


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/PartialSerializer.java
```diff
@@ -336,6 +336,10 @@ public void setErrorAcceptor(ISerializationDiagnostic.Acceptor errorAcceptor) {
 		IFeatureInfo featureInfo = constraint.getFeatures()[owner.eClass().getFeatureID(feature)];
 		List<IConstraintElement> assignments = featureInfo.getAssignments();
 		if (assignments.size() != 1) {
+			ValueTransient valueTransient = transientValues.isValueTransient(owner, feature);
```
szarnekow: What happens if there is more than one assignment? E.g. should the check for valueTransient be enhanced by `&& assignments.isEmpty()`?

cdietrich: Good question. Have to think about that

cdietrich: yes you are right. transient and zero assignments is fine. other cases really might need the sequencer to kick in

cdietrich: updated the pr

cdietrich:
@meysholdt if you find the time to review somewhen it would be nice

PR #1233, opened 2019-09-03
Intern grammar elements when nodes are created from the cache

mark-christiaens:
The node model contains nodes derived from AbstractNode.  An
AbstractNode has an array describing its grammar elements.  This array
is often identical for many AbstractNodes.  There can be millions of
nodes in a node model so it makes sense to remove the duplicate arrays.
This GrammarElementsInterner tries to reuse an interned version of
such an array wherever possible.  In particular, now we also intern
grammar elements when they are created during the load of a node model
from the node model cache.

Testing this modification on a largisch project showed that the
execution time of a clean build that accesses the node model cache
reduces by around 10%.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2019 Sigasi NV (http://www.sigasi.com) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -37,7 +37,7 @@ public class GrammarElementsInterner {
 			this.grammarElements = grammarElements;
 		}
 		
-		EObject[] prefixElement() {
+		EObject[] prependedGrammarElements() {
 			if(grammarElements instanceof EObject) {
 				return new EObject[] {grammarElement, (EObject) grammarElements};
 			} else {
@@ -45,7 +45,7 @@ public class GrammarElementsInterner {
 			}
 		}
 		
-		EObject[] postfixElement() {
+		EObject[] appendedGrammarElements() {
 			if(grammarElements instanceof EObject) {
 				return new EObject[] {(EObject) grammarElements, grammarElement };
 			} else {
@@ -73,7 +73,7 @@ public class GrammarElementsInterner {
 		}
 	}
 
-	Map<GrammarElementsInterner.InternKey, EObject[]> interningMap = Maps.newHashMap();
+	private final Map<GrammarElementsInterner.InternKey, EObject[]> interningMap = Maps.newHashMap();
 	
 	/**
 	 * Take a list of grammar elements and add an additional grammar element to the front.  If this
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
```diff
@@ -0,0 +1,100 @@
...
+				GrammarElementsInterner.InternKey interned = (GrammarElementsInterner.InternKey) o;
+				return Objects.equals(this.grammarElement, interned.grammarElement) 
+						&& Objects.equals(this.grammarElements, interned.grammarElements);
+			}
+			
+			return false;
+		}
+	}
+
+	Map<GrammarElementsInterner.InternKey, EObject[]> interningMap = Maps.newHashMap();
```
szarnekow: This should probably be a concurrent map to support parallel parsing.

szarnekow: Sorry, may fault: the interner is still used per resource.

The field should probably be protected and final, though.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
```diff
@@ -0,0 +1,100 @@
...
+		return interningMap.computeIfAbsent(internKey, key -> key.prefixElement());
+	}
+	
+	/**
+	 * Take a list of grammar elements and add an additional grammar element at the back.  If this
+	 * list has already been produced in the past, return an interned version.  Note that 
+	 * this code will only function efficiently if repeated invocation uses the same grammar element
+	 * objects and arrays or else a new concatenated array will be created every time.
+	 */
+	EObject[] appendAndIntern(Object grammarElements, EObject grammarElement) {
```
tivervac: This could be strongly typed

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
mark-christiaens: Fixed

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
```diff
@@ -0,0 +1,100 @@
+/*******************************************************************************
+ * Copyright (c) 2019 itemis AG (http://www.itemis.eu) and others.
```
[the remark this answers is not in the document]
mark-christiaens: Fixed

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
mark-christiaens: I tried a bit but the resulting code is quite asymetric.  The problem is that in some case I indeed know that the grammarElements are an EObject [] but in some other cases not.  I prefer keeping it as is.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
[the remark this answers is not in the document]
szarnekow: Do you plan to push the fixed version?

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
szarnekow: Since it's package visible and in nodemodel.impl I'm ok with the signatures, too.

Issue #1120, opened 2019-04-09
Running mvn package on Xtext parent project with Xtext 2.17.0 produces IllegalArgumentException


scanyboss:
IllegalArgumentException is thrown after running mvn package on the Xtext parent project generated by Xtext 2.17.0 programmatically. 

How to reproduce: 
1. download https://github.com/scanyboss/hello-world-language project
2. run mvn compile 
3. run Main.main() method - Code contains the path to the maven, as the mvn compile is run programmatically - needs to be changed! 
4. run mvn package on the Xtext parent project generated by the main method. 
The last step produces the error.

cdietrich:
hmmmm in my try the last one does not fail.
can you please checkin the result of (3) too

scanyboss:
I'm using Fedora 27. 
I have no problem with step 3. 
![Screenshot from 2019-04-09 21-14-33](https://user-images.githubusercontent.com/17799565/55828544-f46c8a80-5b0c-11e9-98f5-01fb7c216a00.png)

But when I was running maven action programmatically for the first time I had to add execute permissions to the maven itself because even when the maven action doesnt fail, "permission denied" can be printed in the logs. Dont you see it when running step 3 ?

cdietrich:
no, i have the problem: all 4 steps run smoothly

```
INFO] Reactor Summary for org.hello.parent 1.0.0-SNAPSHOT:
[INFO] 
[INFO] org.hello.parent ................................... SUCCESS [  0.007 s]
[INFO] org.hello .......................................... SUCCESS [ 20.122 s]
[INFO] org.hello.ide ...................................... SUCCESS [  2.589 s]
[INFO] org.hello.ui ....................................... SUCCESS [  3.158 s]
[INFO] org.hello.target ................................... SUCCESS [  0.058 s]
[INFO] ------------------------------------------------------------------------
```
=> to verify your failure at (4) i want to start at the end of (3)

scanyboss:
Are you running it directly in Eclipse IDE ?

cdietrich:
no on bash

scanyboss:
[...]

PR #1242, opened 2019-09-16
#1120 Fix merger in case of line breaks inside identifiers.

ArneDeutsch:
Remove superflues cancelation of creation process in case of missing
version.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
@@ -420,8 +420,6 @@ public class MergeableManifest2 implements Cloneable {
 	}
 
 	private void writeHeader(BufferedWriter writer) throws IOException {
-		if (version == null)
-			return;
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.util/src/org/eclipse/xtext/util/MergeableManifest2.java
```diff
@@ -762,33 +760,35 @@ private Bundle(String input) {
...
+			return Objects.equals(getName(), other.getName());
 		}
 
 		public String getName() {
-			// trim because prefix (such as newline) is encoded in the name
-			return split.get(0).trim();
+			// trim because prefix (such as newline) is encoded in the name,
+			// also remove all newlines to ensure linebreaks in name does not
+			// cause issues
+			return split.get(0).trim().replaceAll("\r\n ", "").replaceAll("\n ", "");
```
cdietrich: what about `\r?\n ` for better readability?

cdietrich: ping @ArneDeutsch

PR #1251, opened 2019-10-02
[util] Fixing the behavior of the cleanFolder function in Files class.

gallandarakhneorg:
The previous behavior does not remove the parent folder when its has no
file inside and the "deleteParentFolder" argument is true.
The behavior of the function is updated to delete the parent folder as
expected.
Additionally, the recursive algorithm is replaced by an iterative
algorithm in order to make it more robust.

Signed-off-by: Stéphane Galland <galland@arakhne.org>


cdietrich:
@szarnekow what do you think?

gallandarakhneorg:
Unit tests on the updated function were aded.

cdietrich:
thx @gallandarakhneorg

Issue #1252, opened 2019-10-03
Performance improvement for StateBasedContainer


tisis2:
In Workspaces with a huge index the implementation of getExportedObjectsByType gets slow, since its first filtering by type and then the uri filter is applied on the result. We see small Projects get affected by big workspaces (content assist, reparse,...)

tisis2:
When profiling we saw a lot of pressure on the EObjectDescriptionLookUp.getExportedObjectsByType in a wokspace with a huge index.

I created an implementaion of IResourceDescription that filters the resource descriptions to the actual visible one in the current container state. On this way the filtering if the resource is acutally in the current state happens before the filtering of all exported Objects by their type.

tisis2:
since the pull request #1253 does not solve this problem anymore, what should we do with this issue?

cdietrich:
would it keep open for now but remove it from any milestone

cdietrich:
=> with https://github.com/eclipse/xtext-core/pull/1253 we will allow to customize this easier but we will leave this issue open for now

szarnekow:
Please re-open in https://github.com/eclipse/xtext if the problem is still present.

PR #1253, opened 2019-10-03
[eclipse/xtext-core#1252] filter for contained resource descriptions

tisis2:
before any other filter operation

there is a performance benefit, when first exclude all irrelevant
resources before considering their IEObjectDescriptions

Signed-off-by: gabrield <d.gabriel@bachmann.info>

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
@@ -12,7 +12,6 @@ import static java.util.Collections.*;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
-import java.util.function.Function;
 
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EClass;
@@ -46,24 +45,19 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 	
 	@Override
 	protected Iterable<IEObjectDescription> filterByURI(Iterable<IEObjectDescription> unfiltered) {
-		Predicate<IEObjectDescription> predicate = getStateContainsPredicate(input -> input.getEObjectURI().trimFragment());
-		return Iterables.filter(unfiltered, predicate);
-	}
-	
-	protected <T> Predicate<T> getStateContainsPredicate(Function<T, URI> uriProvider) {
-		return new Predicate<T>() {
+		return Iterables.filter(unfiltered, new Predicate<IEObjectDescription>() {
 			private Collection<URI> contents = null;
 
 			@Override
-			public boolean apply(T input) {
+			public boolean apply(IEObjectDescription input) {
 				if(contents == null) {
 					contents = state.getContents();
 				}
-				URI resourceURI = uriProvider.apply(input);
+				URI resourceURI = input.getEObjectURI().trimFragment();
 				final boolean contains = contents.contains(resourceURI);
 				return contains;
 			}
-		};
+		});
 	}
 
 	@Override
@@ -118,9 +112,7 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 		if (isEmpty())
 			return emptyList();
 
-		Predicate<IResourceDescription> isResourceDescritptionInState = getStateContainsPredicate(input -> input.getURI());
-		Iterable<IResourceDescription> resourceDescriptionsInState = Iterables.filter(getDescriptions().getAllResourceDescriptions(), isResourceDescritptionInState);
-		return IterableExtensions.flatMap(resourceDescriptionsInState, desc -> desc.getExportedObjectsByType(type));
+		return IterableExtensions.flatMap(getResourceDescriptions(), desc -> desc.getExportedObjectsByType(type));
 	}
 	
 	@Override
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
@@ -12,6 +12,7 @@ import static java.util.Collections.*;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.function.Function;
 
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EClass;
@@ -21,6 +22,7 @@ import org.eclipse.xtext.resource.IEObjectDescription;
 import org.eclipse.xtext.resource.IResourceDescription;
 import org.eclipse.xtext.resource.IResourceDescriptions;
 import org.eclipse.xtext.resource.impl.ResourceDescriptionsBasedContainer;
+import org.eclipse.xtext.xbase.lib.IterableExtensions;
 
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
@@ -44,19 +46,24 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 	
 	@Override
 	protected Iterable<IEObjectDescription> filterByURI(Iterable<IEObjectDescription> unfiltered) {
-		return Iterables.filter(unfiltered, new Predicate<IEObjectDescription>() {
+		Predicate<IEObjectDescription> predicate = getStateContainsPredicate(input -> input.getEObjectURI().trimFragment());
+		return Iterables.filter(unfiltered, predicate);
+	}
+	
+	protected <T> Predicate<T> getStateContainsPredicate(Function<T, URI> uriProvider) {
+		return new Predicate<T>() {
 			private Collection<URI> contents = null;
 
 			@Override
-			public boolean apply(IEObjectDescription input) {
+			public boolean apply(T input) {
 				if(contents == null) {
 					contents = state.getContents();
 				}
-				URI resourceURI = input.getEObjectURI().trimFragment();
+				URI resourceURI = uriProvider.apply(input);
 				final boolean contains = contents.contains(resourceURI);
 				return contains;
 			}
-		});
+		};
 	}
 
 	@Override
@@ -110,7 +117,10 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 	public Iterable<IEObjectDescription> getExportedObjectsByType(EClass type) {
 		if (isEmpty())
 			return emptyList();
-		return super.getExportedObjectsByType(type);
+
+		Predicate<IResourceDescription> isResourceDescritptionInState = getStateContainsPredicate(input -> input.getURI());
+		Iterable<IResourceDescription> resourceDescriptionsInState = Iterables.filter(getDescriptions().getAllResourceDescriptions(), isResourceDescritptionInState);
+		return IterableExtensions.flatMap(resourceDescriptionsInState, desc -> desc.getExportedObjectsByType(type));
 	}
 	
 	@Override
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
@@ -42,6 +42,10 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 		super(descriptions);
 		this.state = state;
 	}
+
+	protected IContainerState getState() {
+		return state;
+	}
 	
 	@Override
 	protected Iterable<IEObjectDescription> filterByURI(Iterable<IEObjectDescription> unfiltered) {
@@ -51,7 +55,7 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 			@Override
 			public boolean apply(IEObjectDescription input) {
 				if(contents == null) {
-					contents = state.getContents();
+					contents = getState().getContents();
 				}
 				URI resourceURI = input.getEObjectURI().trimFragment();
 				final boolean contains = contents.contains(resourceURI);
@@ -62,22 +66,22 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 
 	@Override
 	public boolean hasResourceDescription(URI uri) {
-		return state.contains(uri);
+		return getState().contains(uri);
 	}
 	
 	@Override
 	public int getResourceDescriptionCount() {
-		return state.getContents().size();
+		return getState().getContents().size();
 	}
 
 	@Override
 	public boolean isEmpty() {
-		return state.isEmpty();
+		return getState().isEmpty();
 	}
 	
 	@Override
 	public IResourceDescription getResourceDescription(URI uri) {
-		if (state.contains(uri))
+		if (getState().contains(uri))
 			return getDescriptions().getResourceDescription(uri);
 		return null;
 	}
@@ -92,7 +96,7 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 	@Override
 	protected Map<URI, IResourceDescription> doGetUriToDescription() {
 		Map<URI, IResourceDescription> result = Maps.newLinkedHashMap();
-		for(URI uri: state.getContents()) {
+		for(URI uri: getState().getContents()) {
 			IResourceDescription description = getDescriptions().getResourceDescription(uri);
 			if (description != null)
 				result.put(uri, description);
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
@@ -43,7 +43,7 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 		this.state = state;
 	}
 
-	protected IContainerState getState() {
+	protected final IContainerState getState() {
 		return state;
 	}
 	
@@ -115,8 +115,7 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 	public Iterable<IEObjectDescription> getExportedObjectsByType(EClass type) {
 		if (isEmpty())
 			return emptyList();
-
-		return IterableExtensions.flatMap(getResourceDescriptions(), desc -> desc.getExportedObjectsByType(type));
+		return super.getExportedObjectsByType(type);
 	}
 	
 	@Override
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
@@ -43,6 +43,9 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 		this.state = state;
 	}
 
+	/**
+	 * @since 2.20
+	 */
 	protected final IContainerState getState() {
 		return state;
 	}
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainerManager.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainerManager.java
@@ -85,15 +85,16 @@ public class StateBasedContainerManager implements IContainer.Manager {
 		IAllContainersState state = getState(resourceDescriptions);
 		IContainerState containerState = new ContainerState(handle, state);
 		
-		StateBasedContainer result = createStateBasedContainer(resourceDescriptions, containerState);
-		if (state instanceof FlatResourceSetBasedAllContainersState)
+		return createContainer(resourceDescriptions, state, containerState);
+	}
+
+	protected IContainer createContainer(IResourceDescriptions resourceDescriptions, IAllContainersState allContainerState, IContainerState containerState) {
+		StateBasedContainer result = new StateBasedContainer(resourceDescriptions, containerState);
+		
+		if (allContainerState instanceof FlatResourceSetBasedAllContainersState)
 			result.setUriToDescriptionCacheEnabled(false);
 		
 		return result;
-	}
-
-	protected StateBasedContainer createStateBasedContainer(IResourceDescriptions resourceDescriptions, IContainerState containerState) {
-		return new StateBasedContainer(resourceDescriptions, containerState);
 	}
 
 	protected List<IContainer> getVisibleContainers(List<String> handles, IResourceDescriptions resourceDescriptions) {
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainerManager.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainerManager.java
@@ -84,10 +84,16 @@ public class StateBasedContainerManager implements IContainer.Manager {
 	protected IContainer createContainer(String handle, IResourceDescriptions resourceDescriptions) {
 		IAllContainersState state = getState(resourceDescriptions);
 		IContainerState containerState = new ContainerState(handle, state);
-		StateBasedContainer result = new StateBasedContainer(resourceDescriptions, containerState);
+		
+		StateBasedContainer result = createStateBasedContainer(resourceDescriptions, containerState);
 		if (state instanceof FlatResourceSetBasedAllContainersState)
 			result.setUriToDescriptionCacheEnabled(false);
+		
 		return result;
+	}
+
+	protected StateBasedContainer createStateBasedContainer(IResourceDescriptions resourceDescriptions, IContainerState containerState) {
+		return new StateBasedContainer(resourceDescriptions, containerState);
 	}
 
 	protected List<IContainer> getVisibleContainers(List<String> handles, IResourceDescriptions resourceDescriptions) {
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainerManager.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainerManager.java
@@ -88,6 +88,9 @@ public class StateBasedContainerManager implements IContainer.Manager {
 		return createContainer(resourceDescriptions, state, containerState);
 	}
 
+	/**
+	 * @since 2.20
+	 */
 	protected IContainer createContainer(IResourceDescriptions resourceDescriptions, IAllContainersState allContainerState, IContainerState containerState) {
 		StateBasedContainer result = new StateBasedContainer(resourceDescriptions, containerState);
 		


tisis2:
found out its causing some issues with the actual solution. in the ResourceSetBasedAllContainersStateProvider are instanceof checks and this will now throw an exception since the resource descriptions class the provider receives is now a wrapper

2019-10-04 07_57_30-PLC - org eclipse xtext resource containers ResourceSetBasedAllContainersStatePr

cdietrich:
I saw this comment in my mails

found out its causing some issues with the actual solution. in the ResourceSetBasedAllContainersStateProvider are instanceof checks and this will now throw an exception since the resource descriptions class the provider receives is now a wrapper

Does this no longer hold true?

tisis2:
as far as i have seen this happenes in our productive code since i did the change slightly different there as a workaround. in that workaround i found the issue and assumed its a general issue... but after i commented i've realized its different in that change. im sorry for that confusion

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -38,26 +35,9 @@
 	private final IContainerState state;
 
 	public StateBasedContainer(IResourceDescriptions descriptions, IContainerState state) {
-		super(descriptions);
+		super(new StateBasedResourceDescriptions(descriptions, state));
```
szarnekow: Wrapping the given instance of `IResourceDescriptions` may cause trouble with implementors that check for a specific instance type (as you mentioned in a no longer available comment). Do you have another idea? Maybe using `descriptions.getAllResourceDescriptions().filter(..)` and working with that result from within this impl? Or we could add a default method to the interface `IResourceDescriptions` that allows to implement an optimized query iff the result is supposed to be filtered by URI.

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedResourceDescriptions.java
```diff
@@ -0,0 +1,87 @@
...
+	 * @param state
+	 *            the {@link IContainerState} describing the containing URIs
+	 */
+	public StateBasedResourceDescriptions(IResourceDescriptions delegate, IContainerState state) {
+		this.delegate = delegate;
+		this.state = state;
+	}
+
+	private boolean isInState(URI uri) {
+		return state.getContents().contains(uri);
```
szarnekow: Should be  `state.contains(..)` since `getContents()` may be expensive.

tisis2:
i uploaded an alternative solution to get rid of the wrapper. can u take a look at that version. im currently not sure if the filter should also be applied at other places

org.eclipse.xtext/src/org/eclipse/xtext/resource/impl/ResourceDescriptionsBasedContainer.java
```diff
@@ -62,19 +63,25 @@ public IResourceDescription getResourceDescription(URI uri) {
 	
 	@Override
 	public Iterable<IEObjectDescription> getExportedObjects(EClass type, QualifiedName qualifiedName, boolean ignoreCase) {
-		Iterable<IEObjectDescription> unfiltered = getDescriptions().getExportedObjects(type, qualifiedName, ignoreCase);
+		Iterable<IResourceDescription> resourceDescriptions = filterByURI(getDescriptions());
+		Iterable<IEObjectDescription> unfiltered = IterableExtensions.flatMap(resourceDescriptions, desc -> desc.getExportedObjects(type, qualifiedName, ignoreCase));
```
szarnekow: `resourceDescriptions.getExportedObjects(.., name, ..)` is usually very fast. Here I'd opt for filtering after-the-fact, e.g. based on the few found `IEObjectDescriptions`.

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -43,20 +41,8 @@ public StateBasedContainer(IResourceDescriptions descriptions, IContainerState s
...
-			public boolean apply(IEObjectDescription input) {
-				if(contents == null) {
-					contents = state.getContents();
-				}
-				URI resourceURI = input.getEObjectURI().trimFragment();
-				final boolean contains = contents.contains(resourceURI);
-				return contains;
-			}
-		});
+	protected Iterable<IResourceDescription> filterByURI(IResourceDescriptions unfiltered) {
```
szarnekow: Therefore I'd think an overridden `getExportedObjectsByType` is more straight forward instead of the template method `filterByURI`

tisis2: i was templating it so other implementers of ResourceDescritpionsBasedContainer can make use of that performance improvement too.

org.eclipse.xtext/src/org/eclipse/xtext/resource/impl/ResourceDescriptionsBasedContainer.java
tisis2: you are right that this is usually very fast. i just thought it might get confusing if this method would not apply the filter. so if a API user just overrides the resourceDescription filter he might get a result here that would not apply that filter

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -44,19 +46,24 @@ public StateBasedContainer(IResourceDescriptions descriptions, IContainerState s
...
+	}
+	
+	private <T> Predicate<T> getStateContainsPredicate(Function<T, URI> uriProvider) {
+		return new Predicate<T>() {
 			private Collection<URI> contents = null;
 
 			@Override
-			public boolean apply(IEObjectDescription input) {
+			public boolean apply(T input) {
 				if(contents == null) {
```
szarnekow: Looks good to me, though maybe `state.contains(..)` is the better option here, while we are on it. @cdietrich what's your take on this?

cdietrich: the problem is that i have no idea how to test this. i dont have a usecase for that. we ususally dont have 100s of projects so the change is irrelevent

szarnekow: 2 projects / jars with 10000 files in the first and 1 file in the second should allow to observe an improvement when looking for objects by type in the second container.

cdietrich:
i will try to test this next week but am still unsure as we already have extended `EObjectDescriptionLookup` with a type to objects lookup map

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -12,6 +12,7 @@
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.function.Function;
```
cdietrich: @szarnekow do we have any consistent concept when to use
`java.util.function.Function` or  `org.eclipse.xtext.xbase.lib.Functions.Function1`
or `com.google.common.base.Function`

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -44,19 +46,24 @@ public StateBasedContainer(IResourceDescriptions descriptions, IContainerState s
 	
 	@Override
 	protected Iterable<IEObjectDescription> filterByURI(Iterable<IEObjectDescription> unfiltered) {
-		return Iterables.filter(unfiltered, new Predicate<IEObjectDescription>() {
+		Predicate<IEObjectDescription> predicate = getStateContainsPredicate(input -> input.getEObjectURI().trimFragment());
+		return Iterables.filter(unfiltered, predicate);
+	}
+	
+	private <T> Predicate<T> getStateContainsPredicate(Function<T, URI> uriProvider) {
```
cdietrich: am not sure if protected would be better

cdietrich: https://github.com/search?q=%22extends+StateBasedContainer+%7B%22&type=Code

tisis2:
> i will try to test this next week but am still unsure as we already have extended `EObjectDescriptionLookup` with a type to objects lookup map.
> so we might not really profit from this

can you maybe help me finding that type to object look up you mentioned. i was not able to find that in EObjectDescriptionLookup. i would be interested in that solution.

cdietrich:
it looks something like

```
public final class EObjectDescriptionLookup2 extends EObjectDescriptionLookUp {
    
    // NOTE: must be a linked hash multimap --> we need to keep the order correct, otherwise comparison of
    // IEObjectDescription lists is broken
    private volatile LinkedHashMultimap<EClass, IEObjectDescription> clazzToObjects;

    private volatile List<IEObjectDescription> allDescriptions;

    /**
     * @see EObjectDescriptionLookUp#EObjectDescriptionLookUp(List)
     */
    public EObjectDescriptionLookup2(List<IEObjectDescription> allDescriptions) {
        super(allDescriptions);
    }

    @Override
    public void setExportedObjects(java.util.List<IEObjectDescription> allDescriptions) {
        super.setExportedObjects(allDescriptions);
        synchronized (this) {
            this.allDescriptions = allDescriptions;
            this.clazzToObjects = null;
        }
    }

    private LinkedHashMultimap<EClass, IEObjectDescription> getClazzToObjects() {
        if (clazzToObjects == null) {
            synchronized (this) {
                if (clazzToObjects == null) {
                    LinkedHashMultimap<EClass, IEObjectDescription> clazzToObjectsTemp = LinkedHashMultimap.<EClass, IEObjectDescription>create(200, allDescriptions.size()/200);
                    if (allDescriptions instanceof RandomAccess) {
                        for(int i = 0; i < allDescriptions.size(); i++) {
                            IEObjectDescription description = allDescriptions.get(i);
                            putIntoClazzToObjectsMap(clazzToObjectsTemp, description);
                        }
                    } else {
                        for(IEObjectDescription description: allDescriptions) {
                            putIntoClazzToObjectsMap(clazzToObjectsTemp, description);
                        }
                    }
                    this.clazzToObjects = clazzToObjectsTemp;
                }
            }
        }
        return this.clazzToObjects;
    }

    private static void putIntoClazzToObjectsMap(LinkedHashMultimap<EClass, IEObjectDescription> clazzToObjects,
            IEObjectDescription description) {
        EClass indexKey = description.getEClass();
        clazzToObjects.put(indexKey, description);
        Collection<EClass> superTypes = EcoreUtil2.getAllSuperTypes(indexKey);
        for (EClass superType : superTypes) {
            clazzToObjects.put(superType, description);
        }
    }

    @Override
    public Iterable<IEObjectDescription> getExportedObjectsByType(EClass type) {
        Iterable<IEObjectDescription> values = getClazzToObjects().get(type);
        if (values == null) {
            return Collections.emptyList();
        }
        return values;
    }
}

```

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
tisis2: > i will try to test this next week but am still unsure as we already have extended `EObjectDescriptionLookup` with a type to objects lookup map.
> so we might not really profit from this

we recognized that performance issue on successive pastes in a source file, if there are enough other entries in your index.
you might still recognize the improvement at least in the number of calls on getExportedObjectsByType you get in your EObjectDescriptionLookUp.

cdietrich:
@tisis2 could you please squash your commits into one and force push your branch. thanks

cdietrich:
@tisis2 @szarnekow i did some measurements with this change in our codebase and there is a notable performance drop of 10s (on 640s) 

here is a part of that

some-of-the-bad-performance

thus i am very sceptical about this change

tisis2:
@cdietrich can u explain how you tested that?

i profiled the case where the issue occured (on pasting in an xtext editor). heres the hot spots view i got out of jprofiler before my change:
2019-10-15 13_54_17-Local attach - JProfiler 8 1 4

and after my change:
2019-10-15 13_58_08-Local attach - JProfiler 8 1 4

cdietrich:
i have 2 projects with very few and one project with 1000s of dsl files.
we have the optimized lookup obove.
this gives a a huge reduction and fast access when asking for the type.
so the addiotional filtering wont hurt.
but doing the contains does.

tisis2:
what was that interaction with that projects? that action that took 640s?

cdietrich:
this is the clean build

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
tisis2: finally i did a little test for state.contains vs. contents.contains. on a cleanbuild of my test workspace the build took 220s with state.contains vs 210s with the cached contents.contains.

so it seems there is a benefit of getting the contents and to the check on them

tisis2:
i did also some tests now with a workspace with 18 projects of different sizes. overall we have ~17000 source files. i also measured the times of the clean build of the whole workspace. i tried to compare different source states here.

the xtext version in our target platform without any modification builds that workspace in ~260s.
in the profiler the isSuperTypeOf appeared as a hotspot with ~28s

with that change the build time could be reduced to 210s (isSuperTypeOf appeared with ~2s)

next thing i wanted to try is to remove my change completely and try the change of EObjectDescriptionLookUp2 (i just integrated it into the original EObjectDescriptionLookUp)
with that change the build took ~235s
since here the isSuperTypeOf would not appear anymore i was intrested how much overhead the getExportedObjectsByType would bring and it appeared as hot spot with ~8s

next i just tried to combine my change and that EObjectDescriptionLookUp change and i got the best of all results with ~195s. here the getExportedObjectsByType didn't appear as hot spot at all anymore

so i couldn't see a drop of performance on clean build here. neither with nor without the EObjectDescriptionLookUp change.

i have seen that the EObjectDescriptionLookUp change would also be an improvement for us and may also be an improvement for others. is there a reason this never made it into xtext itself?

cdietrich:
it might have a negative memory footprint

tisis2:
im trying to find the reason why you get a drop of performance in your scenario. Can you maybe provide some informations about the structure your projects have?
how many exported objects a resource description has in average? how many different type they are? Are there maybe many resource that would not have exported objects at all?

cdietrich:
i dont have any number but there is a huge number of files and a not so huge number of exported object, but each file exports some
and the quota should be "balanced"

tisis2:
at the moment i got feeling that a drop of performance here only makes sense if you got less exported objects of one type than you have resource descriptions in your scenario, but im not totally sure... im trying to proof that.

cdietrich:
yes. the EObjectDescriptionLookup2 drops the number of descriptions to look at by a lot.
thus doing this first gives a huge improvement.

tisis2:
i dont think the time needed by the LookUp would make any difference, since this is a constant that didn't change and i also tried the EObjectDescriptionLookup2 and saw that it still improved performance with my scenario.

so if you see a drop of performance with EObjectDescriptionLookup2, i think you should also see it when not using it

cdietrich:
yes i think it will be even worse. but since i dont have many project but
1 with 99,9% of the reources and 2 with 0,05% each
then the overhead for the extra contains will be measureable. (the contains call)

tisis2:
if you export an average of 1 or objects per resource of the type you are asking for the implementation should already be faster, cause actually at that point your are calling contains less often than before. thats why i think you are exporting less than 1 object of the type you are asking for per resource in average and this might be a reason why the contains call is now called more often. 
im not sure at the moment if i am able to create such a scenario with our product

riederm:
Hey guys, I'm working with @tisis2 here and I helped him a bit with this change. I'm asking myself if this discussion still represents the "usual situations" of xtext projects.

The general assumption: "First establish what we can see and then filter the remaining IEObjectDescriptions for a specific type" - sounds reasonable for me. I also understand that @cdietrich has somehow a special situation where you changed the implementation of EObjectDescriptionLookup to better support your scenario. 

Isn't it an option for you to exchange the StateBasedContainer as well back to a behavior that "knows" about the special properties of your Lookup. For the general situation I think the proposed enhancement of @tisis2 makes sense. 
The current discussion is hard since we talk about considering a special implementation that did not make it into xtext mainline for some reasons. So most other people propably don't see the problem.

cdietrich:
the main problem is that theres much package private/protected stuff in that area so in doubt its not possible to easily change back.

i wíll need some time to discuss that with @szarnekow 
i also dont know if he or others (looking at you @tivervac) have evaluated this change in their codebase

tisis2:
the package visibility was the reason why it was initially done by the IResourceDescriptions wrapper since we also were not able to override the StateBasedContainer wihout copy a lot.

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -110,7 +117,10 @@ public IResourceDescription getResourceDescription(URI uri) {
 	public Iterable<IEObjectDescription> getExportedObjectsByType(EClass type) {
 		if (isEmpty())
 			return emptyList();
-		return super.getExportedObjectsByType(type);
+
+		Predicate<IResourceDescription> isResourceDescritptionInState = getStateContainsPredicate(input -> input.getURI());
+		Iterable<IResourceDescription> resourceDescriptionsInState = Iterables.filter(getDescriptions().getAllResourceDescriptions(), isResourceDescritptionInState);
```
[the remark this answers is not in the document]
cdietrich: this change wont use isResourceDescritptionInState (typo btw)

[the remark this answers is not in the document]
cdietrich: its slightly better but still much worse than with the current version.
if we move

https://github.com/eclipse/xtext-core/blob/c2b20283fbd2057d9c4ae1b90ec907c813c7372e/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainerManager.java#L87

to a factory method and expose
org.eclipse.xtext.resource.containers.StateBasedContainer.state
e.g. via a protected getter
then it could work.

even better woud be an addtional flag in the constructor that switches the behaviour.
but thats an ugly solution too.

tisis2:
i did the changes @cdietrich suggested, so you could try if this works for you
also removed filtering by contains. just saw now that there was already a getResourceDescriptions which will fetch the descriptions the slightly faster way

cdietrich:
@tivervac as discussed at eclipsecon it would be nice if you could give this a try with your codebase/testcases

tivervac:
On 4473 of our small (runtime) tests this patch performs ~8% slower, 26.3s without, 28.4s with the patch
On our eclipse project tests it performs 2% better, 132s without and 129s with the patch

I'll see if I can do some better profiling on monday

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
szarnekow: :+1:

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
szarnekow: Personally I have a preference towards `j.u.f.Function` whenever possible.

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
szarnekow: > protected would be better

:+1:

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainerManager.java
```diff
@@ -84,12 +84,18 @@ protected IAllContainersState getState(IResourceDescriptions resourceDescription
 	protected IContainer createContainer(String handle, IResourceDescriptions resourceDescriptions) {
 		IAllContainersState state = getState(resourceDescriptions);
 		IContainerState containerState = new ContainerState(handle, state);
-		StateBasedContainer result = new StateBasedContainer(resourceDescriptions, containerState);
+		
+		StateBasedContainer result = createStateBasedContainer(resourceDescriptions, containerState);
```
szarnekow: What if we move the `if (.. instanceof ..)` below also into the createStateBasedContainer method, rename that one to an overload of `createContainer` and let it return `IContainer` instead of an implementation type? That should offer more flexibility for clients.

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -110,7 +115,8 @@ public IResourceDescription getResourceDescription(URI uri) {
 	public Iterable<IEObjectDescription> getExportedObjectsByType(EClass type) {
```
szarnekow: Given that it is easier to override / provide a custom container, do we really need the changes in this class exception for the newly introduced getter?

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -41,6 +42,10 @@ public StateBasedContainer(IResourceDescriptions descriptions, IContainerState s
 		super(descriptions);
 		this.state = state;
 	}
+
+	protected IContainerState getState() {
```
szarnekow: Getter of a final field should be final I suppose.

tivervac:
@cdietrich I haven't had the time to do more testing, is it important that I do more?

cdietrich:
i just want to make sure we dont have a degredation here

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
tisis2: the idea of that whole change was to solve the problem at its cause, not giving the user an easier way solving the issue on his own.

acutally the fix in our codebase looks like:
`class FilteringResourceDescriptionStateBaseContainerManager extends StateBasedContainerManager {

	@FinalFieldsConstructor
	static class FilteringResourceDescriptions implements IResourceDescriptions {
		val IResourceDescriptions delegate
		val Predicate<IResourceDescription> predicate

		override getAllResourceDescriptions() {
			delegate.allResourceDescriptions.filter[predicate.test(it)]
		}

		override getResourceDescription(URI uri) {
			val res = delegate.getResourceDescription(uri)
			
			if(res === null) return null
			
			if (predicate.test(res)) {
				return res
			}
			return null
		}

		override getExportedObjects() {
			return allResourceDescriptions.map[exportedObjects].flatten
		}

		override getExportedObjects(EClass type, QualifiedName name, boolean ignoreCase) {
			return allResourceDescriptions.map[getExportedObjects(type, name, ignoreCase)].flatten
		}

		override getExportedObjectsByObject(EObject object) {
			return allResourceDescriptions.map[getExportedObjectsByObject(object)].flatten
		}

		override getExportedObjectsByType(EClass type) {
			return allResourceDescriptions.map[getExportedObjectsByType(type)].flatten
		}

		override isEmpty() {
			return getExportedObjects.isEmpty
		}
	}

	// copied from super. created a pull request for xtext https://github.com/eclipse/xtext-core/pull/1253
	override protected createContainer(String handle, IResourceDescriptions resourceDescriptions) {
		val globalState = getState(resourceDescriptions)

		// actual change
		val filteringDescriptions = new FilteringResourceDescriptions(resourceDescriptions) [
			globalState.containsURI(handle, it.URI)
		]
		var IContainerState containerState = new ContainerState(handle, globalState)
		var StateBasedContainer result = new StateBasedContainer(filteringDescriptions, containerState)
		// end
		if(globalState instanceof FlatResourceSetBasedAllContainersState) result.setUriToDescriptionCacheEnabled(false)
		return result
	}

}

class ContainerState implements IContainerState {
	final String root
	final IAllContainersState globalState

	protected new(String root, IAllContainersState globalState) {
		this.root = root
		this.globalState = globalState
	}

	override Collection<URI> getContents() {
		return globalState.getContainedURIs(root)
	}

	override boolean contains(URI uri) {
		return globalState.containsURI(root, uri)
	}

	override boolean isEmpty() {
		return globalState.isEmpty(root)
	}

	override String toString() {
		return '''container '«»«root»' from «globalState»'''
	}
}`

tisis2:
i would be interested to really see a setup with a drop of performance to be able to investigate it a little but actually i can not get into such a situation with our code base. if anyone could provide setup where this occurs i can take a look at it

cdietrich:
unfortunately my codebase cannot given away for testing.
i am also no sure how others out there have currently optimized for this issue

=> i wonder how to compromise

- make the customization more easy (here we are  d’accord i think)
- i wonder if we can ship the old and the new variant and if yes with what version as default

tisis2:
i dont know how the change should be now for the framework. do we want to put anything in that changes performance or just changin visibility of things to offer easier points to overwrite.

is the issue really fixed when we just change visibility?

aren't there any example projects of xtext maybe where we could test the changes better?

cdietrich:
i am totally unsure.
making the overriding more easy is something we should definitively do.
but i still hope @szarnekow finds the time to test this with other big projects.
i think in normal sized Xtext projects the change does not really matter.

@szarnekow what do you think?

szarnekow:
I don't have any bigger project that makes extensive use of the affected APIs, so please don't expect too much insight in that regard.
But I do have a strong preference towards keeping this change as small as it is and focus on easier customisability. My review remarks are still valid, though.

cdietrich:
i asked the colleagues in a former project of mine to measure and these are the numbers

With Patch:

        - 1:55m
        - 1:49m
        - 1:48m

Times without Patch:

        - 2:30m
        - 2:13m
        - 2:13m

so in this case there is a degredation too
=> i propose

- introduce the proposed changes to make extension easier
- for me it also would be ok the ship and unbound subclass as well. but as this one would be untested i would prefer not to do that

szarnekow:
Did you mix up the times?

cdietrich:
yes

szarnekow:
25 secs out of 140 is tremendous :(

tisis2:
that seems to be really bad... but i guess also that project is not open source (or other possibilities to try that out on my machine)?

i will change the pull request when i fin time, so it will just contain changes that would make it easier to extend that behavior.

cdietrich:
yes this other project is not open source neither

cdietrich:
@szarnekow what do you think?

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -61,22 +66,22 @@ public boolean apply(IEObjectDescription input) {
 
 	@Override
 	public boolean hasResourceDescription(URI uri) {
-		return state.contains(uri);
+		return getState().contains(uri);
```
szarnekow: Now that we do have the final getter for the final field, it would be ok to continue using the field here, but each JVM that I know will manage to inline the getter. So this should be ok.

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainerManager.java
```diff
@@ -84,9 +84,16 @@ protected IAllContainersState getState(IResourceDescriptions resourceDescription
 	protected IContainer createContainer(String handle, IResourceDescriptions resourceDescriptions) {
 		IAllContainersState state = getState(resourceDescriptions);
 		IContainerState containerState = new ContainerState(handle, state);
+		
+		return createContainer(resourceDescriptions, state, containerState);
+	}
+
+	protected IContainer createContainer(IResourceDescriptions resourceDescriptions, IAllContainersState allContainerState, IContainerState containerState) {
```
szarnekow: Missing `@since` tag.

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -41,6 +42,10 @@ public StateBasedContainer(IResourceDescriptions descriptions, IContainerState s
 		super(descriptions);
 		this.state = state;
 	}
+
+	protected final IContainerState getState() {
```
szarnekow: Missing `@since` tag

PR #1265, opened 2019-10-28
[eclipse/xtext#1568] Remove Buildship classpath container

kthoms:
These changes were automatically applied to the launch configurations on
a fresh workspace setup.


tivervac:
The change looks ok but I'll try oomphing as well to confirm your findings

PR #1280, opened 2019-11-06
[eclipse/xtext#1573] Provide proper build.properties

szarnekow:
[eclipse/xtext#1573] Provide proper build.properties


org.eclipse.xtext/.cvsignore
```diff
@@ -1,4 +0,0 @@
-emf-gen
```
szarnekow: Beautiful, isn't it :)

PR #1281, opened 2019-11-06
[gitpod] added basic .gitpod.yml

JanKoehnlein:
An initial gitpod config. Pre-builds should be enabled.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,2 @@
+tasks:
+- init: ./1-gradle-build.sh


.gitpod.yml
```diff
@@ -0,0 +1,2 @@
+tasks:
+- init: ./gradlew build
```
cdietrich: any reason not to call  `./1-gradle-build.sh`

cdietrich:
@kthoms @Hejado any idea why https://ci.eclipse.org/xtext/job/xtext-core/job/jk%252Fgitpod/
does not build? bad branch name?

cdietrich:
`12:37:59  ERROR: Labels must follow required specs - https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set: xtext-core-jk/gitpod-4
12:37:59  Finished: FAILURE`

cdietrich:
```
12:37:59  ERROR: Labels must follow required specs - https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set: xtext-core-jk/gitpod-4
12:37:59  Finished: FAILURE
```

.gitpod.yml
JanKoehnlein: Nope, didn't know it existed. Just followed the instructions from `README.MD`.

cdietrich:
created https://github.com/eclipse/xtext/issues/1595 about the label calculation

JanKoehnlein:
I used `jk/gitpod` as a branch name. Is it no longer the default to use a scheme `<user>/<name>`?

cdietrich:
the deault scheme is `<user>_<name>`
looks like your the first one using `/` since we moved to JIRO

but it should work with `/ ` as well. 
thus i created the issue to look at it.

LorenzoBettini:
I've just tried that and it seems to work apart a few compilation errors :)
actually it's the first time I'm using Gitpod and I can't seem to find a way to run unit tests... is that intentional?

cdietrich:
it also looks like we dont have a gradle.properties that sets 
`org.gradle.jvmargs = -Xmx2G`

cdietrich:
@LorenzoBettini to me it looks like java language server does not support project dependency substitution. it uses dependencies from maven central (M2) instead

cdietrich:
dependencies

LorenzoBettini:
> @LorenzoBettini to me it looks like java language server does not support project dependency substitution. it uses dependencies from maven central (M2) instead

I'm not sure I understand what you mean by "project dependency substitution"... is that related to compilation error or to the fact I cannot run JUnit tests?

cdietrich:
no i see compile errors.

LorenzoBettini:
and concerning JUnit tests it looks like that's not implemented yet: https://github.com/eclipse-theia/theia/issues/3003

JanKoehnlein:
I am inclined to revert this. Gradle's Xtend build and the Java LS seem to cause too many synchronization problems.

szarnekow:
I'm also in favour of making this available when it's more robust. The current state looks promising, but it's not quite there yet.

PR #1284, opened 2019-11-06
added gradle properties

cdietrich:
added gradle properties


cdietrich:
i am still experimenting. in general it looks like yes.
but as we will use 2 in xtend and extras we could use 2 here too to be on the safe side

PR #1285, opened 2019-11-07
[gitpod] removed .gitpod.yml again

JanKoehnlein:
Seems like the Java LS is not ready to handle our mix of Xtend and
Java, even when we build with Gradle. Having a Java-only build
doesn't make to much sense.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,2 +1,0 @@
-tasks:
-- init: ./1-gradle-build.sh


JanKoehnlein:
Sorry for the noise. See https://github.com/eclipse/xtext-core/pull/1281 for more details.

szarnekow:
> Sorry for the noise. 

No worries. Would be noise if the interaction between different LS could be improved such that we have a nice interplay between Xtend, Java, other Xtext langs etc.

PR #1289, opened 2019-11-09
prepare wizard projects for gradle 6.0

cdietrich:
prepare wizard projects for gradle 6.0

### DIFF
--- a/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/Scope.java
+++ b/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/Scope.java
@@ -8,9 +8,9 @@
 package org.eclipse.xtext.xtext.wizard;
 
 public enum Scope {
-	COMPILE("compile", "compile"),
-	TESTCOMPILE("test", "testCompile"),
-	TESTRUNTIME("test", "testRuntime"),
+	COMPILE("compile", "implementation"),
+	TESTCOMPILE("test", "testImplementation"),
+	TESTRUNTIME("test", "testRuntimeOnly"),
 	PROVIDED("provided", "providedCompile");
 
 	private String mavenNotation;
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/build.gradle
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/build.gradle
@@ -15,7 +15,7 @@ subprojects {
 	
 	apply plugin: 'java'
 	dependencies {
-		compile platform("org.eclipse.xtext:xtext-dev-bom:${xtextVersion}")
+		implementation platform("org.eclipse.xtext:xtext-dev-bom:${xtextVersion}")
 	}
 
 	apply plugin: 'org.xtext.xtend'
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/org.xtext.example.full.ide/build.gradle
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/org.xtext.example.full.ide/build.gradle
@@ -1,7 +1,7 @@
 dependencies {
-	compile project(':org.xtext.example.full')
-	compile "org.eclipse.xtext:org.eclipse.xtext.ide:${xtextVersion}"
-	compile "org.eclipse.xtext:org.eclipse.xtext.xbase.ide:${xtextVersion}"
+	implementation project(':org.xtext.example.full')
+	implementation "org.eclipse.xtext:org.eclipse.xtext.ide:${xtextVersion}"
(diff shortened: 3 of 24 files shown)


cdietrich:
My assumption is that a dsl s self contained and not reused thus Deps are defined as impl and not as api

cdietrich:
@oehme what do you think / recommend?

oehme:
You should use `api` whenever a dependency is used on the public API of a project. E.g. the `ide` part references types from the core dsl part in its public API und thus `ide` should have an API dependency on the core of the language. `implementation is for things that are used purely internally, e.g. in types that don't belong to the public API of a project or in local variables or private fields.

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/Scope.java
```diff
@@ -8,9 +8,9 @@
 package org.eclipse.xtext.xtext.wizard;
 
 public enum Scope {
-	COMPILE("compile", "compile"),
-	TESTCOMPILE("test", "testCompile"),
-	TESTRUNTIME("test", "testRuntime"),
+	COMPILE("compile", "implementation"),
```
oehme: I'd probably split this into `API` and `IMPLEMENTATION`. You can map them both to `compile` in Maven and to `api`/`implementation` in Gradle.

cdietrich: I don’t think we will have implementation ever then as we don’t distinguish between the two in eclipse and thus won’t in Gradle

oehme: I see, you mean that we reexport everything? In that case `api` would be a more appropriate mapping.

cdietrich: Yes and no. We at least export at least one package and thus the likelihood that that package contains something that leaks into api is quite high

cdietrich: but in the mydsl case we replicate the dependency to mydsl to mydsl.ide, mydsl.web and mydsl.test and thus cirumvent the problem at all. this is why implementation works and i used it here

oehme: Yeah, but that's not really a nice pattern. It's clearly an API dependency and shouldn't have to be repeated all the time :)

cdietrich: Yes but I don’t know how to get in under one hood with Gradle and Maven.
And thus would repeat the dependency even with api

cdietrich: Same with p2 we don’t do reexport and thus need to duplicate deps 
https://github.com/eclipse/xtext-core/blob/0231bea4e46427ae7175adb70cc5888b7d7bd751/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ProjectDescriptor.xtend#L139

cdietrich:
i am still puzzled

- use `api`
- use `api` and do proper reexports for maven/tycho
- keep `implementation`

cdietrich:
@oehme is is possible `api` is not available?

```
> Could not find method api() for arguments [org.eclipse.xtext:org.eclipse.xtext:2.19.0] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

what is the difference between `java` and  `java-libary` plugin. is it possible `api` is available with the latter only?

cdietrich:
i have created a second variant here: https://github.com/eclipse/xtext-core/compare/cd_wizG60_var_api

oehme:
As I said, if a project exposes types on its API and downstream projects are expected to use those types, it has to be an `api` dependency. And yes, only libraries can have an `api` (it doesn't make sense for applications for example). So if you want to do the minimal change, use `api` everywhere.

cdietrich:
to be updated to use API

drkstr101:
`api` Should be used if that dependency is exposed in the public ABI (a paramerter type or return type on a public method/prop). Usually one will want to default to `implementation` unless this causes issues. This mostly just affects dependency resolution in downstream consumers. Two projects would most likely wish to sahe the same version of `api` dependencies, while `implementation` dependencies can be on different versions with little issue.

cdietrich:
the main problem is: we dont know which deps we put into api and which not.
so its hard to judge

drkstr101:
I'll try to sort it out. It can be a bit tricky at times. Sometimes just trial an error.  `¯\_(ツ)_/¯`

drkstr101:
@cdietrich

```
buildscript {
  repositories {
    jcenter()
  }
  dependencies {
    classpath 'org.xtext:xtext-gradle-plugin:2.0.2'
    // trim...
  }
}
```

Was the intent here to link in the subproject, or a previous build of plugin to "bootstrap itself"?

PR #1292, opened 2019-11-15
[eclipse/xtext#1504] added Eclipse-SourceReferences to manifests

cdietrich:
[eclipse/xtext#1504] added Eclipse-SourceReferences to manifests


cdietrich:
the main problem is that gradle does not really like single projects so the project will not be really usable when imported into a eclipse (without root projects etc)

PR #1300, opened 2019-11-22
fix unwanted build properties changes

cdietrich:


org.eclipse.xtext.testlanguages/build.properties
```diff
@@ -1,4 +1,5 @@
-bin.includes = .,\
+bin.includes = model/generated/,\
```
szarnekow: Model/ is already listed

PR #1301, opened 2019-11-25
[releng] Update Xtext version to 2.21.0

genie-xtext:
This change was brought to you by your friendly Xtext Genie.


kthoms:
Just an experiment

PR #1305, opened 2019-11-26
Use mwe2version in POMs and target platform, instead of hard-coding

LorenzoBettini:
This PR generates in the POMs of the projects created by the wizard the mwe2Version variable that is used for running the MWE2 generator from Maven, instead of hardcoding the version. This way, the developer can simply update that variable when he/she also updates the Xtext version in the parent POM.

Similarly, the `.target` file is generated by using `config.xtextVersion.mweVersion`; this has no effect on the generated `.target` file but at least it should be easier to keep the generation of the `.target` file consistent.

### DIFF
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/org.xtext.example.full/pom.xml
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/org.xtext.example.full/org.xtext.example.full.parent/org.xtext.example.full/pom.xml
@@ -39,7 +39,7 @@
 					<dependency>
 						<groupId>org.eclipse.emf</groupId>
 						<artifactId>org.eclipse.emf.mwe2.launch</artifactId>
-						<version>2.11.1</version>
+						<version>${mwe2Version}</version>
 					</dependency>
 					<dependency>
 						<groupId>org.eclipse.xtext</groupId>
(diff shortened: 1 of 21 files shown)


cdietrich:
i dont even know why we have it when we have it in the bom too

LorenzoBettini:
I don't think that would work: I don't think the BOM is taken into consideration for dependencies that are specific to Maven plugins (`maven-exec-plugin` in this case)...

cdietrich:
ok, i remember now

LorenzoBettini:
> ok, i remember now

you mean that it's like I said?

cdietrich:
yes

LorenzoBettini:
OK, then I wait for your review

cdietrich:
@LorenzoBettini  i see this changes parents for non tycho maven projects too.
is this intended?

LorenzoBettini:
The `mwe2Version` is a Maven property so it should work also for non Tycho projects, or did I miss something

cdietrich:
yes but changing it wont change anything so users might be confused
@kthoms what do you think?

LorenzoBettini:
Oh I see what you mean: in that case the `exec-maven-plugin` does not need another dependency!
Please let me know if you think it's worthwhile not to touch those plain Maven projects

cdietrich:
i would prefer so.

LorenzoBettini:
I think I'll open a new PR, OK

LorenzoBettini:
Here it is https://github.com/eclipse/xtext-core/pull/1306

sorry if I merged too soon, but I thought you had approved the PR

Issue #1304, opened 2019-11-26
Cannot run CliWizardIntegrationTest


LorenzoBettini:
I cannot run `org.eclipse.xtext.xtext.wizard.cli.CliWizardIntegrationTest` as a JUnit test, due to this exception from EMF: `Missing properties: org/eclipse/xtext/util/plugin.properties`.

I can run it as a JUnit Plug-in test, but then I cannot run its main to update the expectations:

```
Exception in thread "main" java.lang.ExceptionInInitializerError
Caused by: java.util.MissingResourceException: Missing properties: org/eclipse/xtext/util/plugin.properties
	at org.eclipse.emf.common.util.DelegatingResourceLocator.getBaseURL(DelegatingResourceLocator.java:200)
	at org.eclipse.xtext.util.XtextVersion.readVersionFromManifest(XtextVersion.java:96)
	at org.eclipse.xtext.util.XtextVersion.getCurrent(XtextVersion.java:42)
	at org.eclipse.xtext.xtext.wizard.WizardConfiguration.<init>(WizardConfiguration.java:50)
	at org.eclipse.xtext.xtext.wizard.cli.CliWizardIntegrationTest$1.<init>(CliWizardIntegrationTest.java:390)
	at org.eclipse.xtext.xtext.wizard.cli.CliWizardIntegrationTest.newProjectConfig(CliWizardIntegrationTest.java:390)
	at org.eclipse.xtext.xtext.wizard.cli.CliWizardIntegrationTest.<clinit>(CliWizardIntegrationTest.java:150)
```

cdietrich:
have never seen this. how do you exactly run the test.

LorenzoBettini:
Simply from Eclipse: right-click -> Run As -> JUnit Test

similarly for running its main as a standard Java application

cdietrich:
@szarnekow can it be this is a side effect of the build.properties thing?

cdietrich:
i can reproduce this now too.
was working until last week

LorenzoBettini:
Indeed I've been using that test lots of time in the past and it was always working...

cdietrich:
@kthoms is it possible the xtend2java conversion changed exception handling here?

cdietrich:
indeed the catch exception in xtextversion was not ported properly

cdietrich:
@kthoms this should be fixed for 2.20

kthoms:
will do!

kthoms:
[...]

PR #1309, opened 2019-11-27
[#1304] Refactor retrieval of the own manifest URL

kthoms:
This changes the way the MANIFEST.MF for org.eclipse.xtext.util is
located. It uses the classloader to locate the class resource of
XtextVersion and resolves the URL relative to the resource. It considers
the case that the class resource can be either on file system or
packaged in an archive, which varies the relative location.

Add a test case for XtextVersion#getCurrent() to cover the regression
that arised by the recent refactoring from Xtend to Java.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.java
@@ -7,30 +7,20 @@
  */
 package org.eclipse.xtext.util;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.jar.Manifest;
 
-import org.eclipse.emf.common.EMFPlugin;
-import org.eclipse.emf.common.util.ResourceLocator;
+import org.eclipse.emf.common.util.URI;
 
 /**
  * @since 2.9
  */
 public class XtextVersion {
-	private static class Plugin extends EMFPlugin {
-		public static final XtextVersion.Plugin INSTANCE = new XtextVersion.Plugin();
-
-		private Plugin() {
-			super(new ResourceLocator[] {});
-		}
-
-		@Override
-		public ResourceLocator getPluginResourceLocator() {
-			return null;
-		}
-	}
 
 	private final String version;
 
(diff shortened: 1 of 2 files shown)


cdietrich:
i have no idea if this works in all cases (jar, eclipse, ...)

kthoms:
This shows the unit test scenario. The URL contains `/bin/main` here.
screenshot 346

The computed URL will strip this off and retrieve the right URL for the manifest.
screenshot 347

For the archive scenario I have exported `org.eclipse.xtext.utils` and added it to the test's classpath as first entry.

screenshot 348

This will resolve these URLs for the class and the manifest:
```
jar:file:/Users/thoms/temp/plugins/org.eclipse.xtext.util_2.20.0.201911271326.jar!/org/eclipse/xtext/util/XtextVersion.class
jar:file:/Users/thoms/temp/plugins/org.eclipse.xtext.util_2.20.0.201911271326.jar!/META-INF/MANIFEST.MF
```

kthoms:
The plugin case should be similar to the archive case.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.java
```diff
@@ -92,24 +82,73 @@ public String toString() {
...
+		try {
+			String ownClassAsResourcePath = XtextVersion.class.getName().replace('.', '/')+".class";
+			URL ownClassResourceURL = XtextVersion.class.getClassLoader().getResource(ownClassAsResourcePath);
+			URI ownClassResourceURI = URI.createURI(ownClassResourceURL.toURI().toString());
+			
+			// The XtextVersion class has 5 segments. When the class is bundled in a jar the manifest is located 
+			// relative to the root.
+			// In a unit test the class is located in the build output directory, but the manifest
+			// is found relative to the project root. This can be either 1 or 2 levels above (bin/main in Gradle or target in Maven).
+			for (int i=5; i<=7; i++) {
```
cdietrich: am not sure if i like this.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.java
```diff
@@ -92,24 +82,73 @@ public String toString() {
...
 	}
+	
+	/**
+	 * Locates the URL of this bundle's manifest file. This is done by locating the manifest file
+	 * relative to the class resource of this class. The location can be either be a file URL in the context
+	 * of unit test execution or an archive URL for CI/release bundle.
+	 * 
+	 * @throws FileNotFoundException When the MANIFEST.MF for org.eclipse.xtext.util was not found
+	 */
+	private static URL getOwnManifestURL () throws FileNotFoundException {
```
szarnekow: I'd prefer something along the lines of 
```
ClassLoader cl = getClass().getClassLoader();
Enumeration<URL> allManifestMF = cl.getResources(""META-INF/MANIFEST.MF";
while(allManifestMf.hasMoreElements()) {
  // check the URL for xtext.util and proceed
  
}
```

org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.java
```diff
@@ -92,24 +82,73 @@ public String toString() {
...
-				if (version.endsWith(".qualifier")) {
-					return version.replace(".qualifier", "-SNAPSHOT");
-				} else {
-					return version.substring(0, version.lastIndexOf("."));
+		try {
+			URL ownManifestURL = getOwnManifestURL();
+			
+			try (InputStream is = ownManifestURL.openStream()) {
+				Manifest manifest = new Manifest(is);
+				String version = manifest.getMainAttributes().getValue("Maven-Version");
```
szarnekow: Please add a guard here for the bundle name to ensure, that we are working with the correct manifest

org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.java
kthoms: I had this approach first, too, but it did not work. I did not came across the manifest of org.eclipse.xtext.util. What was I missing?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.java
kthoms: Could check for file or archive URLs. But how to make the URL reliable relative to the base URL? I agree it does not look too nice, but I did not came to a better solution.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.java
cdietrich: My assumption is that Gradle knows nothing about that 
As the manifest is not on the classpath when not executed with a jar but from eclipse

org.eclipse.xtext.util/src/org/eclipse/xtext/util/XtextVersion.java
kthoms: Makes sense, will do.

szarnekow:
I'm most likely missing something. What was wrong with the EMFPlugin approach? And why isn't the same thing wrong for `org.eclipse.xtext.generator.Generator.Plugin` ?

kthoms:
The issue #1304 arised since in [XtextVersion.readVersionFromManifest()](https://github.com/eclipse/xtext-core/blob/v2.20.0.M2/org.eclipse.xtext.util/xtend-gen/org/eclipse/xtext/util/XtextVersion.java) the line `XtextVersion.Plugin.INSTANCE.getBaseURL()` was guarded in a try block, and with the refactoring to Java I have moved that outside of the try-block. So in a unit test there was the exception that we now saw thrown and catched. That resulted in a `null` return value, and `getCurrent()` would return `unknown`.

> And why isn't the same thing wrong for org.eclipse.xtext.generator.Generator.Plugin ?

Right, the `Plugin.INSTANCE.getBaseURL()` is used by `getVersion()` in the same manner. But there is no test scenario that would check that `getVersion()` does not return `null` due to a catched exception. So it can be that the same problem is hidden.

I decided to go this way to avoid the exception and to try to read the actual value from the own manifest.

Here it is shown that the exception occured with the old xtend code:
screenshot 351

szarnekow:
Does it make sense to move the dance around indizes etc into the catch block and rely on the old mechanism for the cases, where it did work like expected?

kthoms:
I'll provide a second patch that just moves the problematic line into the try block to solve the regression first. The original code was only sufficient enough because it was not necessary to use a real version in dev mode.
I understand that the "magic indizes" are a bit ugly. I'll refactor that in a way to check for either file or archive URIs and compute the manifest's URL based on that.

cdietrich:
PR is obsolete

PR #1316, opened 2019-12-05
Add end line and end column positions to Issue. Also initialise lines and columns with 0.

mmews-n4:
Currently, end line and end column information is not part of `Issue`, `IssueImpl`, `IssueLocation`, `AbstractDiagnostic`, and `ExceptionDiagnostic`. It is helpful to have this information in these classes, because otherwise (i.e. now in `LanguageServerImpl#toDiagnostic(Issue)`) this information has to be computed by loading the `Document` again.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
@@ -453,13 +453,7 @@ public class LanguageServerImpl implements LanguageServer, WorkspaceService, Tex
 		initialized.thenAccept((initParams) -> {
 			PublishDiagnosticsParams publishDiagnosticsParams = new PublishDiagnosticsParams();
 			publishDiagnosticsParams.setUri(uriExtensions.toUriString(uri));
-			// this is not a premature optimization but a trick to handle issues of deleted resources
-			if (!issues.iterator().hasNext()) {
-				publishDiagnosticsParams.setDiagnostics(Collections.emptyList());
-			} else {
-				publishDiagnosticsParams.setDiagnostics(
-						workspaceManager.doRead(uri, (document, resource) -> toDiagnostics(issues, document)));
-			}
+			publishDiagnosticsParams.setDiagnostics(toDiagnostics(issues));
 			client.publishDiagnostics(publishDiagnosticsParams);
 		});
 	}
@@ -468,11 +462,11 @@ public class LanguageServerImpl implements LanguageServer, WorkspaceService, Tex
 	 * Convert the given issues to diagnostics. Does not return any issue with severity {@link Severity#IGNORE ignore}
 	 * by default.
 	 */
-	protected List<Diagnostic> toDiagnostics(Iterable<? extends Issue> issues, Document document) {
+	protected List<Diagnostic> toDiagnostics(Iterable<? extends Issue> issues) {
 		List<Diagnostic> result = new ArrayList<>();
 		for (Issue issue : issues) {
 			if (issue.getSeverity() != Severity.IGNORE) {
-				result.add(toDiagnostic(issue, document));
+				result.add(toDiagnostic(issue));
 			}
 		}
 		return result;
@@ -481,19 +475,21 @@ public class LanguageServerImpl implements LanguageServer, WorkspaceService, Tex
 	/**
 	 * Convert the given issue to a diagnostic.
 	 */
-	protected Diagnostic toDiagnostic(Issue issue, Document document) {
+	protected Diagnostic toDiagnostic(Issue issue) {
 		Diagnostic result = new Diagnostic();
 		result.setCode(issue.getCode());
 		result.setMessage(issue.getMessage());
 		result.setSeverity(toDiagnosticSeverity(issue.getSeverity()));
-		Position start = document.getPosition(issue.getOffset());
-		Position end = document.getPosition(issue.getOffset() + issue.getLength());
+
+		// line and column numbers in LSP are 0-based
+		Position start = new Position(issue.getLineNumber() - 1, issue.getColumn() - 1);
+		Position end = new Position(issue.getLineNumberEnd() - 1, issue.getColumnEnd() - 1);
 		result.setRange(new Range(start, end));
 		return result;
 	}
 
 	/**
-	 * Convert the serverity to a lsp {@link DiagnosticSeverity}.
+	 * Convert the severity to a lsp {@link DiagnosticSeverity}.
 	 *
 	 * Defaults to severity {@link DiagnosticSeverity#Hint hint}.
 	 */
--- a/org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/AbstractDiagnostic.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/AbstractDiagnostic.java
@@ -58,6 +58,25 @@ public abstract class AbstractDiagnostic implements Diagnostic {
 			return node.getStartLine();
 		return -1;
 	}
+	
+
+	@Override
+	public int getLineEnd() {
+		INode node = getNode();
+		if (node != null)
+			return node.getEndLine();
+		return -1;
+	}
+
+	@Override
+	public int getColumnEnd() {
+		INode node = getNode();
+		if (node != null) {
+			LineAndColumn lineAndColumn = NodeModelUtils.getLineAndColumn(node, getOffset() + getLength());
+			return lineAndColumn.getColumn();
+		}
+		return 0;
+	}
 
 	@Override
 	public String getLocation() {
@@ -86,5 +105,4 @@ public abstract class AbstractDiagnostic implements Diagnostic {
 		b.append(getMessage());
 		return b.toString();
 	}
-
 }
--- a/org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
@@ -58,6 +58,8 @@ public interface Diagnostic extends org.eclipse.emf.ecore.resource.Resource.Diag
      * Returns the end line location of the issue within the source.
      * Line <code>1</code> is the first line.
      * @return the end line location of the issue.
+	 * 
+	 * @since 2.21
      */
     int getLineEnd();
 
@@ -65,6 +67,8 @@ public interface Diagnostic extends org.eclipse.emf.ecore.resource.Resource.Diag
      * Returns the end column location of the issue within the source.
      * Column <code>1</code> is the first column.
      * @return the end column location of the issue.
+	 * 
+	 * @since 2.21
      */
     int getColumnEnd();
 
--- a/org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
@@ -54,4 +54,18 @@ public interface Diagnostic extends org.eclipse.emf.ecore.resource.Resource.Diag
 	 */
 	int getLength();
 
+    /**
+     * Returns the end line location of the issue within the source.
+     * Line <code>1</code> is the first line.
+     * @return the end line location of the issue.
+     */
+    int getLineEnd();
+
+    /**
+     * Returns the end column location of the issue within the source.
+     * Column <code>1</code> is the first column.
+     * @return the end column location of the issue.
+     */
+    int getColumnEnd();
+
 }
--- a/org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
@@ -10,7 +10,8 @@ package org.eclipse.xtext.diagnostics;
 /**
  * A specialized {@link org.eclipse.emf.ecore.resource.Resource.Diagnostic} that knows 
  * about the region in the document, e.g. the {@link #getOffset()} and {@link #getLength()}
- * of the issue source.
+ * of the issue source. A region starts in a start line before a start column and ends in an end
+ * line before an end column.
  * 
  * Implementors should inherit from {@link AbstractDiagnostic} instead of implementing this 
  * interface directly.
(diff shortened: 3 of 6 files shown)


cdietrich:
@mmews-n4 is this a (better) fix for https://github.com/eclipse/xtext/issues/2538

org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
```diff
@@ -54,4 +54,18 @@
 	 */
 	int getLength();
 
+    /**
+     * Returns the end line location of the issue within the source.
+     * Line <code>1</code> is the first line.
+     * @return the end line location of the issue.
+     */
+    int getLineEnd();
```
cdietrich: ihmo this is a breaking api change

org.eclipse.xtext/src/org/eclipse/xtext/validation/Issue.java
```diff
@@ -43,17 +43,29 @@
 	 * Returns the one-based line number of the issue.
 	 */
 	Integer getLineNumber();
+
+	/**
+	 * Returns the one-based line number of the end of the issue.
+	 */
+	Integer getLineNumberEnd();
```
cdietrich: this is also a breaking api change

org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
mmews-n4: Yes, but this class and class `Issue` have the annotation `@noimplement`. Maybe that would be okay then? However, I add `@since 2.21` to the new methods.

cdietrich: i see. @szarnekow what do you think?

mmews-n4:
@cdietrich Regarding your question about eclipse/xtext#2538:
No, this PR will not change the region according to LSP specification cited in eclipse/xtext#2538. I can confirm that VSCode tolerates indices of end column to be greater than the column length and will set the issue underlines correctly (see the screenshot below). I did not check IntelliJ.

Bildschirmfoto 2019-12-05 um 13 32 16

To solve eclipse/xtext#2538, I would introduce another method (e.g. `isLineBreakAfterColumnEnd`) that would be necessary when calculating the correct end of the issue range. I'll prepare that.

cdietrich:
@mmews-n4 did you have a look at thw workaround i introduced with eclipse/xtext#2538 
the workaround is there to avoid this change here

cdietrich:
https://github.com/eclipse/xtext-core/commit/632dfd084616323317eb2ead75dd968bb18c2d77

mmews-n4:
@cdietrich Thanks, I wasn't aware of 632dfd0 before. Technically, both of the solutions are similar. However, 632dfd084616323317eb2ead75dd968bb18c2d77 would need the following changes (imho):
- rework `LanguageServerImpl#toDiagnostic(Issue)`, especially regarding `endColumn`
- initialise variables in `Issue` to avoid NPE when autounboxing.

mmews-n4:
how do we proceed?

szarnekow:
We discussed which semantics we want to support and figured it would be good to implement the protocol exactly as specified. This is what I understood. Please correct me if I'm wrong, @cdietrich 
Assuming this file contents
```
first\n
second
```
where `\n` is an invisible newline character. There is no newline at the end of the file.
(Notation `((startline, startcol), (endline, endcol))`)
- Put an error on `first` by sending the range `((0,0), (1,0))`
- Put an error on `second` by sending the range `((1,0), (1,6))` - this is to be verified
- Put an error on the _newline_, which is not strictly supported by lsp, by marking the last char in the line with the newline: ((0, 4), (1,0))
- Put an error on the `i` in `first` by sending `((0, 1), (0,2))`

cdietrich:
since the end is exclusive
`((0,0), (0,5))` should also work.
setting the end to the first char of next line should be only relevant if we want to include the newline in the error.

but we should have some look how vscode (the ref impl) deals with these cases.

szarnekow:
Remark: from https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-15.md#position I understand that it possible to describe the column after the last char in the file, but not after the newline.
```
first\n
second
```
So `(0,5)` is after the `t` but before the line end.

cdietrich:
yes this is, when you want to mark the NL, include it in the marking, we have toi set it to the first char of the next line. which should be possible, as there is a newline

szarnekow:
> since the end is exclusive
> `((0,0), (0,5))` should also work.
> setting the end to the first char of next line should be only relevant if we want to include the newline in the error.

True. The examples above should most likely be updated to this:
```
first\n
second
```

- Put an error on first by sending the range ((0,0), (0,5))
- Put an error on second by sending the range ((1,0), (1,6))
- Put an error on the newline, which is not strictly supported by lsp, by marking the last char in the line with the newline: ((0, 5), (1,0)) - if the editor manages to highlight that, to be verified
- Put an error on the i in first by sending ((0, 1), (0,2))
- Put an error on EOF by marking the last char in the file ((1,5),(1,6)) - also to be verified

mmews-n4:
Acknowledged.

Regarding the IntelliJ issue: I cannot take that into account, since the documented information lacks the example that lead to this StringIndexOutOfBoundsException.

Do I understand correctly that including the line break as the last character is only necessary, if the xtext issue itself included the line break already? Put different: Assuming that xtext issues never end on a line break (which sounds reasonable), we do not need to care for jumping to the next line, right? (Same goes for EOF)

szarnekow:
A test with an issue as described in bullet point (3) must succeed. Same for the other described cases.

cdietrich:
there should be already a test for that be created with the issue i think

mmews-n4:
Ok. Then we need some special handling for this and a third variable in the Issue class (e.g. `endsAfterLineBreak`).

szarnekow:
I don't think so. The offset after the line break should already deliver the correct line number when asked for line and column.

mmews-n4:
In that case: What changes do you request?

org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
```diff
@@ -54,4 +54,22 @@
 	 */
 	int getLength();
 
+    /**
+     * Returns the end line location of the issue within the source.
+     * Line <code>1</code> is the first line.
+     * @return the end line location of the issue.
+	 *
```
szarnekow: Indentation glitch

org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
szarnekow: In general I'm fine with these being added here. I'm still a little but undecided if I'd like these to have default impls.

org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
```diff
@@ -54,4 +54,22 @@
 	 */
 	int getLength();
 
+    /**
+     * Returns the end line location of the issue within the source.
```
szarnekow: Please also state that the end line is inclusive. I'd like to see a sentence about issues that span multiple lines, especially issues that have a region that ends with a newline.

org.eclipse.xtext/src/org/eclipse/xtext/validation/DiagnosticConverterImpl.java
```diff
@@ -239,21 +244,28 @@ protected IssueLocation getLocationData(EObject obj, EStructuralFeature structur
 					containingFeature.isMany() ? ((EList<?>) container.eGet(containingFeature)).indexOf(obj)
 							: ValidationMessageAcceptor.INSIGNIFICANT_INDEX);
 		}
-		IssueLocation result = new IssueLocation();
-		result.lineNumber = 1;
-		result.column = 1;
-		result.offset = 0;
-		result.length = 0;
-		return result;
+		return new IssueLocation();
```
szarnekow: We used to init line and column to one. The new default is zero. We should stick to the old default for this case

org.eclipse.xtext/src/org/eclipse/xtext/validation/Issue.java
```diff
@@ -43,17 +43,31 @@
...
 	 * The first char in a line has column number 1, the number is one-based.
 	 * 
-	 * If no column information is available, returns -1.
-	 * 
 	 * @since 2.9
 	 */
 	Integer getColumn();
+	
+	/**
+	 * Returns the end column in the line of the issue. It's not the virtual end column but literally
```
szarnekow: Please state that this is exclusive

org.eclipse.xtext/src/org/eclipse/xtext/validation/Issue.java
szarnekow: Should this be a default method?

org.eclipse.xtext/src/org/eclipse/xtext/validation/Issue.java
```diff
@@ -116,6 +140,14 @@ public void setColumn(Integer column) {
 			this.column = column;
 		}
 
+		public Integer getColumnEnd() {
```
szarnekow: Missing since tags on the setters

org.eclipse.xtext/src/org/eclipse/xtext/validation/Issue.java
```diff
@@ -70,7 +84,9 @@
 		
 		private static Logger LOG = Logger.getLogger(IssueImpl.class);
 
-		private Integer length, lineNumber, offset, column;
+		private Integer offset = 0, length = 0;
```
szarnekow: I understand why it is a good idea to fix these to avoid NPEs.
We have to be careful though about the defaults that we assign.

org.eclipse.xtext/src/org/eclipse/xtext/validation/Issue.java
```diff
@@ -43,17 +43,31 @@
 	 * Returns the one-based line number of the issue.
 	 */
 	Integer getLineNumber();
+
+	/**
+	 * Returns the one-based line number of the end of the issue.
```
szarnekow: Please also state that a value '0' indicates 'no line information'

szarnekow (changes requested):

org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/Diagnostic.java
mmews-n4: If the end line character is inclusive, then I don't know why test `org.eclipse.xtext.ide.tests.server.ValidatorTest.testDiagnosticAtEndOfLineIncludingNewline()` is green. It looks to me as if the jump to the start of the next line is also done in `org.eclipse.emf.ecore.resource.Resource.Diagnostic`.

org.eclipse.xtext/src/org/eclipse/xtext/validation/DiagnosticConverterImpl.java
mmews-n4: It looks to me as if the contract*) was violated by the old implementation.

*) contract being
> Please also state that a value '0' indicates 'no line information'

org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/AbstractDiagnostic.java
```diff
@@ -58,6 +58,25 @@ public int getLine() {
 			return node.getStartLine();
 		return -1;
```
mmews-n4: mind the contract
> '0' indicates 'no line information'

org.eclipse.xtext/src/org/eclipse/xtext/validation/Issue.java
mmews-n4: AFAIR, we said the contract was: '<1' indicates 'no information'

org.eclipse.xtext/src/org/eclipse/xtext/validation/DiagnosticConverterImpl.java
szarnekow: At this particular point, we wanted to create an artificial location for a diagnostic, if we were not able to figure the "right" location. E.g. put it on the first line before the first char (offset zero, length zero, line 1, column 1)

org.eclipse.xtext/src/org/eclipse/xtext/diagnostics/AbstractDiagnostic.java
szarnekow: Maybe we should reverse the contract: valid line informations is 1 or higher.
-1, 0, null all indicate absent line information :/

org.eclipse.xtext/src/org/eclipse/xtext/validation/Issue.java
szarnekow: :+1:

org.eclipse.xtext/src/org/eclipse/xtext/validation/DiagnosticConverterImpl.java
mmews-n4: In that case, I can just set some "valid" values for lines and columns.

szarnekow: :+1:

mmews-n4:
The failing test is green on my machine. What might be the cause for failing on Jenkins?

cdietrich:
virtualization FTW / crappy Jenkins

org.eclipse.xtext/src/org/eclipse/xtext/validation/Issue.java
szarnekow: `Values smaller than 1 are invalid and indicate the absence of line information on this issue.`

szarnekow (approved):
Thank you for your patience.

szarnekow:
@cdietrich What do you think about this?

szarnekow:
@cdietrich What do you think about this?

cdietrich (approved):

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
```diff
@@ -481,19 +475,21 @@ private void publishDiagnostics(URI uri, Iterable<? extends Issue> issues) {
...
+		// line and column numbers in LSP are 0-based
+		Position start = new Position(issue.getLineNumber() - 1, issue.getColumn() - 1);
+		Position end = new Position(issue.getLineNumberEnd() - 1, issue.getColumnEnd() - 1);
 		result.setRange(new Range(start, end));
 		return result;
 	}
 
 	/**
-	 * Convert the serverity to a lsp {@link DiagnosticSeverity}.
+	 * Convert the severity to a lsp {@link DiagnosticSeverity}.
```
szarnekow: Appreciated attention to detail. 👍

org.eclipse.xtext/src/org/eclipse/xtext/validation/DiagnosticConverterImpl.java
```diff
@@ -239,21 +244,37 @@ protected IssueLocation getLocationData(EObject obj, EStructuralFeature structur
...
 	}
 
 	protected IssueLocation getLocationForNode(INode node) {
 		ITextRegionWithLineInformation nodeRegion = node.getTextRegionWithLineInformation();
+		int offset = nodeRegion.getOffset();
+		int length = nodeRegion.getLength();
+		return getLocationForNode(node, offset, length);
+	}
+
+	protected IssueLocation getLocationForNode(INode node, int offset, int length) {
```
szarnekow: Please add `@since` tag. Sorry, this slipped through.

PR #1322, opened 2019-12-10
Log Exceptions thrown in futures created by the RequestManger

mmews-n4:
The `LanguageServerImpl` uses the pattern of implementing methods from `org.eclipse.lsp4j` e.g. ` org.eclipse.lsp4j.services.TextDocumentService` and returns futures directly, see here:

```java
@Override
public CompletableFuture<List<Either<Command, CodeAction>>> codeAction(CodeActionParams params) {
	return requestManager.runRead((cancelIndicator) -> codeAction(params, cancelIndicator));
}
```

When an exception is thrown during the call to `codeAction(...)` this exception gets visible when joining the future, e.g. by calling `future.get()`. Since `lsp4j` does not do this, those exceptions are lost and never show up. Hence, the `RequestManager` should anticipate this and log exceptions itself.

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/concurrent/RequestManagerTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/concurrent/RequestManagerTest.xtend
@@ -9,16 +9,20 @@ package org.eclipse.xtext.ide.tests.server.concurrent
 
 import com.google.inject.Guice
 import com.google.inject.Inject
+import java.util.concurrent.ExecutionException
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicInteger
+import org.apache.log4j.Level
 import org.eclipse.xtext.ide.server.ServerModule
 import org.eclipse.xtext.ide.server.concurrent.RequestManager
+import org.eclipse.xtext.testing.logging.LoggingTester
 import org.junit.After
 import org.junit.Before
 import org.junit.Ignore
 import org.junit.Test
 
 import static org.junit.Assert.*
+import org.junit.Assert
 
 /**
  * @author kosyakov - Initial contribution and API
@@ -40,6 +44,35 @@ class RequestManagerTest {
 	def void tearDown() {
 		requestManager.shutdown
 		sharedState = null
+	}
+
+	@Test(timeout = 1000)
+	def void testLogException() {
+		val logResult = LoggingTester.captureLogging(Level.ALL, RequestManager, [
+			val future = requestManager.runWrite([
+				throw new RuntimeException();
+			], [])
+			
+			// join future to assert log later
+			try {
+				future.get
+			} catch (Exception e) {}
+		])
+		
+		logResult.assertLogEntry("Error during request:")
+	}
+
+	@Test(timeout = 1000, expected = ExecutionException)
+	def void testCatchException() {
+		LoggingTester.captureLogging(Level.ALL, RequestManager, [
+			val future = requestManager.runWrite([
+				throw new RuntimeException()
+			], [])
+
+			assertEquals('Foo', future.get)
+		])
+		
+		Assert.fail
 	}
 
 	@Test(timeout = 1000)
@@ -132,5 +165,4 @@ class RequestManagerTest {
 			Thread.sleep(10)
 		}
 	}
-
 }
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/concurrent/RequestManagerTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/concurrent/RequestManagerTest.xtend
@@ -23,6 +23,8 @@ import org.junit.Test
 
 import static org.junit.Assert.*
 import org.junit.Assert
+import org.eclipse.xtext.ide.server.concurrent.WriteRequest
+import org.eclipse.xtext.ide.server.concurrent.ReadRequest
 
 /**
  * @author kosyakov - Initial contribution and API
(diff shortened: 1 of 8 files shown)


mmews-n4:
Test might be green now.

cdietrich:
there are still test fails

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/concurrent/RequestManagerTest.xtend
```diff
@@ -42,6 +46,35 @@ class RequestManagerTest {
...
+	def void testCatchException() {
+		LoggingTester.captureLogging(Level.ALL, RequestManager, [
+			val future = requestManager.runWrite([
+				throw new RuntimeException()
+			], [])
+
+			assertEquals('Foo', future.get)
+		])
+		
+		Assert.fail
```
szarnekow: This last `Assert.fail` is not necessary

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/concurrent/RequestManagerTest.xtend
```diff
@@ -42,6 +46,35 @@ class RequestManagerTest {
...
+	def void testLogException() {
+		val logResult = LoggingTester.captureLogging(Level.ALL, RequestManager, [
+			val future = requestManager.runWrite([
+				throw new RuntimeException();
+			], [])
+			
+			// join future to assert log later
+			try {
+				future.get
+			} catch (Exception e) {}
```
szarnekow: We should probably propagate the exception here. Or use `future.join`

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/concurrent/RequestManagerTest.xtend
```diff
@@ -42,6 +46,35 @@ class RequestManagerTest {
 		sharedState = null
 	}
 
+	@Test(timeout = 1000)
+	def void testLogException() {
+		val logResult = LoggingTester.captureLogging(Level.ALL, RequestManager, [
```
szarnekow: Please add also tests for runRead and the second lambda of runWrite

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/concurrent/RequestManagerTest.xtend
mmews-n4: I know. I added that to highlight that the test case will not reach that point, but instead end with the thrown exception at `future.get`. You want that to be removed?

mmews-n4:
Okay, I made some changes. The tests are green locally... it might be due to synchronization of threads when waiting for the future, but I am not sure.

szarnekow:
The test is still yellow on the server. Don't understand why yet.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/concurrent/RequestManagerTest.xtend
szarnekow: I'm ok with it being there. Maybe `Assert.fail("unreachable")`

szarnekow:
I can reproduce the test failure

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
```diff
@@ -56,7 +58,13 @@ class RequestManager {
 	protected def <V> CompletableFuture<V> submit(AbstractRequest<V> request) {
 		requests += request
 		queue.submit(request)
-		return request.get
+		val future = request.get;
+		future.whenComplete[v, thr|
```
szarnekow: whenComplete returns a new future. That one should become the result of submit

szarnekow: This breaks the cancellation again.

szarnekow:
This will be fun ...
```
	@Test
	def void testCompleteableFuture() {
		val first = new CompletableFuture<Object>()
		val second = first.whenComplete[a,b|]
		second.cancel(true)
		Assert.assertTrue(second.isCancelled)
		Assert.assertTrue(first.isCancelled) // Fails here ...
	}
```

mmews-n4:
@szarnekow Thanks for your investigations. Let's see if the last commit can help it.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.xtend
```diff
@@ -101,6 +106,7 @@ class ReadRequest<V> extends AbstractRequest<V> {
 
 @FinalFieldsConstructor
 class WriteRequest<U, V> extends AbstractRequest<V> {
+	static final Logger LOG = Logger.getLogger(WriteRequest);
```
cdietrich: is this non private by intention?

cdietrich: why do we use final and not val here?

szarnekow: We have prepared the Java port of server.concurrent so this won't be an issue anymore.

cdietrich:
@szarnekow anything left?

szarnekow:
Looks certainly good for me. @tivervac please add your feedback post-merge. I'm closing this now.

Thank you, @mmews-n4 .

PR #1323, opened 2019-12-10
Make use of already implemented flag indexOnly and set it in IProjectConfig

mmews-n4:
Currently, the flag `indexOnly` from class `BuildRequest` is read in `IncrementalBuilder` but never set in production code (i.e. only set in some test code). This PR introduces the default interface method `IProjectConfig#isIndexOnly()` which can be overridden by client code. The result of this method will later be used to create the `BuildRequest` and then be respected by the incremental builder. This way, this flag allows e.g. library projects which neither have to be validated nor have to be generated.

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/IndexOnlyProjectTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/IndexOnlyProjectTest.xtend
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2017 TypeFox GmbH (http://www.typefox.io) and others.
+ * Copyright (c) 2019 NumberFour AG (http://www.enfore.com) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/IndexOnlyProjectTest.xtend
```diff
@@ -0,0 +1,118 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2017 TypeFox GmbH (http://www.typefox.io) and others.
```
cdietrich: wrong year

org.eclipse.xtext/src/org/eclipse/xtext/workspace/IProjectConfig.java
```diff
@@ -37,4 +37,16 @@
 	 * @return the workspace config
 	 */
 	IWorkspaceConfig getWorkspaceConfig();
+
+	/**
+	 * During the build, projects are indexed, validated and compiled. The return of this method configures whether this
+	 * project is validated and compiled or not.
+	 * 
+	 * @return true iff this project is indexed only
```
cdietrich: typo?

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/IndexOnlyProjectTest.xtend
szarnekow: The copyright should be from 2019 onwards for enfore

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/IndexOnlyProjectTest.xtend
```diff
@@ -0,0 +1,118 @@
...
+			'''
+			column = 'type F'.length
+			expectedReferences = '''
+				MyModel.testlang [[2, 1] .. [2, 4]]
+			'''
+		]
+	}
+	
+	/** Shows that the resource is not validated during initial build */
+    @Test
```
szarnekow: Mixed indentation

org.eclipse.xtext/src/org/eclipse/xtext/workspace/IProjectConfig.java
mmews-n4: AFAIK, 'iff' with two 'f' means 'if and only if' which would mean that this condition is complete, or to put it this way, that all other cases always return false (whereas 'if' would could be incomplete in that sense). Anyway, I can remove the second 'f' to avoid confusion.

PR #1327, opened 2019-12-13
[eclipse/xtext#1629] Migrate server.concurrent from Xtend to Java

szarnekow:
Migrate server.concurrent from Xtend to Java

see eclipse/xtext#1629


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/ReadRequest.java
```diff
@@ -0,0 +1,52 @@
...
+	public void run() {
+		if (result.isCancelled()) {
+			return;
+		}
+		this.executor.submit(() -> {
+			try {
+				cancelIndicator.checkCanceled();
+				result.complete(cancellable.apply(cancelIndicator));
+			} catch (Throwable t) {
+				if (!requestManager.isCancelException(t)) {
```
ArneDeutsch: This condition is not in the original code. Intentional or debug leftover?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.java
```diff
@@ -0,0 +1,107 @@
...
+	@Inject
+	private ExecutorService parallel;
+
+	@Inject
+	private OperationCanceledManager operationCanceledManager;
+
+	private final ExecutorService queue = Executors.newSingleThreadExecutor(
+			new ThreadFactoryBuilder().setDaemon(true).setNameFormat("RequestManager-Queue-%d").build());
+
+	private ArrayList<AbstractRequest<?>> requests = new ArrayList<>();
```
ArneDeutsch: Using "List" would be slightly more nice.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/ReadRequest.java
szarnekow: I see the description in the original code:
```
if (((e != null) && (!this.requestManager.isCancelException(e)))) {
```

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/AbstractRequest.java
```diff
@@ -0,0 +1,51 @@
...
+
+	protected AbstractRequest(RequestManager requestManager) {
+		this.requestManager = requestManager;
+		this.result = new CompletableFuture<>();
+		this.cancelIndicator = new RequestCancelIndicator(this.result);
+	}
+
+	@Override
+	public void cancel() {
+		this.cancelIndicator.cancel();
```
tivervac: No need for most `this.`'s in this class but I can live with it

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestCancelIndicator.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	@Override
+	public void cancel() {
+		this.requestFuture.cancel(true);
+	}
+
+	@Override
+	public boolean isCanceled() {
+		this.checkCanceled();
```
tivervac: Was already like this, but this is pretty nasty. It'll never return true, but instead throw an error

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.java
```diff
@@ -0,0 +1,107 @@
...
+	public <V> CompletableFuture<V> runRead(Function1<? super CancelIndicator, ? extends V> cancellable) {
+		return submit(new ReadRequest<>(this, cancellable, parallel));
+	}
+
+	/**
+	 * Perform the given write and run the cancellable logic afterwards.
+	 */
+	public <U, V> CompletableFuture<V> runWrite(
+			Function0<? extends U> nonCancellable,
+			Function2<? super CancelIndicator, ? super U, ? extends V> cancellable) {
```
tivervac: Can't we use `Supplier` and `BiFunction` here now?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestCancelIndicator.java
szarnekow: I don't appreciate this either. I'd like to keep the semantic cleanup separate, though.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.java
szarnekow: I'd opt for `Callable` and some `FunctionWithExceptionLike` thing where both allow exceptions to be thrown.
I've to think about the impact on backwards compatibility, though and if this is a thing that we should keep compatible.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/ReadRequest.java
ArneDeutsch: Interesting. I compared to my local code (pulled yesterday) and there the code is defnitly not there. Probably a statement that was introduced since yesterday and I'm a commit or two behind HEAD.

szarnekow: Indeed. Another PR with added logging was merged before I submitted this one here on top of the updated code.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/AbstractRequest.java
szarnekow: Updated

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestCancelIndicator.java
szarnekow: I'll add a comment to this impl of isCanceled rather than adjusting the semantics on the fly.

PR #1332, opened 2019-12-15
Cleaning up Xtend code: Organising imports to improve readability.

miklossy:


szarnekow:
Can we please avoid these cosmetic bulk changes. They are hard to review and don't add too much value.

PR #1333, opened 2019-12-16
[eclipse/xtext#1455] Resolve test failure in DocumentTest.

ArneDeutsch:
Document#getLineContent returns the '\r' of '\r\n' on windows. This
leads to a failing test. But the line content should explicitly NOT
return the newline marker. Hence fixing the implementation to fix the
test.


ArneDeutsch:
Really revert the URI thing? In my opinion it is more nice now AND it will not change again if organize imports in executed in that file.

miklossy:
Ok, in that case I am fine with that.

PR #1339, opened 2020-01-06
[eclipse/xtext#1605] update orbit to 2020-03

cdietrich:
[eclipse/xtext#1605] update orbit to 2020-03


miklossy:
Is Release_2.19 the right label?

Issue #1343, opened 2020-01-07
Javadoc contains reference to a non-existing interface


miklossy:
Currently, when building the `xtext-core`, the following javadoc warning is produced:

`org.eclipse.xtext\src\org\eclipse\xtext\formatting2\regionaccess\ITextRegionDiffBuilder.java:36: warning - Tag @link: reference not found: IHiddenRegionPartMerger`

**Hint:**
The `IHiddenRegionPartMerger` interface has been renamed to `IHiddenRegionPartAssociator` with the [commit](https://github.com/eclipse/xtext-core/commit/43b7d1df8c597178fc2fe74db0e173f6a0bbae15).

miklossy:
Fixed in 2.21.

PR #1344, opened 2020-01-07
[#1343] Fix invalid link in javadoc.

miklossy:


org.eclipse.xtext/src/org/eclipse/xtext/formatting2/regionaccess/ITextRegionDiffBuilder.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 TypeFox GmbH (http://www.typefox.io) and others.
+ * Copyright (c) 2017, 2019 TypeFox GmbH (http://www.typefox.io) and others.
```
cdietrich: We have 2020 now I guess

miklossy: You are absolutely right, I updated the PR.

PR #1347, opened 2020-01-09
Convert EMFGeneratorFragment2.trimMultiLineComment to an instance method

miklossy:
- to be able to override that method in derived classes.

### DIFF
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ecore/EMFGeneratorFragment2.xtend
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ecore/EMFGeneratorFragment2.xtend
@@ -623,7 +623,7 @@ class EMFGeneratorFragment2 extends AbstractXtextGeneratorFragment {
 		return genModel
 	}
 	
-	def static String trimMultiLineComment(String string) {
+	def String trimMultiLineComment(String string) {
 		return string.replace(' * ','').replaceAll("/\\*+\\s*|\\s*\\*+/", "").trim
 	}
 	
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/generator/EMFGeneratorFragment2Test.xtend
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/generator/EMFGeneratorFragment2Test.xtend
@@ -16,6 +16,8 @@ import org.eclipse.xtext.xtext.generator.ecore.EMFGeneratorFragment2
  */
 class EMFGeneratorFragment2Test {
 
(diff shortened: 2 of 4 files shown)


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ecore/EMFGeneratorFragment2.xtend
```diff
@@ -623,7 +623,7 @@ class EMFGeneratorFragment2 extends AbstractXtextGeneratorFragment {
 		return genModel
 	}
 	
-	def static String trimMultiLineComment(String string) {
+	def String trimMultiLineComment(String string) {
```
cdietrich: am not sure how api we consider this. @szarnekow what do you think?

szarnekow: My gut feeling is that this should have been an instance method from the beginning.

Question though: Why do you want to change it now?

szarnekow: As in: What override scenario do you have in mind?

cdietrich: exactly for this. if somebody wants to customize it .e.g cause they want to have 
```***********************************```
in the comment

cdietrich: (basicaly the old behaviour back)

szarnekow: I've doubts that someone will create an own plugin with a subtype to include that one in its build to be able to pick it up during the mwe2 workflow execution just for the sake of having `/***** ...` comments in the file header. And iff it is so very important, they can do it also right now.
In other words: If we want to have this kind of backwards compatibility, it should be a boolean flag for the "legacy" behaviour.
But as I said: This method should have have been an instance method from the beginning.

miklossy: We have just change the implementation of that method so that all starts are removed from the copyright header defined in the `fileHeader` configuration of the mwe2 workflow. I think this is a good default behaviour, but there should be a possibility to override that in client code if another behaviour is desired.

PR #1355, opened 2020-01-15
Add hashCode() and equals() to ReplaceRegion

mmews-n4:
`org.eclipse.xtext.util.ReplaceRegion` does not provide the methods `hashCode()` or `equals()`. This is a problem when using for instance `org.eclipse.xtext.ide.editor.contentassist.ContentAssistEntry`. `ContentAssistEntry` uses Xtend annotation `@EqualsHashCode` and also defines a field of type `ArrayList<ReplaceRegion>`. Since `ReplaceRegion` lacks method `hashCode()`, the hash code method of `ContentAssistEntry` is unreliable and instances of `ReplaceRegion` cannot be used e.g. in HashSets.

Please let me know if you have suggestions to improve the implementations of `hashCode()` and `equals()`.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/ReplaceRegion.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/ReplaceRegion.java
@@ -7,8 +7,6 @@
  *******************************************************************************/
 package org.eclipse.xtext.util;
 
-import java.util.Objects;
-
 /**
  * @author Jan Koehnlein - Initial contribution and API
  * @author Sebastian Zarnekow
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.util/src/org/eclipse/xtext/util/ReplaceRegion.java
```diff
@@ -56,6 +58,25 @@ public void applyTo(StringBuilder builder) {
 		builder.replace(offset, getEndOffset(), text);
 	}
 	
+	@Override
+	public boolean equals(Object object) {
+		if (object instanceof ReplaceRegion) {
+			ReplaceRegion rr = (ReplaceRegion) object;
+
+			boolean equals = true;
+			equals = equals && Objects.equals(getOffset(), rr.getOffset());
```
szarnekow: I'm not a big fan of Object.equals(..) for primitives. It causes boxing and is distracting for the reader.
How does the equals / hashcode methods look like when they are generated by Eclipse?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/ReplaceRegion.java
```diff
@@ -56,6 +58,25 @@ public void applyTo(StringBuilder builder) {
...
+			equals = equals && Objects.equals(getLength(), rr.getLength());
+			equals = equals && Objects.equals(getText(), rr.getText());
+			return equals;
+		}
+		return false;
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(getOffset(), getLength(), getText());
```
szarnekow: This builds an array of values including boxing conversion. I think the defaults as generated by Eclipse are better.

cdietrich:
thx @mmews-n4

Issue #1262, opened 2019-10-21
Endless loop in NodeIterator when start node is null


dhuebner:
We should assert `INode startWith` is not null or better fix the  hasNext()  implementation.

To reproduce:

```
import org.eclipse.xtext.nodemodel.util.NodeIterator;

public class Test {

	public static void main(String[] args) {
		new NodeIterator(null).forEachRemaining(e -> {
			System.out.println("Loop");
		});
	}

}
```

rablewis:
I'm willing to take this as my first issue.
It should be easy to fix by modifying hasNext() like this:

```java
public boolean hasNext() {
	return (lastReturned == null && startWith != null) || lastReturned.hasNextSibling();
}
```

I would propose also making a similar change to hasPrevious().

Any comments on the above approach or on appropriate testing before I proceed?

Thanks.
-R

szarnekow:
Sounds very good to me. Please go ahead.

cdietrich:
@rablewis are you still going to work on this issue?

rablewis:
I'd like to. I am having dependency issues that are preventing me from getting started. I'm going to seek advice today. I'll report back on progress tomorrow.

rablewis:
There was a failure in the Maven build stage of the Jenkins build. I don't see how this could be related to my changes, but please let me know if there's something I need to fix. Thanks. -rab

cdietrich:
hi, the ci is currently running slow and admins not available over the weekend. https://bugs.eclipse.org/bugs/show_bug.cgi?id=559314

cdietrich:
@szarnekow anything left?

szarnekow:
Fixed. Thank you @rablewis

PR #1358, opened 2020-01-19
[#1262] Prevent infinite loop with NodeIterator.

rablewis:
Throw NullPointerException if null is passed to NodeIterator constructor
to prevent iterator state where hasNext() and hasPrevious() always
return true, but getNext() and getPrevious() always return null.

Signed-off-by: Robert Lewis <rablewis@algo-pop.com>

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/nodemodel/BasicNodeIterable.java
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/nodemodel/BasicNodeIterable.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2020 Robert Lewis and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/nodemodel/BasicNodeIterable.java
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/nodemodel/BasicNodeIterable.java
@@ -1,0 +1,21 @@
+/*******************************************************************************
+ * Copyright (c) 2020 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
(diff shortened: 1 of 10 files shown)


kthoms:
I'm not sure about this. IMO this can be improved:

1. The check should be in `NodeIterable`.
2. Use `Objects.requireNonNull()` to make the check a single liner (in `NodeIterable`)
3. The check can be in `NodeIterator`, too (with `Objects.requireNonNull()`), or make `hasNext(), hasPrevious()` return `false` when no start node was given
4. (optional) Provide a unit test to cover the behavior of `NodeIterator`.

org.eclipse.xtext.tests/src/org/eclipse/xtext/nodemodel/BasicNodeIterable.java
```diff
@@ -0,0 +1,21 @@
+/*******************************************************************************
+ * Copyright (c) 2020 itemis AG (http://www.itemis.eu) and others.
```
szarnekow: The copyright header should list your name I suppose

org.eclipse.xtext.tests/src/org/eclipse/xtext/nodemodel/BasicNodeIteratorTest.java
```diff
@@ -0,0 +1,138 @@
...
+		ILeafNode gammaNode = builder.newLeafNode(text.indexOf("gamma"), "gamma".length(), gamma, false, null, root);
+		
+		return root.basicGetFirstChild();
+
+	}
+		
+	@Test(expected=NullPointerException.class)
+	public void testStartWithNullThrowsNPE() {
+		BasicNodeIterator it = new BasicNodeIterator(null);
+	}
```
szarnekow: The variable declaration is not necessary

org.eclipse.xtext.tests/src/org/eclipse/xtext/nodemodel/BasicNodeIterable.java
rablewis: "(c) 2020 Robert Lewis" or "(c) 2020 Robert Lewis and others"?
Given the work I'm building on here, it could be argued this isn't 100% my work.

cdietrich: we ususally do the and others pattern

org.eclipse.xtext.tests/src/org/eclipse/xtext/nodemodel/BasicNodeIteratorTest.java
rablewis: Removed unnecessary variable declaration here and in other files.

szarnekow:
Thank you @rablewis

PR #1359, opened 2020-01-20
[eclipse/xtext#1651] made error handling in StorageAwareResource null-safe

cdietrich:
[eclipse/xtext#1651] made error handling in StorageAwareResource null-safe


org.eclipse.xtext/src/org/eclipse/xtext/resource/persistence/StorageAwareResource.xtend
```diff
@@ -44,8 +44,8 @@ class StorageAwareResource extends LazyLinkingResource {
 				return;
 			} catch(IOException e) {
 				// revert the resource into a clean state
-				contents.clear
-				eAdapters.clear
+				contents?.clear
```
szarnekow: Both lists should never return null. Did you see some exception there?

szarnekow: Aaah getContents() vs field access ...

PR #1362, opened 2020-01-21
[eclipse/xtext#1569] Refactor Xtend to Java

kthoms:
The usage of Xtend in the Xtext code base should be reduced where the
usage of Xtend is not super beneficial. Replaced Xtend classes by the
compiled code and refactored the code to more readable Java.

This change extracts classes formerly defined in
AbstractLanguageServerTests.xtend to own files.


kthoms:
Thx @miklossy

Issue #1303, opened 2019-11-26
Javadoc incorrect for INode#getText()


tomvdbussche:
The documentation for the `getText()` method in `INode` states the returned values will never be `null`.

https://github.com/eclipse/xtext-core/blob/d9a45cfd05de3037afed498cba1029cc02bf8642/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/INode.java#L156-L161

However looking at the implementation of `AbstractNode` it seems it is possible that `null` is returned when `getRootNode()` is `null`.

https://github.com/eclipse/xtext-core/blob/d9a45cfd05de3037afed498cba1029cc02bf8642/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/AbstractNode.java#L151-L160

szarnekow:
Hey Tom, why is your root node null?

tomvdbussche:
> Hey Tom, why is your root node null?

I am not sure, it was part of a logged stacktrace but I don't know that caused the root node to be `null`.

cdietrich:
@tomvdbussche @szarnekow  how to continue with this issue

szarnekow:
I'd like to throw IllegalStateException and document it to be thrown if the text cannot be computed.

tomvdbussche:
That sounds fine to me, I was just confusing to get a null value when it is explicitly stated that a null value should never be returned.

ArneDeutsch:
I've made a short search if we (internally) use the 'null'. Most of the time we use the value without a nul check indeed, indicating that the change would be save (because a null here would most probably lead to a lot of NPEs anyway). But there are (rare) cases where we explicitely check for null (and users might do as well). Example: `org.eclipse.xtext.GrammarUtil#getTypeRefName`

```
public static String getTypeRefName(TypeRef typeRef) {
	if (typeRef.getClassifier() != null)
		return typeRef.getClassifier().getName();
	final ICompositeNode node = NodeModelUtils.getNode(typeRef);
	if (node != null) {
		final BidiIterator<INode> leafNodes = node.getAsTreeIterable().iterator();
		while (leafNodes.hasPrevious()) {
			INode previous = leafNodes.previous();
			if (previous instanceof ILeafNode && !((ILeafNode) previous).isHidden()) {
				String result = previous.getText();
				if (result != null && result.startsWith("^")) {
					result = result.substring(1);
				}
				return result;
			}
		}
	}
	return null;
}
```

[...]

PR #1365, opened 2020-01-24
[#1303] Throw exception instead of returning null.

ArneDeutsch:
Extend JavaDoc.


org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/INode.java
```diff
@@ -156,6 +156,7 @@
 	/**
 	 * Returns the parsed text that is covered by this node (including hidden tokens). The result is never <code>null</code>
 	 * but may be empty.
+	 * @throws IllegalStateException if text can't be computed because rootNode is 'null'.
```
szarnekow: I'd rephrase this to `if the text is not available. A possible reason is, that the node is not part of a valid node model`

PR #1383, opened 2020-02-11
Hn issue1382

hannesN:
I added the suggested line into the IdeProjectDescriptor and updated the two tests using the wizards.

The gradle and maven build are working.


cdietrich:
@miklossy i have the change in the xtend file? which java change do you refer to?

cdietrich:
thx for the fix @hannesN

cdietrich:
@kthoms would still be nice if you would find a few mins to have a look

kthoms:
+1 The change looks reasonable. Don't have a LS that I could test with yet.

Issue #1394, opened 2020-02-20
Cleaning up all deprecated methods/classes in the xtext.ide.server package


miklossy:
See [comment](https://github.com/eclipse/xtext-core/pull/1393#pullrequestreview-361786523)

JanKoehnlein:
I'd also want to deprecate and schedule for removal this single workspace root stuff. I'll create a PR

cdietrich:
@JanKoehnlein can you please take care about the cleanup of the single root deprecations.
i should have cleaned up everything else if nothing slipped through

JanKoehnlein:
I will, but not before Monday

Von meinem iPhone gesendet

cdietrich:
any update @JanKoehnlein

cdietrich:
Fixed in 2.22

thx @JanKoehnlein

PR #1400, opened 2020-02-24
[#1394] deprecate single WS root config and schedule it for removal

JanKoehnlein:


JanKoehnlein:
Any objections against merging this?

PR #1420, opened 2020-03-13
[lsp] Remove deprecated code for single root workspaces

JanKoehnlein:
See #1394


JanKoehnlein:
Sorry for taking so long. Vacation and illness took their toll.

JanKoehnlein:
And thanks for the review!

Issue #1427, opened 2020-03-22
Get rid of the Xtext IDEA generator


miklossy:
Since the Xtext Framework has already [dropped support](https://github.com/eclipse/xtext/issues/1278) for the IntelliJ IDEA, I think the [XtextProjectConfig.ideaPlugin](https://github.com/eclipse/xtext-core/blob/f9b6d1bebe09dc5775d45671d255384a8160015c/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/project/XtextProjectConfig.xtend#L40) field should be marked as `Deprecated`.

@cdietrich What do you think? If you agree, I can provide the corresponding PR.

szarnekow:
Please delete it instead of marking as deprecated.

miklossy:
Thanks for the PR @cdietrich ! I changed the title of this issue to better reflect to work to be done.

cdietrich:
Fixed in .22

Issue #1414, opened 2020-03-06
GeneratorNodePerformanceTest in flaky


cdietrich:
GeneratorNodePerformanceTest is flaky on xxxxx Jenkins environment

cdietrich:
seems to better recently

PR #1429, opened 2020-03-23
[#1427] ditch idea generator from xtext generator

cdietrich:
[#1427] ditch idea generator from xtext generator


miklossy:
@cdietrich Please take a look at the test failures, otherwise it look good to me.

cdietrich:
see https://github.com/eclipse/xtext-core/issues/1414 @miklossy

miklossy:
Ah ok, thanks for the hint.

PR #1434, opened 2020-03-27
provided more 'disabled' fragments to ease workflow customization

cdietrich:
provided more 'disabled' fragments to ease workflow customization

### DIFF
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/generator/GeneratorFragment2.xtend
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/generator/GeneratorFragment2.xtend
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2015, 2020 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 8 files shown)


cdietrich:
meanwile stting e.g.,
```
JunitSupport = null

```
is working, thus this is obsolete

miklossy:
@cdietrich: Do you see any places in the Xtext-Exampes where disabling the generation fragments could be defined in a more elegant/readable manner?

cdietrich:
?

miklossy:
I mean if you consider e.g the worklow definition of the Xtext Arithmetics example https://github.com/eclipse/xtext-eclipse/blob/77876803d5427677630ccff80a51fad3706ba246/org.eclipse.xtext.xtext.ui.examples/projects/arithmetics/org.eclipse.xtext.example.arithmetics/src/org/eclipse/xtext/example/arithmetics/GenerateArithmetics.mwe2#L52

the generator fragment is currently disabled. Do you think defining it using generator = null or generator = Disabled would be better readable/more explanatory?

cdietrich:
generator = null and generateStub=false is not the same
(have a look at the binding and plugin.xml of the ui project)

miklossy:
I see. Do you think that `generator = null` would be a better choice to complete disable the generator fragment in case of the Xtext Arithmetics exmaple?

cdietrich:
you know more about the example than i do. so i dont know

miklossy:
I think using `generator = null` would be a better choice there.
@szarnekow What do you think?

szarnekow:
@miklossy with `generateStub = false` you still get a preference page for the compiler by default. That does not make too much sense. `generator = null` may improve that.

miklossy:
I had the same assumption, but I was not sure about that. Thanks for clarifying that, @szarnekow !
I created the corresponding [PR](https://github.com/eclipse/xtext-eclipse/pull/1386) to clean up the Xtext SimpleArithmetics example.

Issue #1443, opened 2020-04-13
Xtext formatter behaves differently on Windows in case of different line delimiters


miklossy:
The Xtext formatter formats the Xbase grammar differently on Windows according to the used line endings. When formatting e.g the OpOther rule from the Xbase grammar with Windows line endings, there is no change, but formatting it with Linux line endings, everything will be formatted into one line:
![screencast](https://user-images.githubusercontent.com/20393472/79113313-edf89a80-7d80-11ea-8e96-f5f05ed479a4.gif)

**Hint:** I encountered this problem when regenerating the Xtext-Seven-Languages on Windows, see [PR](https://github.com/xtext/seven-languages-xtext/pull/17)

miklossy:
See also https://bugs.eclipse.org/bugs/show_bug.cgi?id=429483

miklossy:
Fixed in 2.22

PR #1444, opened 2020-04-13
[#1443] Fix XtextFormatter behaviour on different line endings.

miklossy:
- Modify the FormattingConfigBasedStream to handle the different line
endings properly.
- Implement corresponding XtextFormatterTest and
GrammarAccessExtensionsTest test cases.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting/impl/FormattingConfigBasedStream.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting/impl/FormattingConfigBasedStream.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2009, 2020 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -494,9 +494,9 @@ public class FormattingConfigBasedStream extends BaseTokenStream {
 				&& hiddenTokenHelper.isWhitespace((AbstractRule) grammarElement);
 		if (isWhitespace || cfg.getWhitespaceRule() == grammarElement) {
 			if (preservedWS == null)
-				preservedWS = value;
+				preservedWS = harmonizeLineSeparator(value);
 			else
-				preservedWS += value;
+				preservedWS += harmonizeLineSeparator(value);
 		} else
 			addLineEntry(grammarElement, value, true);
 	}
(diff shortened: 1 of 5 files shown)


miklossy:
I regenerated the Xtext-Seven-Languages with this fix and there is no changes in the *GrammarAccess.java files on Windows.

miklossy:
@szarnekow The `ILineSeparatorInformation` is already considered in the [getLineSeparator()](https://github.com/eclipse/xtext-core/blob/f9b6d1bebe09dc5775d45671d255384a8160015c/org.eclipse.xtext/src/org/eclipse/xtext/formatting/impl/FormattingConfigBasedStream.java#L366) method

cdietrich:
@miklossy @szarnekow  what is the state of this PR?

miklossy:
IMHO it is ready to merge.

org.eclipse.xtext/src/org/eclipse/xtext/formatting/impl/FormattingConfigBasedStream.java
```diff
@@ -269,7 +269,9 @@ public LineEntry(EObject grammarElement, String value, boolean isHidden, Set<Ele
 			this.isHidden = isHidden;
 			this.leadingLocators = beforeLocators;
 			this.indent = indent;
-			this.leadingWS = leadingWS;
+			if (leadingWS != null) {
```
szarnekow, line 272: I'd rather like to fix the caller of the constructor.

szarnekow:
I think the present fix is more of a workaround than a fix of the underlying problem.

org.eclipse.xtext/src/org/eclipse/xtext/formatting/impl/FormattingConfigBasedStream.java
miklossy: Done

miklossy:
> 
> 
> I think the present fix is more of a workaround than a fix of the underlying problem.

Could be true. But I think since the `formatter` package is more or less deprecated, it could be acceptable to provide a workaround rather than hunting a proper fix. Sooner or later, the [XtextFormatter](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextFormatter.java) should be migrated to the `formatter2` API infrastructure, at that point we can verify if the problem still occurs (without the workaround) and provide a proper fix there, if necessary.

szarnekow:
I'm sorry to say that I don't agree. Converting spaces with each and every LineEntry seems wrong to me. Please try to find the proper location for the fix.

miklossy:
@szarnekow I updated the PR and found a better place to harmonize the line endings:

```java
public void writeHidden(EObject grammarElement, String value) throws IOException {
	boolean isWhitespace = grammarElement instanceof AbstractRule
			&& hiddenTokenHelper.isWhitespace((AbstractRule) grammarElement);
	if (isWhitespace || cfg.getWhitespaceRule() == grammarElement) {
		if (preservedWS == null)
			preservedWS = harmonizeLineSeparator(value);
		else
			preservedWS += harmonizeLineSeparator(value);
	} else
		addLineEntry(grammarElement, value, true);
}

/**
 * @since 2.22
 */
protected String harmonizeLineSeparator(String value) {
	if (value != null) {
		return value.replaceAll("\r?\n", getLineSeparator());
	}
	return null;
}
```

PR #1460, opened 2020-04-29
[https://github.com/eclipse/xtext-eclipse/issues/1221] Serializer.

miklossy:
- Improve the serializeReplacement implementation by modifying the
ReplaceRegion length calculation so that it takes not only the
ICompositeNode.getTotalLength() into account, but considers if the new
text contains additional whitespaces and the old text is also followed
by white spaces, than the whitespaces contained by the original document
is also consumed by the quickfix.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2011, 2020 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -23,6 +23,8 @@ import org.eclipse.xtext.formatting2.regionaccess.ITextRegionAccess;
 import org.eclipse.xtext.formatting2.regionaccess.ITextReplacement;
 import org.eclipse.xtext.formatting2.regionaccess.TextRegionAccessBuilder;
 import org.eclipse.xtext.nodemodel.ICompositeNode;
+import org.eclipse.xtext.nodemodel.ILeafNode;
+import org.eclipse.xtext.nodemodel.INode;
 import org.eclipse.xtext.nodemodel.util.NodeModelUtils;
 import org.eclipse.xtext.parsetree.reconstr.ITokenStream;
 import org.eclipse.xtext.parsetree.reconstr.impl.TokenStringBuffer;
@@ -43,6 +45,7 @@ import org.eclipse.xtext.util.EmfFormatter;
 import org.eclipse.xtext.util.ReplaceRegion;
 import org.eclipse.xtext.validation.IConcreteSyntaxValidator;
 
+import com.google.common.base.CharMatcher;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 
@@ -210,7 +213,60 @@ public class Serializer implements ISerializer {
 			throw new IllegalStateException("Cannot replace an obj that has no associated node");
 		}
 		String text = serialize(obj, options);
-		return new ReplaceRegion(node.getTotalOffset(), node.getTotalLength(), text);
+		int replaceRegionLength = calculateReplaceRegionLength(node, text);
+		return new ReplaceRegion(node.getTotalOffset(), replaceRegionLength, text);
 	}
 
+	/**
+	 * @since 2.22
+	 */
+	protected int calculateReplaceRegionLength(ICompositeNode node, String text) {
+		int oldTextLength = node.getTotalLength();
+		int newTextLength = text.length();
+
+		if (newTextLength > oldTextLength) {
+			String remainingText = text.substring(oldTextLength);
+			if (isWhitespace(remainingText) && hiddenNodeFollows(node)) {
+				return newTextLength;
+			}
+		}
+		return oldTextLength;
+	}
+
+	/**
+	 * @since 2.22
+	 */
+	protected boolean hiddenNodeFollows(ICompositeNode node) {
+		INode followingNode = getFollowingNode(node);
+		if (followingNode instanceof ILeafNode) {
+			return ((ILeafNode)followingNode).isHidden();
+		}
+		return false;
+	}
+	
+	/**
+	 * Returns the node that follows the <i>node</i>, independently, if they have the same parent.
+	 * @since 2.22
+	 */
+	protected INode getFollowingNode(ICompositeNode node) {
+		if (node != null) {
+			if (node.hasNextSibling()) {
+				INode nextSibling = node.getNextSibling();
+				Iterator<ILeafNode> nextSiblingLeafNodes = nextSibling.getLeafNodes().iterator();
+				if (nextSiblingLeafNodes.hasNext()) {
+					return nextSiblingLeafNodes.next();
+				} else {
+					return getFollowingNode(node.getParent());
+				}	
+			}
+		}
+		return null;
+	}
+
+	/**
+	 * @since 2.22
+	 */
+	protected boolean isWhitespace(String text) {
+		return CharMatcher.whitespace().matchesAllOf(text);
+	}
 }
--- a/org.eclipse.xtext.testlanguages/src/org/eclipse/xtext/testlanguages/noJdt/formatting2/NoJdtTestLanguageFormatter.xtend
+++ b/org.eclipse.xtext.testlanguages/src/org/eclipse/xtext/testlanguages/noJdt/formatting2/NoJdtTestLanguageFormatter.xtend
@@ -11,11 +11,12 @@ import org.eclipse.xtext.testlanguages.noJdt.noJdt.Model
 class NoJdtTestLanguageFormatter extends AbstractFormatter2 {
 	
 	def dispatch void format(Model model, extension IFormattableDocument document) {
-		// TODO: format HiddenRegions around keywords, attributes, cross references, etc. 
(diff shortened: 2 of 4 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -210,7 +211,28 @@ public ReplaceRegion serializeReplacement(EObject obj, SaveOptions options) {
 			throw new IllegalStateException("Cannot replace an obj that has no associated node");
 		}
 		String text = serialize(obj, options);
-		return new ReplaceRegion(node.getTotalOffset(), node.getTotalLength(), text);
+		int replaceRegionLength = calculateReplaceRegionLength(node, text);
+		return new ReplaceRegion(node.getTotalOffset(), replaceRegionLength, text);
 	}
 
+	private int calculateReplaceRegionLength(ICompositeNode node, String text) {
```
szarnekow, line 218: Unfortunately this is not generally applicable, e.g. languages with synthetic tokens or semantic whitespace will break. This method must at least be protected

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -210,7 +211,28 @@ public ReplaceRegion serializeReplacement(EObject obj, SaveOptions options) {
...
+			String remainingText = text.substring(oldTextLength);
+			if (containsOnlyWhiteSpaces(remainingText) && whiteSpaceFollows(node)) {
+				return newTextLength;
+			}
+		}
+		return oldTextLength;
+	}
+
+	private boolean whiteSpaceFollows(ICompositeNode node) {
+		return node.hasNextSibling() && node.getNextSibling() instanceof HiddenLeafNode;
```
szarnekow, line 232: Checks against implementation classes of the node model must be avoided. Please use `instanceof ILeafNode && node.isHidden()` instead.

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -210,7 +211,28 @@ public ReplaceRegion serializeReplacement(EObject obj, SaveOptions options) {
...
+	}
+
+	private boolean containsOnlyWhiteSpaces(String text) {
```
szarnekow, line 235: Could be `CharMatcher.whitespac().matchesAllOf(text)`

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
szarnekow: In any case, the logic here is unfortunately flawed. If I understand the intention behind this code correctly, it must be something like `node.getLeadNodes().head().isHidden()` - though the name `whiteSpaceFollows` is misleading given that comment nodes are also hidden.

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -210,7 +211,28 @@ public ReplaceRegion serializeReplacement(EObject obj, SaveOptions options) {
...
+		if (newTextLength > oldTextLength) {
+			String remainingText = text.substring(oldTextLength);
+			if (containsOnlyWhiteSpaces(remainingText) && whiteSpaceFollows(node)) {
+				return newTextLength;
+			}
+		}
+		return oldTextLength;
+	}
+
+	private boolean whiteSpaceFollows(ICompositeNode node) {
```
szarnekow, line 231: whiteSpace -> whitespace

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
szarnekow, line 235: containsOnlyWhiteSpaces -> isWhitespace

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
miklossy: Fixed.

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
miklossy: Fixed.

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
miklossy: Fixed.

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
miklossy: Fixed.

miklossy: Fixed.

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
miklossy: > Unfortunately this is not generally applicable, e.g. languages with synthetic tokens or semantic whitespace will break.

Actually I think that it is really too dangerous to modify the code in the `xtext-core` code base. Hence it is quite difficult to provide a general solution there, I would prefer to fix the problem only on the Domainmodel example side. We can still consider a general solution later on, if lot of "client projects" claims about the problem, but for the time being a fix only for the Domainmodel example would satisfy my needs.

@cdietrich @szarnekow : If you agree, I will close this PR and provide another one in `xtext-eclipse`.

szarnekow: @miklossy I didn't want to scare you and I'm sure that you'll get this straight. If you don't mind, I'll be happy to give advice and provide feedback.

miklossy:
@szarnekow I updated the PR as requested: I added a SerializerReplacementCalculationTest test case to the xtext-core repository based on the  [NoJdtTestLanguage](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext.testlanguages/src/org/eclipse/xtext/testlanguages/noJdt/NoJdtTestLanguage.xtext).

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -210,7 +213,35 @@ public ReplaceRegion serializeReplacement(EObject obj, SaveOptions options) {
...
+				return newTextLength;
+			}
+		}
+		return oldTextLength;
+	}
+
+	protected boolean hiddenNodeFollows(ICompositeNode node) {
+		if (node.hasNextSibling()) {
+			INode nextSibling = node.getNextSibling();
+			if (nextSibling instanceof ILeafNode) {
```
szarnekow, line 236: If the next sibling is a composite node, it's first child may still be a hidden node.
If you want to catch those, too, I recommend to use something like nextSibling.getLeafNodes() and check the first node in that list.

miklossy:
Thanks for the feedback @szarnekow . I updated the PR as requested.

> There may be corner cases where the next sibling does not exist but the next node is still a hidden node (if current is the last child in its parent).
The current fix is actually strong enough to catch two different use cases (`Domainmodel example` and `NoJdtTestLanguage`), so I considered this fix already as an improvement, even if it does not cover all use cases. I would consider the corner cases if we can find a use case that is not covered by the current fix, so we can translate that use case into a test case and improve the implementation based on that. But that I would consider as future work in a separate issue.

szarnekow:
I assume that you are fully aware of the fact that Xtext is neither a framework that helps to run the domainmodel example nor a framework to build a painfully trivial test language.
If every bugfix was done under the proposition of ignoring corner cases, the framework would not be where it was now.

miklossy:
> There may be corner cases where the next sibling does not exist but the next node is still a hidden node (if current is the last child in its parent).

Is there already any `getNextNode()` method that I can use for this purpose?

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -210,7 +213,50 @@ public ReplaceRegion serializeReplacement(EObject obj, SaveOptions options) {
 			throw new IllegalStateException("Cannot replace an obj that has no associated node");
 		}
 		String text = serialize(obj, options);
-		return new ReplaceRegion(node.getTotalOffset(), node.getTotalLength(), text);
+		int replaceRegionLength = calculateReplaceRegionLength(node, text);
+		return new ReplaceRegion(node.getTotalOffset(), replaceRegionLength, text);
 	}
 
+	protected int calculateReplaceRegionLength(ICompositeNode node, String text) {
```
szarnekow, line 220: Please also add `@Since` tags to the new methods.

miklossy:
`@since 2.22` tags have been added.

org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -210,7 +213,60 @@ public ReplaceRegion serializeReplacement(EObject obj, SaveOptions options) {
...
+				}	
+			}
+		}
+		return null;
+	}
+
+	/**
+	 * @since 2.22
+	 */
+	private boolean isWhitespace(String text) {
```
szarnekow, line 269: Last thing that slipped through: This method should either be protected or the since tag is unnecessary.

miklossy: I changed the method visibility to `protected`.

Issue #269, opened 2017-02-03
[Xtext-Generator]JavaFileAccess creates unnecessary imports


cdietrich:
Consider the following Grammar

```
Model:
	test=WarnTest
;
enum WarnTest:
	A = 'a' |
	B = 'b'
;
```

this gives the following unnecessary import in the generated Grammar Access

```
import org.eclipse.xtext.service.AbstractElementFinder.AbstractEnumRuleElementFinder; // this one is not needed
import org.eclipse.xtext.service.AbstractElementFinder.AbstractGrammarElementFinder;
```

cdietrich:
we already have the `JavaFileAccess.DONT_IMPORT_NESTED_TYPES` feature. maybe applying it to the grammar access solves the problem

cdietrich:
@meysholdt do you have any comments on this issue?

miklossy:
I also encountered this problem recently:
![screenshot](https://user-images.githubusercontent.com/20393472/69157584-70887100-0ae5-11ea-8256-e31d9ed4affa.png)

cdietrich:
@miklossy are you going to look into this issue?

miklossy:
Also reproducible with the Eclipse GEF DOT [DotArrowType.xtext](https://github.com/eclipse/gef/blob/master/org.eclipse.gef.dot/src/org/eclipse/gef/dot/internal/language/DotArrowType.xtext) grammar:
![image](https://user-images.githubusercontent.com/20393472/72611586-c7b29280-392a-11ea-851b-adf6bab04d95.png)

miklossy:
Currently, the following 4 Xtext grammars of the [Eclipse GEF DOT](https://github.com/eclipse/gef/wiki/DOT-User-Guide) project are affected by this problem:
![screenshot](https://user-images.githubusercontent.com/20393472/73069728-c6dfaa80-3eae-11ea-9bf4-507ac7a5bdf5.png)

miklossy:
I also encountered this problem on the [TargetPlatform-DSL](https://github.com/eclipse-cbi/targetplatform-dsl) project.
![screenshot](https://user-images.githubusercontent.com/20393472/80958364-34cc3400-8e05-11ea-97c3-695438b307bb.png)

miklossy:
Fixed in 2.22

PR #1478, opened 2020-05-05
[#269] Regenerate languages

miklossy:


miklossy:
The fix provided by #https://github.com/eclipse/xtext-core/pull/1472 results in a lot of changes, since not only the unnecessary `import org.eclipse.xtext.service.AbstractElementFinder.AbstractEnumRuleElementFinder;` will be removed but the

```
import org.eclipse.xtext.service.AbstractElementFinder.AbstractGrammarElementFinder;
class ... extends AbstractGrammarElementFinder {
   ...
}
```
will also be refactored to 
```
import org.eclipse.xtext.service.AbstractElementFinder.AbstractGrammarElementFinder;
class ... extends AbstractGrammarElementFinder {
   ...
}
```

That's why all languages, not only languages with enum definitions are affected.

@cdietrich @szarnekow what do you think about that?

szarnekow:
Fine by me as long as we generate the test- and production languages for all repos such that nobody will be surprised by these changes later on.

miklossy:
Ok, thanks. I will create the corresponding PRs.

PR #1487, opened 2020-05-11
Update reference must convert the qualified name text

dhuebner:
In case of STRING typed crossrefs `scope.getSingleElement(oldName)` will not work properly

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ReferenceUpdater.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ReferenceUpdater.java
@@ -245,7 +245,7 @@ public class ReferenceUpdater implements IReferenceUpdater {
 		}
 		IScope scope = scopeProvider.getScope(updatable.getSourceEObject(), updatable.getEReference());
 		ISemanticRegion region = updatable.getReferenceRegion();
-		QualifiedName oldName = nameConverter.toQualifiedName(region.getText());
+		QualifiedName oldName = getQualifiedName(updatable);
 		IEObjectDescription oldDesc = scope.getSingleElement(oldName);
 		if (oldDesc != null && oldDesc.getEObjectOrProxy() == updatable.getTargetEObject()) {
 			return;


cdietrich:
@dhuebner  can you explain what is not working?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/ReferenceUpdater.java
```diff
@@ -245,7 +245,7 @@ public void updateReference(ITextRegionDiffBuilder rewriter, IUpdatableReference
 		}
 		IScope scope = scopeProvider.getScope(updatable.getSourceEObject(), updatable.getEReference());
 		ISemanticRegion region = updatable.getReferenceRegion();
-		QualifiedName oldName = nameConverter.toQualifiedName(region.getText());
+		QualifiedName oldName = getQualifiedName(updatable);
```
cdietrich, line 248: i am not sure if this changes semantics when there are two scope elements with the same qualified but different names (e.g. 1 imported, 1 not)

dhuebner: Should not. It's more about getting qnames right (using default delimiter) :

`module "my.module" {} `

The old code would produce `"my` `module"`  qname , but should be `my` `module`

cdietrich:
ahh getQualifiedName actually calls linkingHelper

szarnekow:
Could you please add a test for this scenario?

dhuebner:
> Could you please add a test for this scenario?

@szarnekow I knew you'd say that. I have no idea where to place such a test. We will probably need a new language which uses something like `ref = [Rule|STRING]`  `Rule: name = STRING ;` 
Or do we have a test language which uses STRING to reference other rules?

szarnekow:
Let me grep this for you:
```\[\s*\w+\\s*|\s*STRING\s*\]``` in `*.xtext` yields `Xtext.xtext` among others.

szarnekow:
How is this PR related to #1490 ?

dhuebner:
@szarnekow I don't know. @JanKoehnlein ?

JanKoehnlein:
It's a similar area, but not directly related. #1490 was about the pre-rename of LSP that prohibited renaming quoted identifiers, this one is about applying the reference update of a STRING syntaxed cross-ref. 

AFAICS, #1490 should also work for STRING based FQNs.

szarnekow:
@dhuebner Any plan to proceed on this one. Next week is M3 and the current week has a holiday in Germany. Do you think you can provide a test along with this patch?

dhuebner:
@szarnekow I started generating a testlanguage and than noticed that the `updateReference()` method is overwritten in test ReferenceUpdater (org.eclipse.xtext.ide.tests.testlanguage.ide.serializer.TestLanguageReferenceUpdater).
Also I need time to figure out how to test that the implementation inside the method body takes the right path. The error occurred in my project because the implementation created QNames with `"` other things worked somehow.
Overall I will need more time to add a test. At the moment I can't say when I will find time to proceed.

cdietrich:
@dhuebner  i also tried to reproduce this in a plain project using a value converter for
QualifiedName and a model like ` ref /*x*/foo/*y*/ . /*z*/Foo`
and failed to do so as `findValidName(updatable, scope)` seems to find a nice new name

dhuebner:
@cdietrich 
> @dhuebner i also tried to reproduce this in a plain project using a value converter for
> QualifiedName and a model like ` ref /*x*/foo/*y*/ . /*z*/Foo`
> and failed to do so as `findValidName(updatable, scope)` seems to find a nice new name

Right, but the implementation is obviously bogus and can produce follow up errors. So what I could suggest is to apply the patch as is or wait till I, or someone else, can find time and a **good** way to test ist.

szarnekow:
@dhuebner Do you have any plans to proceed with the work on the change?

dhuebner:
> @dhuebner Do you have any plans to proceed with the work on the change?

Not really. Not enough time at the moment

cdietrich:
any update here?

PR #1490, opened 2020-05-11
[lsp] properly rename quoted identifiers

### LINKED ISSUE #1488: [lsp] rename fails for quoted identifiers

JanKoehnlein:
Fixes #1488

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
@@ -28,6 +28,7 @@ import org.eclipse.lsp4j.jsonrpc.messages.Either
 import org.eclipse.xtend.lib.annotations.Accessors
 import org.eclipse.xtext.CrossReference
 import org.eclipse.xtext.RuleCall
+import org.eclipse.xtext.conversion.IValueConverterService
 import org.eclipse.xtext.ide.refactoring.IRenameStrategy2
 import org.eclipse.xtext.ide.refactoring.RenameChange
 import org.eclipse.xtext.ide.refactoring.RenameContext
@@ -65,6 +66,8 @@ class RenameService2 implements IRenameService2 {
 
 	@Inject TokenUtil tokenUtil
 
+	@Inject IValueConverterService valueConverterService 
+	
 	Function<EObject, String> attributeResolver = SimpleAttributeResolver.newResolver(String, 'name')
 
 	override rename(Options options) {
@@ -197,11 +200,11 @@ class RenameService2 implements IRenameService2 {
 					if (element !== null && !element.eIsProxy) {
 						val leaf = NodeModelUtils.findLeafNodeAtOffset(rootNode, candidateOffset)
 						if (leaf !== null && leaf.isIdentifier) {
-							val leafText = NodeModelUtils.getTokenText(leaf)
+							val leafText = getConvertedValue(leaf.grammarElement, leaf)
 							val elementName = element.elementName
 							if (!leafText.nullOrEmpty && !elementName.nullOrEmpty && leafText == elementName) {
 								val start = document.getPosition(leaf.offset)
-								val end = document.getPosition(leaf.offset + elementName.length)
+								val end = document.getPosition(leaf.endOffset)
 								return Either.forLeft(new Range(start, end))
 							}
 						}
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
```diff
@@ -219,6 +222,14 @@ class RenameService2 implements IRenameService2 {
 		return null
 	}
 
+	protected def String getConvertedValue(EObject grammarElement, ILeafNode leaf) {
+		switch (grammarElement) {
+			RuleCall: valueConverterService.toValue(leaf.text, grammarElement.rule.name, leaf).toString()
+			CrossReference: getConvertedValue(grammarElement.terminal, leaf)
```
cdietrich, line 228: shouldt we consult `linkingHelper.getRuleNameFrom(updatable.getCrossReference())` e.g. as in referenceupdater.getQualifiedName

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
```diff
@@ -197,11 +200,11 @@ class RenameService2 implements IRenameService2 {
 					if (element !== null && !element.eIsProxy) {
 						val leaf = NodeModelUtils.findLeafNodeAtOffset(rootNode, candidateOffset)
```
szarnekow, line 201: Probably beyond this small bugfix here, but the rename infrastructure appears to be based on the assumption that IDs are comprised of leaf nodes and ignores the fact that `MyId: ID | 'someKeyword';` or `VersionedId: ID '-''(' INT ')';` are valid identifiers, too.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
```diff
@@ -197,11 +200,11 @@ class RenameService2 implements IRenameService2 {
...
 						if (leaf !== null && leaf.isIdentifier) {
-							val leafText = NodeModelUtils.getTokenText(leaf)
+							val leafText = getConvertedValue(leaf.grammarElement, leaf)
```
szarnekow, line 203: Given that this is already quite a bit of complexity in the logic, please change the variable name to better reflect what it is now.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
```diff
@@ -219,6 +222,14 @@ class RenameService2 implements IRenameService2 {
 		return null
 	}
 
+	protected def String getConvertedValue(EObject grammarElement, ILeafNode leaf) {
+		switch (grammarElement) {
+			RuleCall: valueConverterService.toValue(leaf.text, grammarElement.rule.name, leaf).toString()
```
szarnekow, line 227: `toValue()` on a leaf node could throw a value converter exception if the leaf node was used in a data type rule e.g.
`IntAsID: INT;` with `999999999999999999`.
The value converter exception is not caught / handled in `doPrepareRename`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
szarnekow: Maybe even use `org.eclipse.xtext.linking.impl.LinkingHelper.getCrossRefNodeAsString(INode, boolean)` and find a proper cross ref node starting at the leaf node.

JanKoehnlein: As you wish.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
JanKoehnlein: The truth is, it doesn't even work with qualified names (as a datatype rule), unless you have registered a special value converter. My goal was to cover the basic behavior of the default ID rule. I think the method `doPrepareRename` is small enough to be overridden for more specific use cases.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
JanKoehnlein: Done_

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.xtend
JanKoehnlein: Would you really expect rename to work OOTB when integers are used as names? 
Anyway, I added some code to handle this.

JanKoehnlein:
Thanks for the feedback.

cdietrich:
@JanKoehnlein you missed the generated code https://github.com/eclipse/xtext-core/pull/1493

PR #1516, opened 2020-06-03
[lsp] avoid negative positions

### LINKED ISSUE #1515: [lsp] bad position for exception issues
When an exception is thrown in the validator, the diagnostic`CompositeEValidator.createExceptionDiagnostic(String, Object, Throwable)` will lead to an issue with position (0,0) which will be transformed to an position (-1,-1) which causes VS Code to stop the LS.

JanKoehnlein:
Fixes #1515


kittaakos:
> I wonder if we should also fix the ExceptionDiagnostic to produce correct 1-based line/column information.

I agree.

Also from the [issue](https://github.com/eclipse/xtext-core/issues/1515#issue-629918634):

>  which causes VS Code to stop the LS.

It would be nice to see the error or figure out why it stops the LS. Can we have the same issue with any other non-valid text document positions?

JanKoehnlein:
>I wonder if we should also fix the ExceptionDiagnostic to produce correct 1-based line/column information.

Wouldn't that be a breaking change? Note that this is also used in all non-LSP contexts.
If not, shouldn't we rather set the init value to -1 for these fields in `Issue`?

PR #1539, opened 2020-07-05
[ls] copy build listener list before notification

### LINKED ISSUE #1538: [ls] Allow adding/removing build listeners concurrently
Two issues here
- `ILanguageServerAccess` has a method to add a build listener but none to remove it
- `WorkspaceManager` traverses the list of listeners instead of a copy, which is prone to concurrent modification exceptions

JanKoehnlein:
Fixes #1538

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.java
@@ -15,6 +15,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.URI;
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.java
```diff
@@ -258,7 +258,7 @@ protected void setWorkspaceConfig(IWorkspaceConfig workspaceConfig) {
 	 * Callback after a build was performend
 	 */
 	protected void afterBuild(List<IResourceDescription.Delta> deltas) {
-		for (ILanguageServerAccess.IBuildListener listener : buildListeners) {
+		for (ILanguageServerAccess.IBuildListener listener : new ArrayList<>(buildListeners)) {
```
cdietrich, line 261: should we use a safe datastructure as ListenerList or CopyOnWriteArraySet as we do it at other places? what do you think @szarnekow ?

szarnekow: Yes, I think a CopyOnWriteArrayList (preserving semantics) or CopyOnWriteArraySet would be better to avoid unnecessary copying and also be save wrt adding listeners from read commands.

JanKoehnlein:
Thanks for the feedback. Better like this?

PR #1542, opened 2020-07-13
Add message to IndexOutOfBoundsException in InternalNodeModelUtils

tivervac:
An exception without any message is not all that useful

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/InternalNodeModelUtils.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/InternalNodeModelUtils.java
@@ -15,6 +15,7 @@ import java.util.List;
 import org.eclipse.xtext.nodemodel.INode;
 import org.eclipse.xtext.util.LineAndColumn;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 
 /**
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/InternalNodeModelUtils.java
```diff
@@ -58,7 +58,7 @@ protected static LineAndColumn getLineAndColumn(INode anyNode, int documentOffse
 	 */
 	protected static LineAndColumn getLineAndColumn(String text, int[] lineBreaks, int offset) {
 		if (offset > text.length() || offset < 0) {
-			throw new IndexOutOfBoundsException();
+			throw new IndexOutOfBoundsException("Offset (" + offset + ") is longer than the text's length (" + text.length() + ")");
```
cdietrich, line 61: might miss negative offsets

szarnekow: Maybe use `com.google.common.base.Preconditions.checkElementIndex`?

szarnekow: Well in fact it would need to be checkPositionIndex.

tivervac: Done

PR #1549, opened 2020-08-05
API cleanup after Xtend2Java

szarnekow:
The methods in question had a return type `Object` but the return value was never used.
Updated to return `void` instead.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/persistence/ResourceStorageLoadable.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/persistence/ResourceStorageLoadable.java
@@ -83,9 +83,8 @@ public class ResourceStorageLoadable {
 		}.loadResource(resource);
 	}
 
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/resource/persistence/ResourceStorageLoadable.java
```diff
@@ -83,9 +83,8 @@ public InternalEObject loadEObject() throws IOException {
 		}.loadResource(resource);
 	}
```
cdietrich, line 85: i assume we do this api breakage intentionally

szarnekow: Yes, it was intentionally. If you think it's not worth it, I'll abandon the change.

cdietrich: is not api anyway

PR #1556, opened 2020-08-17
fixed warnings when compiling against Java 11

cdietrich:
fixed warnings when compiling against Java 11


cdietrich:
@szarnekow do you remember why we use new in the first place?

PR #1560, opened 2020-08-26
RegionDiffFormatter may drop essential changes at the boundary of a change region #1559

stephan-herrmann:
The new test fails without a code change: it misses to add the additional newline as requested by the change in PartialSerializationTestLanguageFormatter.

I addressed the issue inside the RegionDiffFormatter in a conservative way, i.e., trying to precisely capture the situation and then by modifying the text replacement from "\n" -> "\n\n" to "" -> "\n" and adjusting the offset correspondingly. After this change, the text replacement fits into the region under consideration and can be successfully applied.

A colleague still reported an issue on windows, where "\r\n" -> "\r\n\r\n" isn't yet correctly handled. I'll come back here, once we've debugged the situation on windows.

Signed-off-by: Stephan Herrmann <stephan.herrmann@berlin.de>

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/serializer/ChangeSerializerTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/serializer/ChangeSerializerTest.xtend
@@ -506,6 +506,25 @@ class ChangeSerializerTest {
 	}
 
 	@Test
+	def void testAddElementsWithEmptyLineUnix() {
+		val oldLineSeparator = System.getProperty("line.separator")
+		System.setProperty("line.separator", "\n")
+		try {
+			testAddElementsWithEmptyLine()
+		} finally {
+			System.setProperty("line.separator", oldLineSeparator)
+		}	
+	}
+	@Test
+	def void testAddElementsWithEmptyLineWindows() {
+		val oldLineSeparator = System.getProperty("line.separator")
+		System.setProperty("line.separator", "\r\n")
+		try {
+			testAddElementsWithEmptyLine()
+		} finally {
+			System.setProperty("line.separator", oldLineSeparator)
+		}
+	}
 	def void testAddElementsWithEmptyLine() {
 		val uri = "inmemory:/file-add.pstl"
 		val fs = new InMemoryURIHandler()
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/serializer/ChangeSerializerTest.xtend
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/serializer/ChangeSerializerTest.xtend
@@ -504,4 +504,32 @@ class ChangeSerializerTest {
 			0 3 "#23" -> "#23 subs A A2"
 		'''
 	}
+
+	@Test
+	def void testAddElementsWithEmptyLine() {
+		val uri = "inmemory:/file-add.pstl"
+		val fs = new InMemoryURIHandler()
+		fs += uri -> '''
+		#1 {
+			N1;
+		}
+		'''
+
+		val rs = fs.createResourceSet
+		val model = rs.contents(uri, Node)
+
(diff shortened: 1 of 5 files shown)


stephan-herrmann:
The windows case looks ugly: `ITextReplacement.getText()` and `ITextSegment.getText()` have unix line ends, and only `ITextReplacement.getReplacementText()` contains windows line ends.

I want to trim the replacement text corresponding to what we find in `ITextReplacement.getText()`, but simple character counting doesn't seem to work.

Does the formatter framework have any utilities for dealing with mixed line ends?

At a closer look we may need another prefix-check anyway: does the text to be cut-off also match the replacement text?

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/serializer/ChangeSerializerTest.xtend
```diff
@@ -506,6 +506,25 @@ class ChangeSerializerTest {
...
+		try {
+			testAddElementsWithEmptyLine()
+		} finally {
+			System.setProperty("line.separator", oldLineSeparator)
+		}	
+	}
+	@Test
+	def void testAddElementsWithEmptyLineWindows() {
+		val oldLineSeparator = System.getProperty("line.separator")
+		System.setProperty("line.separator", "\r\n")
```
stephan-herrmann, line 521: Naive attempt at testing in unix & windows modes, it seems, however, that the formatter preferences are cached and hence the first value wins for all tests. Ergo, running tests individually works as designed, running the suite still only tests one mode.
Feel free to suggest a way how individual tests can use different line separator settings ...

stephan-herrmann:
As a github noob I didn't succeed to link the PR to issue eclipse/xtext#2497 - can this be done after the fact?

stephan-herrmann:
The proposed change is a bit convoluted, simply because I tried to create minimal impact on existing behavior.

Further testing, however, indicates that perhaps `SequentialRegionDiff.getRegion(..)` is indeed the culprit (see also description eclipse/xtext#2497). The only reason I didn't propose to change that method is, the most probable change causes regressions in serializer tests. Perhaps s.o. could look at those regressions to check if it's OK to adjust test expectations?

[org.eclipse.xtext.ide.tests.serializer 20200828-104600.xml.gz](https://github.com/eclipse/xtext-core/files/5141187/org.eclipse.xtext.ide.tests.serializer.20200828-104600.xml.gz)

* On my machine the failure in testChangeInXML happens always, independent of my changes
* DIfferences in the two new tests look morally OK to me (change includes the previous "\n" without changing it)

Here's the **much simpler change** which produced the above test results:
[SequentialRegionDiff.patch.gz](https://github.com/eclipse/xtext-core/files/5141218/SequentialRegionDiff.patch.gz)

cdietrich:
still need to think about this / diff into it

stephan-herrmann:
> still need to think about this / diff into it

Thanks. Do you want me to create another pull request for the alternative, simpler change/diff attached in my previous comment? I'd update all affected tests so you can easily see the effect of it.

cdietrich:
i first have to understand the problem. this is what i mean with "thinking about it"

cdietrich:
@szarnekow would be nice if you could have a look too

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/serializer/ChangeSerializerTest.xtend
cdietrich: i also observed this problem. either there is a bug with caching or we should use unix endings all over the place.

cdietrich: the caching problem is also there when the cached value is guessed by looking at a file and this file has non-typical endings

cdietrich:
branch needs rebasing

cdietrich:
needs rebasing. will drop target milestone

cdietrich:

PR #1574, opened 2020-09-16
use dev bom in tycho build

cdietrich:
use dev bom in tycho build


szarnekow (approved):
Thanks for fixing this!

PR #1587, opened 2020-10-11
update to latest jetty

cdietrich:
update to latest jetty


szarnekow (approved):
Makes sense.

PR #1592, opened 2020-10-18
fixed platform targetplatform badness with Java 11 in wizard

kthoms:


kthoms:
Created accidently

PR #1597, opened 2020-10-24
Semantic quickfix (generic IDE integration)

imhotep82:
This PR adds declarative quickfix support for LSP (loosely based on the Eclipse implementation). This is my first PR, so it would be nice if anyone could give me hints what is missing.
If I did something wrong, please just let me know. Hope this can be useful. Cheers, Heiner. 😄


cdietrich:
Hi, i will try to have a more detailed look tomorrow
But here are some general remarks

- Please mark new api classes with `@since 2.24`
- please avoid to write new code in Xtend if it is not necessary (in your case even the test would be perfectly fine in java)
- java code style: no finals in parameters and method bodies if not needed
- why no code action service 2 baseclass or delegate  in your code and Test

imhotep82:
Thanks @cdietrich for the quick reply. I'll rewrite the classes. I have a base class for the CodeActionService here. Will commit it. Cheers.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/QuickFixInjectorProvider.java
```diff
@@ -0,0 +1,39 @@
...
+	// overridden to avoid NPE in org.eclipse.xtext.resource.impl.LiveShadowedResourceDescriptions.getAllResourceDescriptions()
+	@Override
+	public Injector internalCreateInjector() {
+		return new FileAwareTestLanguageIdeSetup() {
+			@Override
+			public Injector createInjector() {
+				return Guice.createInjector(Modules2.mixin(new FileAwareTestLanguageRuntimeModule(), new FileAwareTestLanguageIdeModule() {
+					@Override
+					public void configureIResourceDescriptionsLiveScope(Binder binder) {
+						// do not override runtime binding
```
cdietrich, line 32: Why this

imhotep82: The Unit test failed with an NPE because of the IDE implementation. I used the runtime binding and resolved it with it. I can attach a stacktrace of the failing test. If I can resolve it another way I'll try to.

cdietrich: please do so

imhotep82: Will push a commit that fixes the NPE

cdietrich: i assume the problem is that you dont test on full lsp level and thus no ChunkedResourceDescriptions
are created. so you might have to change that or do something similar as others tests do

```
	protected def EObject createInMemoryFile(CharSequence content, EClass type) {
.....
// might also go to org.eclipse.xtext.ide.tests.serializer.ChangeSerializerTestHelper.createResourceSet(InMemoryURIHandler), dont know why it is not there
		val Map<String, ResourceDescriptionsData> dataMap = new HashMap();
		dataMap.put("egal", ResourceSetAdapter.findResourceDescriptionsData(rs));
		new ChunkedResourceDescriptions(Collections.unmodifiableMap(dataMap), rs); 
		return rs.contents(fileName, type.instanceClass) as EObject

```

this might also be a LiveShadowedChunkedResourceDescriptions vs LiveShadowedResourceDescriptions
problem (latter is used in normal changeserializer tests vs lsp/ide usually usss LiveShadowedChunkedResourceDescriptions and thus ChunkedResourceDescriptions to be installed (usually project and workspace manager care about this

imhotep82: Thanks for the hints. I'll consider that fix.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.xtend
```diff
@@ -0,0 +1,174 @@
...
+	@Inject
+	extension QuickFixTestHelper
+	@Inject
+	extension FileExtensionProvider
+
+	/**
+	 * Test that the expected quickfixes are offered on a given validation issue in a given DSL text.
+	 * 
+	 * @param it The initial DSL text.
+	 * @param issueCode The code of the validation issue to that the offered quickfixes are to test.
```
tivervac, line 52: s/that the/the/

s/are to/to/

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.xtend
```diff
@@ -0,0 +1,174 @@
...
+		DiagnosticResolution actualIssueResolution, String originalText, Document doc) {
+		val edit = actualIssueResolution.apply
+		val changedDocument = doc.applyChanges(edit.changes.values.flatMap[toList])
+
+		assertEquals(expectedResult, changedDocument.contents)
+	}
+
+	@Data
+	protected static class Quickfix {
+
```
tivervac, line 167: Nitpick probably don't need any empty lines in this `@Data`, it's supposed to be very concise

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/QuickFixTestHelper.xtend
```diff
@@ -0,0 +1,27 @@
...
+package org.eclipse.xtext.ide.tests.editor.quickfix
+
+import com.google.common.base.Preconditions
+import org.eclipse.emf.common.util.URI
+import org.eclipse.emf.ecore.resource.ResourceSet
+import org.eclipse.xtext.ide.tests.serializer.ChangeSerializerTestHelper
+
+class QuickFixTestHelper extends ChangeSerializerTestHelper {
+
+	//TODO resource text marker <|> like in Proposal tests
```
tivervac, line 18: 👍

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/QuickFixTestHelper.xtend
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.editor.quickfix
+
+import com.google.common.base.Preconditions
+import org.eclipse.emf.common.util.URI
+import org.eclipse.emf.ecore.resource.ResourceSet
+import org.eclipse.xtext.ide.tests.serializer.ChangeSerializerTestHelper
+
+class QuickFixTestHelper extends ChangeSerializerTestHelper {
```
tivervac, line 16: Xtext is trying to reduce the amount of xtend code in the codebase. This class doesn't gain too much from being xtend, could you write it in Java instead?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.inject.Provider;
+
+public class AbstractDeclarativeIdeQuickfixProvider implements IQuickFixProvider {
+
+	private static final Logger LOG = Logger.getLogger(AbstractDeclarativeIdeQuickfixProvider.class);
+
+	@Inject
+	private Provider<DiagnosticResolutionAcceptor> issueResolutionAcceptorProvider;
+
+	private Predicate<Method> getFixMethodPredicate(final String issueCode) {
```
tivervac, line 32: No need for final arguments or final parameters, they don't offer much value and make the code harder to read. There's quite a few more in the code

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -0,0 +1,95 @@
...
+
+	private Options options;
+
+	public DiagnosticResolution(final String label, final Factory modificationContextFactory,
+			final IModification<EObject> modification) {
+		this(label, modificationContextFactory, modification, 0);
+	}
+
+	/**
+	 * @since 2.22
```
tivervac, line 44: Aren't we 2.24 already?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolutionAcceptor.java
```diff
@@ -0,0 +1,40 @@
...
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.xtext.ide.serializer.IChangeSerializer.IModification;
+import org.eclipse.xtext.ide.server.codeActions.ICodeActionService2.Options;
+
+import com.google.common.collect.Lists;
+import com.google.inject.Inject;
+
+public class DiagnosticResolutionAcceptor {
+
+	private final List<DiagnosticResolution> issues = Lists.newArrayList();
```
tivervac, line 22: Might as well just use `new ArrayList<>();`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/QuickFixes.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation Container for @{@link QuickFix} annotations. You can mark a method in a {@link AbstractDeclarativeIdeQuickfixProvider} with this annotation to
+ * specify which issue codes the method offers quick-fixes for.
+ *
```
tivervac, line 19: Abundant empty line

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/QuickFixes.java
```diff
@@ -0,0 +1,27 @@
...
+package org.eclipse.xtext.ide.editor.quickfix;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation Container for @{@link QuickFix} annotations. You can mark a method in a {@link AbstractDeclarativeIdeQuickfixProvider} with this annotation to
+ * specify which issue codes the method offers quick-fixes for.
```
tivervac, line 18: Hmm, I only see `@QuickFix` being mentioned, not `@QuickFixes`, am I missing something?

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/QuickFixTestHelper.xtend
imhotep82: will do

imhotep82:
@cdietrich : just a quick note: I extended the fileAware IDE module for the LSP integration test. I hope this is OK. I needed "some" test-language to do the test with. I saw that there's a more generic test language, but this already has a binding for CodeActions. If there's a better way, please let me know. Cheers

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,223 @@
...
+import org.eclipse.xtext.xbase.lib.IterableExtensions;
+import org.eclipse.xtext.xbase.lib.ObjectExtensions;
+import org.eclipse.xtext.xbase.lib.Pair;
+import org.eclipse.xtext.xbase.lib.Procedures.Procedure1;
+import org.junit.Assert;
+
+import com.google.inject.Inject;
+
+/**
+ * @author Heinrich Weichert
```
cdietrich, line 59: add `@since 2.24`

org.eclipse.xtext.ide/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,4 +1,7 @@
 eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.8
```
cdietrich, line 2: undo this change

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -0,0 +1,98 @@
...
+	private int relevance;
+
+	private Options options;
+
+	public DiagnosticResolution(String label, Factory modificationContextFactory, IModification<EObject> modification) {
+		this(label, modificationContextFactory, modification, 0);
+	}
+
+	/**
+	 * @since 2.24
```
cdietrich, line 48: `@since 2.24` on the class is enhough

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.google.common.base.Predicate;
+import com.google.common.base.Strings;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.inject.Inject;
+import com.google.inject.Provider;
+
+/**
+ * @since 2.24
```
cdietrich, line 27: please move the `@since`s after the `@auther` (check also other places)

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,223 @@
...
+ */
+public abstract class AbstractIdeQuickfixTest {
+
+	protected static class Quickfix {
+		private final String label;
+		private final String description;
+		private final String result;
+
+		public Quickfix(final String label, final String description, final String result) {
```
cdietrich, line 68: please remove final from parameters (for fields it is ok)

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,223 @@
...
+	 */
+	public void testQuickfixesOn(final String fileContents, final String issueCode, final EClass type, final AbstractIdeQuickfixTest.Quickfix... quickfixes) {
+		quickfixesAreOffered(createInMemoryFile(fileContents, type), issueCode, fileContents, quickfixes);
+	}
+
+	@SuppressWarnings("unchecked")
+	private <T> T createInMemoryFile(final CharSequence content, final EClass type) {
+		InMemoryURIHandler fs = new InMemoryURIHandler();
+		String _primaryFileExtension = this.fileExtensionProvider.getPrimaryFileExtension();
+		final String fileName = "inmemory:/file1." + _primaryFileExtension;
```
cdietrich, line 118: remove finals from variables

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/QuickFixTestHelper.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.google.common.collect.Iterators;
+
+/**
+ * 
+ * @author Heinrich Weichert
+ */
+public class QuickFixTestHelper extends ChangeSerializerTestHelper {
+  @Override
+  public <T extends Object> T contents(final ResourceSet rs, final String fileName, final Class<T> type) {
```
cdietrich, line 24: finals

org.eclipse.xtext.testlanguages.ide/src/org/eclipse/xtext/testlanguages/fileAware/ide/quickfix/FileAwareTestLanguageQuickFixProvider.java
```diff
@@ -0,0 +1,28 @@
...
+		private IssueCodes() {
+			// utility
+		}
+
+		public static final String TESTISSUECODE = "test";
+		public static final String TESTLABEL = "Change element name to first upper";
+	}
+
+	@QuickFix(IssueCodes.TESTISSUECODE)
+	public void removeUnusedVariable(final DiagnosticResolutionAcceptor acceptor) {
```
cdietrich, line 21: finals

org.eclipse.xtext.testlanguages/src/org/eclipse/xtext/testlanguages/fileAware/FileAwareTestLanguageRuntimeModule.java
```diff
@@ -17,16 +17,18 @@
  * Use this class to register components to be used at runtime / without the Equinox extension registry.
  */
 public class FileAwareTestLanguageRuntimeModule extends AbstractFileAwareTestLanguageRuntimeModule {
-	
-	public void configureIResourceDescriptionsLiveScope(Binder binder) {
+
+	@Override
+	public void configureIResourceDescriptionsLiveScope(final Binder binder) {
```
cdietrich, line 22: finals

org.eclipse.xtext.testlanguages/src/org/eclipse/xtext/testlanguages/fileAware/validation/FileAwareTestLanguageValidator.java
```diff
@@ -3,23 +3,20 @@
...
-//		if (!Character.isUpperCase(greeting.getName().charAt(0))) {
-//			warning("Name should start with a capital",
-//					FileAwareTestLanguagePackage.Literals.GREETING__NAME,
-//					INVALID_NAME);
-//		}
-//	}
-	
+
+	@Check
+	public void checkElementNameStartsWithUpperCase(final Element element) {
```
cdietrich, line 17: finals

org.eclipse.xtext.testlanguages/src/org/eclipse/xtext/testlanguages/fileAware/validation/FileAwareTestLanguageValidator.java
```diff
@@ -3,23 +3,20 @@
...
+		if (!Character.isUpperCase(element.getName().charAt(0))) {
+			warning("Element name should start with a capital", element, null, "test");
```
cdietrich, line 19: no constant for issue code

org.eclipse.xtext.testlanguages.ide/src/org/eclipse/xtext/testlanguages/fileAware/ide/quickfix/FileAwareTestLanguageQuickFixProvider.java
```diff
@@ -0,0 +1,28 @@
...
+import org.eclipse.xtext.xbase.lib.StringExtensions;
+
+public class FileAwareTestLanguageQuickFixProvider extends AbstractDeclarativeIdeQuickfixProvider {
+
+	public static class IssueCodes {
+		private IssueCodes() {
+			// utility
+		}
+
+		public static final String TESTISSUECODE = "test";
```
cdietrich, line 16: move to validator

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,223 @@
...
+				final ResourceSet resourceSet = options.getResource().getResourceSet();
+				return resourceSet;
+			}
+		});
+		CodeActionParams codeActionParams = new CodeActionParams();
+		codeActionParams.setRange(new Range(startPos, endPos));
+		codeActionParams.setTextDocument(new TextDocumentIdentifier(target.eResource().getURI().toString()));
+	
+		CodeActionContext codeActionContext = new CodeActionContext();
+		codeActionContext.setDiagnostics(Collections.unmodifiableList(CollectionLiterals.newArrayList(issue)));
```
cdietrich, line 194: use new ArrayList<>() and remove the Unmodifiable list

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,223 @@
...
+import org.eclipse.xtext.resource.XtextResource;
+import org.eclipse.xtext.resource.impl.ChunkedResourceDescriptions;
+import org.eclipse.xtext.resource.impl.ResourceDescriptionsData;
+import org.eclipse.xtext.testing.util.InMemoryURIHandler;
+import org.eclipse.xtext.util.CancelIndicator;
+import org.eclipse.xtext.util.LineAndColumn;
+import org.eclipse.xtext.util.Strings;
+import org.eclipse.xtext.xbase.lib.CollectionLiterals;
+import org.eclipse.xtext.xbase.lib.IterableExtensions;
+import org.eclipse.xtext.xbase.lib.ObjectExtensions;
```
cdietrich, line 51: check for unused imports

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,223 @@
...
+	 *            </ol>
+	 */
+	public void testQuickfixesOn(final String fileContents, final String issueCode, final EClass type, final AbstractIdeQuickfixTest.Quickfix... quickfixes) {
+		quickfixesAreOffered(createInMemoryFile(fileContents, type), issueCode, fileContents, quickfixes);
+	}
+
+	@SuppressWarnings("unchecked")
+	private <T> T createInMemoryFile(final CharSequence content, final EClass type) {
+		InMemoryURIHandler fs = new InMemoryURIHandler();
+		String _primaryFileExtension = this.fileExtensionProvider.getPrimaryFileExtension();
```
cdietrich, line 117: please inline the "_" local vars

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,223 @@
...
+		final ResourceSet rs = this.quickFixTestHelper.createResourceSet(fs);
+		final Map<String, ResourceDescriptionsData> dataMap = new HashMap<>();
+		dataMap.put("project", ResourceDescriptionsData.ResourceSetAdapter.findResourceDescriptionsData(rs));
+		new ChunkedResourceDescriptions(dataMap, rs);
+		return (T) this.quickFixTestHelper.contents(rs, fileName, type.getInstanceClass());
+	}
+
+	private void quickfixesAreOffered(final EObject target, final String issueCode, final String originalText,
+			final AbstractIdeQuickfixTest.Quickfix... expected) {
+		final List<AbstractIdeQuickfixTest.Quickfix> expectedSorted = IterableExtensions.sortBy(Arrays.asList(expected), it -> it.label);
```
cdietrich, line 131: remove finals

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
```diff
@@ -0,0 +1,92 @@
...
+ * 
+ * @author Heinrich Weichert
+ */
+public class AbstractDeclarativeIdeQuickfixProvider implements IQuickFixProvider {
```
cdietrich, line 31: i wonder if we should mark this as `@Beta`. that do you think @szarnekow

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixCodeActionService.java
```diff
@@ -0,0 +1,65 @@
...
+ * {@link ICodeActionService2} handling quick-fixes annoted with {@link CodeActionKind#QuickFix}.
+ * 
+ * @author Heinrich Weichert
+ * 
+ * @see AbstractDeclarativeIdeQuickfixProvider
+ * @see QuickFix
+ * 
+ * @since 2.24
+ */
+public class QuickFixCodeActionService implements ICodeActionService2 {
```
cdietrich, line 37: i wonder if we should mark this as `@Beta`. that do you think @szarnekow

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolutionAcceptor.java
```diff
@@ -0,0 +1,46 @@
...
+package org.eclipse.xtext.ide.editor.quickfix;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.xtext.ide.serializer.IChangeSerializer.IModification;
+import org.eclipse.xtext.ide.server.codeActions.ICodeActionService2.Options;
+
+import com.google.common.collect.Lists;
```
cdietrich, line 18: unused import

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/QuickFixCodeActionServiceTest.java
```diff
@@ -0,0 +1,62 @@
...
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.server;
+
+import org.eclipse.xtext.ide.server.ServerModule;
+import org.eclipse.xtext.ide.server.concurrent.RequestManager;
+import org.eclipse.xtext.testing.AbstractLanguageServerTest;
+import org.eclipse.xtext.testing.AbstractLanguageServerTest.DirectRequestManager;
```
cdietrich, line 14: unused imports

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolutionAcceptor.java
imhotep82: Thanks. I noticed a lot of .project and .classpath files. Do you generally gitignore eclipse prefs? Can I add/enhance save actions? I'll do an organize imports over my files anyway.

cdietrich: nope. they should be checked in. 
but as we are in transition to Java 11 such
discrepancies as yours may happen unfortunately

org.eclipse.xtext.testlanguages.ide/src/org/eclipse/xtext/testlanguages/fileAware/ide/quickfix/FileAwareTestLanguageQuickFixProvider.java
```diff
@@ -0,0 +1,28 @@
...
+	public static class IssueCodes {
+		private IssueCodes() {
+			// utility
+		}
+
+		public static final String TESTLABEL = "Change element name to first upper";
+	}
+
+	@QuickFix(FileAwareTestLanguageValidator.TESTISSUECODE)
+	public void removeUnusedVariable(final DiagnosticResolutionAcceptor acceptor) {
```
cdietrich, line 21: remove finals

imhotep82: @cdietrich see my other comment: If I move this into the testlang, this class will be removed. I'll check if there are any finals left, thanks.

cdietrich:
@imhotep82 can you please squash your commits into 1

cdietrich:
please also check codeactiontest.testcodeaction

imhotep82:
> please also check codeactiontest.testcodeaction

done, thanks. the test needed to be adapted because testlang now produces two code actions

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.java
```diff
@@ -41,13 +42,13 @@
...
  */
-public class CodeActionService implements ICodeActionService2 {
+public class CodeActionService extends QuickFixCodeActionService {
 	@Inject
 	private IChangeSerializer serializer;
 
 	@Override
 	public List<Either<Command, CodeAction>> getCodeActions(ICodeActionService2.Options options) {
-		List<Either<Command, CodeAction>> actions = new ArrayList<>();
+		List<Either<Command, CodeAction>> actions = super.getCodeActions(options);
```
cdietrich, line 51: shouldnt the to upper case which mimics the quickfix be dropped here?

cdietrich: ahh no, that one is a command

imhotep82: What do you mean? This accumulates the quickfix and the hard-coded code-action in `org.eclipse.xtext.ide.tests.testlanguage.ide.server.CodeActionService.fixInvalidName(Diagnostic, Options)`.
This leads to two actions being returned for the same issue ID. One with the type QuickFix from the super implementation. What would you expect?

cdietrich: both things do a to upper case.
do we need both.
if i read your quickfixcode action service correct it does a code acction not a command?

cdietrich: so maybe we should change it to

command.setTitle("Make '" + string + "' upper case (Command)");

cdietrich: but the impl is bogus as there is no edit command in
TestLanguageExecutableCommandService

cdietrich: => have no idea why the current impl is as it is

imhotep82: Ok, I'll just drop the "old" one and use the quickfix action instead in the tests. Then we have a real implementation test.

cdietrich: i wonder if we want to have a testcase for command code action anyway thus having both would be better.
even if its not completely implemented

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.java
```diff
@@ -41,39 +42,22 @@
...
-				actions.add(Either.forLeft(fixInvalidName(d, options)));
-			} else if (TestLanguageValidator.UNSORTED_MEMBERS.equals(code)) {
+			if (TestLanguageValidator.UNSORTED_MEMBERS.equals(code)) {
 				actions.add(Either.forRight(fixUnsortedMembers(d, options)));
 			}
 		}
 		return actions;
 	}
 
-	private Command fixInvalidName(Diagnostic d, ICodeActionService2.Options options) {
```
cdietrich, line 62: please restore. otherwise we would have code action as command untested.

imhotep82: Ok, then I'll do as you suggested

cdietrich: i assume the test (framework) can handle multiple results. correct?

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/CodeActionTest.xtend
```diff
@@ -32,6 +32,12 @@ class CodeActionTest extends AbstractTestLangLanguageServerTest {
 				edit : changes :
 				    MyModel.testlang : Foo [[0, 5] .. [0, 8]]
 				documentChanges : 
+				command : my.textedit.command
```
cdietrich, line 35: :+1:

cdietrich:
i already did a short discussion with @szarnekow 
can you please add the `@Beta` annotation to the two places mentioned.
he will also do a review when he finds the time later this week.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/QuickFixCodeActionServiceTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.eclipse.xtext.util.Modules2;
+import org.eclipse.xtext.xbase.lib.Procedures.Procedure1;
+import org.junit.Test;
+
+import com.google.inject.Module;
+
+/**
+ * @author Heinrich Weichert
+ */
+public class QuickFixCodeActionServiceTest extends AbstractTestLangLanguageServerTest {
```
cdietrich, line 22: i assume this class can be dropped/is redundant now?

imhotep82: I basically tests the same stuff, so yes :D

imhotep82:
> @imhotep82 can you please squash your commits into 1



> i already did a short discussion with @szarnekow
> can you please add the `@Beta` annotation to the two places mentioned.
> he will also do a review when he finds the time later this week.

Will do

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.java
imhotep82: yes

szarnekow (changes requested):
I can't seem to find any reference to the LSP CodeActionContext.only property which could be used to disable / ignore the code actions that are quickfixes.

The default implementation and the test suggest that the quickfix is implemented by a code action based on edits rather than commands. This implies that all possible fixes are computed with a dry run. If the client does not apply these, the work was done for nothing. I'd expect that many quickfixes are more involving. Can we change the default implementation such that it establishes a protocol based on Commands instead? This would make it more comparable to what we have in Eclipse, too.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,220 @@
...
+		issue.setRange(new Range(startPos, endPos));
+
+		ICodeActionService2.Options options = new ICodeActionService2.Options();
+		options.setCancelIndicator(CancelIndicator.NullImpl);
+		options.setDocument(new Document(Integer.valueOf(0), Strings.toUnixLineSeparator(originalText)));
+		options.setResource((XtextResource) target.eResource());
+		options.setLanguageServerAccess(new ILanguageServerAccess() {
+			@Override
+			public void addBuildListener(ILanguageServerAccess.IBuildListener listener) {
+				//
```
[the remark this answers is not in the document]
cdietrich: it might be called.but we dont seem to care here

cdietrich:
to me it looks like the only flag was never implemented here

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticModificationContext.java
```diff
@@ -0,0 +1,67 @@
...
+import org.eclipse.xtext.ide.server.rename.ChangeConverter2;
+
+import com.google.inject.ImplementedBy;
+
+/**
+ * @author Heinrich Weichert
+
+ * @since 2.24
+ */
+class DiagnosticModificationContext {
```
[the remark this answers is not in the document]
imhotep82: Agreed, I'll make the surrounding class public

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/QuickFixTestHelper.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.google.common.collect.Iterators;
+
+/**
+ * 
+ * @author Heinrich Weichert
+ */
+class QuickFixTestHelper extends ChangeSerializerTestHelper {
+  @Override
+  public <T extends Object> T contents(ResourceSet rs, String fileName, Class<T> type) {
```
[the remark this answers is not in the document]
imhotep82: It's overridden, I'll add a custom method that's more "speaking"

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,220 @@
...
+	 *            The quickfixes that are expected to be offered on the given <code>issueCode</code>. Each expected quickfix should be described by the
+	 *            following triple:
+	 *            <ol>
+	 *            <li>the quickfix label</li>
+	 *            <li>the quickfix description</li>
+	 *            <li>the DSL text after the quickfix application</li>
+	 *            </ol>
+	 */
+	public void testQuickfixesOn(String fileContents, String issueCode, EClass type, AbstractIdeQuickfixTest.Quickfix... quickfixes) {
+		quickfixesAreOffered(createInMemoryFile(fileContents, type), issueCode, fileContents, quickfixes);
```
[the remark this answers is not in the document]
imhotep82: I'll do that here, pack it into a variable and pass it to the in-mem resource and method

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,220 @@
...
+			}
+
+			@Override
+			public LanguageClient getLanguageClient() {
+				return null;
+			}
+
+			@Override
+			public ResourceSet newLiveScopeResourceSet(URI uri) {
+				ResourceSet resourceSet = options.getResource().getResourceSet();
```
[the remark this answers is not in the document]
imhotep82: `//re-using the existing ResourceSet because it contains the URI protocol mapping for "inmemory" resources.` perhaps?

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -0,0 +1,220 @@
...
+import com.google.inject.Inject;
+
+/**
+ * @author Heinrich Weichert
+ * 
+ * @since 2.24
+ */
+public abstract class AbstractIdeQuickfixTest {
+
+	protected static class Quickfix {
```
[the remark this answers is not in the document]
imhotep82: I can rename that, yes

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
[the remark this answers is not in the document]
imhotep82: Added the throw

imhotep82:
> I can't seem to find any reference to the LSP CodeActionContext.only property which could be used to disable / ignore the code actions that are quickfixes.
> 
> The default implementation and the test suggest that the quickfix is implemented by a code action based on edits rather than commands. This implies that all possible fixes are computed with a dry run. If the client does not apply these, the work was done for nothing. I'd expect that many quickfixes are more involving. Can we change the default implementation such that it establishes a protocol based on Commands instead? This would make it more comparable to what we have in Eclipse, too.

I fear you lost me there. What command would I use? Is there any example for this?
I added a line to filter the quickfix commands if they are not in the white-list in CodeActionsContext.
`boolean handleQuickfixes = options.getCodeActionParams().getContext().getOnly().contains(CodeActionKind.QuickFix);`

cdietrich:
guess something like 

boolean handleQuickfixes = options.getCodeActionParams().getContext().getOnly().contains(CodeActionKind.QuickFix.name); // only is list of string

@szarnekow what do you think / thought of?

cdietrich:
(i am not sure what the semantic of an empty or null list is

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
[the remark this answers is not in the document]
szarnekow: :+1:

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/QuickFixTestHelper.java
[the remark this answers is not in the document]
szarnekow: :+1:

szarnekow:
> > I can't seem to find any reference to the LSP CodeActionContext.only property which could be used to disable / ignore the code actions that are quickfixes.
> > The default implementation and the test suggest that the quickfix is implemented by a code action based on edits rather than commands. This implies that all possible fixes are computed with a dry run. If the client does not apply these, the work was done for nothing. I'd expect that many quickfixes are more involving. Can we change the default implementation such that it establishes a protocol based on Commands instead? This would make it more comparable to what we have in Eclipse, too.
> 
> I fear you lost me there. What command would I use? Is there any example for this?

No, there is no such example. I'd prefer to send the CodeAction along with a data object that carries enough information to compute the edits lazily, e.g. a command that knows the diagnostic and the issue resolution. When the client executes the command, the edits would be computed only on demand.

> I added a line to filter the quickfix commands if they are not in the white-list in CodeActionsContext.
> `boolean handleQuickfixes = options.getCodeActionParams().getContext().getOnly().contains(CodeActionKind.QuickFix);`

:+1:
As @cdietrich pointed out, beware nulls / empty lists which should imply that quickfixes are desired.

imhotep82:
> The default implementation and the test suggest that the quickfix is implemented by a code action based on edits rather than commands. This implies that all possible fixes are computed with a dry run. If the client does not apply these, the work was done for nothing. I'd expect that many quickfixes are more involving. Can we change the default implementation such that it establishes a protocol based on Commands instead? This would make it more comparable to what we have in Eclipse, too.

cdietrich:
i am not sure about that. commands are so "heavy weight".
i assume you have to declare a command for all quickfixes then clientside

@szarnekow can you elaborate?

cdietrich:
@imhotep82 can you also check the test fails

imhotep82:
> @imhotep82 can you also check the test fails

thanks, stupid typo

szarnekow:
> i am not sure about that. commands are so "heavy weight".
> i assume you have to declare a command for all quickfixes then clientside
> 
> @szarnekow can you elaborate?
> (did not try this is vscode yet, for normal command you have to)
> 
> or would we work with a generic quickfix command
> and stuff everything else into the args?

I don't follow. Commands are essentially JSON objects that are sent to the client and the client can apply these by sending them as they are back to the server. There is nothing to be declared on the client side.

imhotep82:
> > i am not sure about that. commands are so "heavy weight".
> > i assume you have to declare a command for all quickfixes then clientside
> > @szarnekow can you elaborate?
> > (did not try this is vscode yet, for normal command you have to)
> > or would we work with a generic quickfix command
> > and stuff everything else into the args?
> 
> I don't follow. Commands are essentially JSON objects that are sent to the client and the client can apply these by sending them as they are back to the server. There is nothing to be declared on the client side.

In `org.eclipse.xtext.ide.server.commands.ExecutableCommandRegistry` there is a Map of command-IDs to commands. Do we need to "register" our command there?

szarnekow:
Step backwards: What do you think about keeping the logic as is in this PR and discuss a potential change of the design in a subsequent, isolated ticket / pull request? @cdietrich @imhotep82

cdietrich:
@szarnekow i mean this one
https://github.com/itemis/xtext-languageserver-example/blob/f2291241f2a1e97c8ebab8187e2d30460923dee1/vscode-extension-self-contained/package.json#L46
when do i need to define these, when not

szarnekow:
These are only necessary if you have a command without any context, e.g. something that can be invoked from the command palette. A code-action has a field of type Command. Theses commands don't need to be registered on the client.

imhotep82:
> These are only necessary if you have a command without any context, e.g. something that can be invoked from the command palette. A code-action has a field of type Command. Theses commands don't need to be registered on the client.



> Step backwards: What do you think about keeping the logic as is in this PR and discuss a potential change of the design in a subsequent, isolated ticket / pull request? @cdietrich @imhotep82

Yes. Good idea. Thanks.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Quickfix for code actions in files.<br>
+ * The quickfix itself is applied to a temporary resource.<br>
+ * The textual edit can be composed by calling DiagnosticResolution#apply().
+ *
+ * @author Heinrich Weichert
+ * 
+ * @since 2.24
+ */
+public interface IQuickFixProvider {
```
cdietrich, line 25: add `@Beta`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/QuickFix.java
```diff
@@ -0,0 +1,30 @@
...
+ * annotation.
+ * 
+ * @author Heinrich Weichert
+ *
+ * @since 2.24
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD })
+@Repeatable(QuickFixes.class)
+public @interface QuickFix {
```
cdietrich, line 28: add `@Beta`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/QuickFixes.java
```diff
@@ -0,0 +1,31 @@
...
+ * {@link AbstractDeclarativeIdeQuickfixProvider} with this annotation to specify which issue codes the method offers
+ * quick-fixes for. 
+
+ * @author Heinrich Weichert
+ * 
+ * @since 2.24
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD })
+public @interface QuickFixes {
```
cdietrich, line 27: add `@Beta`

cdietrich:
@imhotep82  with the last minor findings this looks ok to me

@szarnekow @tivervac  what do you think?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/QuickFixes.java
imhotep82: done

szarnekow (approved):

cdietrich:
many thanks @imhotep82

imhotep82:
> many thanks @imhotep82

Thank you, that was fun.

Issue #1566, opened 2020-09-04
Update to next LSP4J release (likely 0.10.0)


cdietrich:
Update to next LSP4J release (likely 0.10.0)

cdietrich:
there are also some new features (semantic tokens) in this release

cdietrich:
we dont plan to implement any of the new features for now

cdietrich:
Non feature update done. for feature update we have separate issue

PR #1611, opened 2020-11-05
[eclipse/xtext-core#1566] update to LSP4J 0.10.0

cdietrich:
[eclipse/xtext-core#1566] update to LSP4J 0.10.0

### DIFF
--- a/org.eclipse.xtext.ide.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.ide.tests/META-INF/MANIFEST.MF
@@ -14,8 +14,8 @@ Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.eclipse.core.runtime;bundle-version="3.13.0",
  org.eclipse.xtext.ide,
  org.junit;bundle-version="4.12.0",
- org.eclipse.lsp4j;bundle-version="[0.9.0,0.10.0)";resolution:=optional,
- org.eclipse.lsp4j.jsonrpc;bundle-version="[0.9.0,0.10.0)";resolution:=optional,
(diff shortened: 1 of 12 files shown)


cdietrich:
TODO: how to deal with deprecated "deprecated" on DocumentSymbol

org.eclipse.xtext.ide.tests/META-INF/MANIFEST.MF
```diff
@@ -14,8 +14,8 @@ Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.eclipse.core.runtime;bundle-version="3.13.0",
  org.eclipse.xtext.ide,
  org.junit;bundle-version="4.12.0",
- org.eclipse.lsp4j;bundle-version="[0.9.0,0.10.0)";resolution:=optional,
- org.eclipse.lsp4j.jsonrpc;bundle-version="[0.9.0,0.10.0)";resolution:=optional,
```
szarnekow, line 18: We should keep 0.9.0 as lower bound, don’t you think so?

cdietrich: am not sure about this. if so we cant implement any of the new api

szarnekow: As long as we don’t implement any of these, there shouldn’t be a problem

cdietrich: ok, but when i will live with the deprecations for symbol.depreacted for now
its hard to circumvent them

szarnekow (approved):

Issue #1489, opened 2020-05-11
Migrate the Xtext formatter to the formatter2 API 


miklossy:
The [XtextFormatter](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextFormatter.java) and the corresponding test cases should be migrated to the formatter2 API infrastructure.

cdietrich:
we should have a java api for the new formatter first

miklossy:
See also https://bugs.eclipse.org/bugs/show_bug.cgi?id=405184

> Xtext still uses the old formatter, this should be changed also.

ewillink:
Not "should" be changed, "must" be changed. The new infrastructure does not support the old Formatter consequently ALL primary Xtext related tools that use a formatter MUST be migrated well before the old infrastructure can be deleted.

miklossy:
> The new infrastructure does not support the old Formatter
@ewillink It is possible to use the `old formatter` with the `new workflow`, currently, the Eclipse GEF project also uses that combination, see the [GenerateDot.mwe2](https://github.com/eclipse/gef/blob/master/org.eclipse.gef.dot/src/org/eclipse/gef/dot/internal/language/GenerateDot.mwe2) and [DotFormatter](https://github.com/eclipse/gef/blob/master/org.eclipse.gef.dot/src/org/eclipse/gef/dot/internal/language/formatting/DotFormatter.java) files.

ewillink:
Thanks, although GenerateDot.mwe2 is commented as supporting the old Formatter, it doesn't but it provides the critical FragmentAdapter clue (using deprecated classes).

	fragment = org.eclipse.xtext.generator.adapter.FragmentAdapter {
		fragment = org.eclipse.xtext.generator.formatting.FormatterFragment {}
	}

ArneDeutsch:
I will do this as a first example for the java API for the formatter and to spot additional improvements to the API.

ewillink:
Please be aware that there may soon be a new Serializer/Formatter option. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=564265

I have this working for 5 OCL editors and am just polishing the Idioms DSL, prior to writing Xtext.idioms and Terminals.idioms to demonstrate its utility for Xtext itself.

[...]

PR #1624, opened 2020-11-20
[#1489] Several formatter fixes.

ArneDeutsch:
Added unit tests as well.

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/formatting2/internal/JavaFormatterGrammarTest.java
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/formatting2/internal/JavaFormatterGrammarTest.java
@@ -1,0 +1,200 @@
+/*******************************************************************************
+ * Copyright (c) 2020 itemis AG (http://www.itemis.com) and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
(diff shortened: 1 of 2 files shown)


cdietrich:
looks better now, but there are still flaws.
should we handle them in follow up PRs?

org.eclipse.xtext.tests/src/org/eclipse/xtext/formatting2/internal/JavaFormatterGrammarTest.java
```diff
@@ -0,0 +1,200 @@
...
+		// @formatter:off
+		assertFormattedGrammar(
+				"AssignmentExpression <In, Yield> returns Expression:" + NL +
+				TAB + "<Yield> YieldExpression<In>" + NL +
+				TAB + "| RelationalExpression<In,Yield> (=> ({AssignmentExpression.lhs=current} op='=') rhs=AssignmentExpression<In,Yield>)?" + NL +
+				";"
+				,
+				"AssignmentExpression <In, Yield> returns Expression:" + NL +
+				TAB + "<Yield> YieldExpression<In>" + NL +
+				TAB + "| RelationalExpression<In, Yield> (=> ({AssignmentExpression.lhs=current} op='=')" + NL +
```
szarnekow, line 135: It would be really great to get rid of the newline in this example before the last `rhs=AssignmentExpression<In, Yield>`

szarnekow (commented):
Very nice improvements. Thank you for the effort. Somehow I'm not sure yet about the formatGroupElements, though.
Let's see how we can make even more progress from here.

cdietrich (approved):

Issue #1517, opened 2020-06-04
Fix findings from conversion Xtend->Java


szarnekow:
This will break if the region spans multiple lines. If the endOffset if larger than the document lenght, end should simply point the the last position in the document.

_Originally posted by @szarnekow in https://github.com/eclipse/xtext-core/pull/1507_

loradd:
Hi all, can I work on this bug?

szarnekow:
Sure, please go ahead

PR #1645, opened 2020-12-10
Fixes eclipse/xtext-core#1517

loradd:
Signed-off-by: Lorenzo Addazi <lorenzo.addazi@mdh.se>


szarnekow (approved):
Thank you for the fix.

Issue #1625, opened 2020-11-23
Cleanup Xtext formatter (Java)


cdietrich:
Findings from formatting grammar accesses

https://github.com/eclipse/xtext-eclipse/pull/1586
https://github.com/eclipse/xtext-core/pull/1616
https://github.com/eclipse/xtext-extras/pull/665
https://github.com/eclipse/xtext-xtend/pull/1140
https://github.com/eclipse/xtext-web/pull/272

https://github.com/eclipse/xtext-core/pull/1624/files#r527848735

https://github.com/eclipse/xtext-xtend/pull/1140/files#r528538843

https://github.com/eclipse/xtext-core/pull/1616/files#r528535270

extra spaces: https://github.com/eclipse/xtext-eclipse/pull/1586/files#r524969980

https://github.com/eclipse/xtext-web/pull/272/files#r524985690

https://github.com/eclipse/xtext-extras/pull/665#discussion_r528543795

https://github.com/eclipse/xtext-extras/pull/665#discussion_r528544302

https://github.com/eclipse/xtext-extras/pull/665#discussion_r528544070

to be discussed: https://github.com/eclipse/xtext-eclipse/pull/1586/files#r524971356

serializer bug ?!?: return keyword is missing https://github.com/eclipse/xtext-eclipse/pull/1586/files#r524965534

szarnekow:
Findings from xtext-extras that we need to look at:
https://github.com/eclipse/xtext-extras/pull/665/files#r525025543
https://github.com/eclipse/xtext-extras/pull/665/files#r528554752
https://github.com/eclipse/xtext-extras/pull/665/files#r528555084
https://github.com/eclipse/xtext-extras/pull/665/files#r528556142
https://github.com/eclipse/xtext-extras/pull/665/files#r528557245
https://github.com/eclipse/xtext-extras/pull/665/files#r528558082

szarnekow:
Findings from xtext-eclipse that we need to look at:
https://github.com/eclipse/xtext-eclipse/pull/1586/files#r528546666
https://github.com/eclipse/xtext-eclipse/pull/1586/files#r524969813
https://github.com/eclipse/xtext-eclipse/pull/1586/files#r524969980
https://github.com/eclipse/xtext-eclipse/pull/1586/files#r524971356
https://github.com/eclipse/xtext-eclipse/pull/1586/files#r524972856
https://github.com/eclipse/xtext-eclipse/pull/1586/files#r528546666
[...]

PR #1646, opened 2020-12-11
[#1625] Make test less strict to avoid failing because of bad serialize

ArneDeutsch:

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/formatting2/internal/JavaFormatterGrammarTest.java
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/formatting2/internal/JavaFormatterGrammarTest.java
@@ -29,7 +29,6 @@ import com.google.inject.Injector;
 public class JavaFormatterGrammarTest extends AbstractXtextTests {
 
 	private static final String NL = System.lineSeparator();
-	private static final String TAB = "\t";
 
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.tests/src/org/eclipse/xtext/formatting2/internal/JavaFormatterGrammarTest.java
```diff
@@ -508,7 +508,33 @@ public void beeLanguage() {
...
+				"// comment before Assignment" + NL +
+				"  bars+= /* comment in assignment */ Bar // comment after assignment" + NL +
+				"// comment before keywod" + NL +
+				"'foo';" + NL +
+				"Bar: 'bar';"
+				,
+				"generate foo \"http://foo.net/foo\"" + NL +
+				""  + NL +
+				"Foo: // comment in Foo " + NL +
+				"// comment before Assignment" + NL +
```
cdietrich, line 527: will would prefer to see tabs inside the strings
than as constant

ArneDeutsch: Done.

cdietrich (approved):
approved when tests still are green

Issue #1618, opened 2020-11-17
Enhancements to Java Formatter2 API


cdietrich:
Proposal by @szarnekow  (if i got it correct)

define delegate methods like

```
	protected void oneSpace(IHiddenRegionFormatter hrf) {
		hrf.oneSpace();
	}
```

so that one can write

```
		doc.prepend(equalsSignEqualsSignKeyword, this::oneSpace);
		doc.append(equalsSignEqualsSignKeyword, this::oneSpace);
```

instead of 

```
		doc.prepend(equalsSignEqualsSignKeyword, it -> it.oneSpace());
		doc.append(equalsSignEqualsSignKeyword, it -> it.oneSpace());
```
cc @ArneDeutsch

cdietrich:
possible example (from statemachine.web)
```
/*******************************************************************************
 * Copyright (c) 2015, 2020 itemis AG (http://www.itemis.eu) and others.
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License 2.0 which is available at
 * http://www.eclipse.org/legal/epl-2.0.
 *
 * SPDX-License-Identifier: EPL-2.0
 *******************************************************************************/
package org.eclipse.xtext.web.example.statemachine.formatting2;

import org.eclipse.xtext.formatting2.AbstractJavaFormatter;
import org.eclipse.xtext.formatting2.IFormattableDocument;
import org.eclipse.xtext.formatting2.IHiddenRegionFormatter;
import org.eclipse.xtext.formatting2.regionaccess.ISemanticRegion;
import org.eclipse.xtext.web.example.statemachine.services.StatemachineGrammarAccess;
import org.eclipse.xtext.web.example.statemachine.statemachine.Command;
import org.eclipse.xtext.web.example.statemachine.statemachine.Condition;
import org.eclipse.xtext.web.example.statemachine.statemachine.Event;
import org.eclipse.xtext.web.example.statemachine.statemachine.Signal;
import org.eclipse.xtext.web.example.statemachine.statemachine.State;
import org.eclipse.xtext.web.example.statemachine.statemachine.Statemachine;
import org.eclipse.xtext.web.example.statemachine.statemachine.Transition;

import com.google.inject.Inject;

public class StatemachineFormatter extends AbstractJavaFormatter {

	@Inject
	private StatemachineGrammarAccess ga;

	protected void format(Statemachine statemachine, IFormattableDocument doc) {
		for (Signal signal : statemachine.getSignals()) {
[...]

PR #1650, opened 2020-12-18
[#1618] Delegate methods to simplify java formatter code.

ArneDeutsch:

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/AbstractJavaFormatter.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/AbstractJavaFormatter.java
@@ -49,20 +49,20 @@ import com.google.common.annotations.Beta;
  * 	}
  * 
  * 	protected void _format(Parent parent, IFormattableDocument doc) {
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/formatting2/AbstractJavaFormatter.java
```diff
@@ -177,4 +177,70 @@ protected IEObjectRegion regionForEObject(EObject semanticElement) {
 		return textRegionExtensions.semanticRegions(semanticElement);
 	}
 
+	// implementations that delegate the methods of IHiddenRegionFormatter to allow e.g. 'this::noSpace' instead of 'it -> it.noSpace()'.
+
+	/**
+	 * Configure autowrap.
```
cdietrich, line 183: we should add `@since 2.25`s here and to the other methods

cdietrich:
in general: looks good to me. @szarnekow what do you think?

szarnekow (approved):
Pending the missing `@since` tags, I like the change. Approved.

cdietrich (approved):

Issue #1648, opened 2020-12-14
Junit5 InjectionExtension is broken in  Xtext 2.24.0


msun68:
ValidationTestHelper does not invoke Custom Validation methods. The custom validation example below worked on 2.23.0, but not 2.24.0. The checkGreetingStartsWithCapital  does not seem be called.

```
@Check
def checkGreetingStartsWithCapital(Greeting greeting) {
	if (!Character.isUpperCase(greeting.name.charAt(0))) {
		warning('Name should start with a capital', 
				MyDslPackage.Literals.GREETING__NAME,
				INVALID_NAME)
	}
}
```

The assertNoIssues below is passed on 2.24.0
```
val result = parseHelper.parse('''
	Hello xtext!
''')
validationTestHelper.assertNoIssues(result)
```

cdietrich:
maybe there is a change in emf that stopps the validator from beeing called

cdietrich:
@merks do have any idea if a change in emf could have caused this?

cdietrich:
seems also to fail in a maven project

cdietrich:
but it does not in domainmodel example

cdietrich:
is a problem with junit5 only.
the makeCopyOfGlobalState is called after creating the registry again

cdietrich:
here is the stacktrace for the 2nd

```
GlobalRegistries.makeCopyOfGlobalState() line: 67	
MyDslInjectorProvider.getInjector() line: 29	
MyDslInjectorProvider.setupRegistry() line: 65	
InjectionExtension$RegistryReset.<init>(IRegistryConfigurator) line: 65	
573487274.apply(Object) line: not available	
ExtensionValuesStore.lambda$getOrComputeIfAbsent$4(Function, Object) line: 86	
249177573.get() line: not available	
ExtensionValuesStore$MemoizingSupplier.get() line: 205	
ExtensionValuesStore$StoredValue.evaluate() line: 182	
ExtensionValuesStore$StoredValue.access$100(ExtensionValuesStore$StoredValue) line: 171	
ExtensionValuesStore.getOrComputeIfAbsent(Namespace, K, Function<K,V>) line: 89	
ExtensionValuesStore.getOrComputeIfAbsent(Namespace, K, Function<K,V>, Class<V>) line: 93	
NamespaceAwareStore.getOrComputeIfAbsent(K, Function<K,V>, Class<V>) line: 61	
[...]

PR #1651, opened 2020-12-18
[#1648] fixed regression in InjectionExtension

cdietrich:
[#1648] fixed regression in InjectionExtension


cdietrich:
@msun68 @frankbenoit would really love if you could test this to make sure we dont introduce more/different regression and the original regression is fixed

szarnekow (approved):

Issue #1626, opened 2020-11-23
Dont use Formatter in Grammar Access


cdietrich:
currently we use Formatter in Grammar Access
maybe its better to use the original text from
the original grammar with trimmed leading whitespace

cdietrich:
TODO: regenerate all languages

cdietrich:
fixed in 2.25

PR #1653, opened 2020-12-18
[#1626] Use node model to generate comments with original format.

ArneDeutsch:

### DIFF
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/generator/GrammarAccessExtensions2Test.java
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/generator/GrammarAccessExtensions2Test.java
@@ -70,19 +70,20 @@ public class GrammarAccessExtensions2Test {
 				"	| '?:';\n";
 		String expected = 
 				"//OpOther:" + NL +
-				"//	'->'" + NL +
-				"//	| '..<'" + NL +
-				"//	| '>' '..'" + NL +
-				"//	| '..'" + NL +
-				"//	| '=>'" + NL +
-				"//	| '>' (=> ('>' '>') | '>') | '<' (=> ('<' '<') | '<' | '=>') | '<>'" + NL +
-				"//	| '?:';";
+				"//      '->'" + NL +
+				"//    | '..<'" + NL +
+				"//    | '>' '..'" + NL +
+				"//    | '..'" + NL +
+				"//    | '=>'" + NL +
+				"//    | '>' (=>('>' '>') | '>')" + NL +
+				"//    | '<' (=>('<' '<') | '<' | '=>')" + NL +
+				"//    | '<>'" + NL +
+				"//    | '?:';";
 		firstRuleIsConvertedTo(grammar, expected);
 	}
 	
 	@Test
 	public void testGrammarFragmentToString2() throws Exception {
-		String NL = System.lineSeparator();
 		String grammar =
 				"grammar org.xtext.example.mydsl.MyDsl hidden (ML_COMMENT)\n" +
 				"import 'http://www.eclipse.org/emf/2002/Ecore' as ecore\n" +
@@ -91,8 +92,7 @@ public class GrammarAccessExtensions2Test {
 				"terminal ML_COMMENT: '/*'(!'*')->'*/';\n" +
 				"terminal ID: '^'?('a'..'z'|'A'..'Z'|'_') ('a'..'z'|'A'..'Z'|'_'|'0'..'9')*;";
 		String expected = 
-				"//terminal ML_COMMENT:" + NL
-				+ "//	'/*' !'*'->'*/';";
+				"//terminal ML_COMMENT: '/*'(!'*')->'*/';";
 		secondRuleIsConvertedTo(grammar, expected);
 	}
 
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
@@ -8,9 +8,7 @@
  *******************************************************************************/
 package org.eclipse.xtext.xtext.generator.grammarAccess
 
-import com.google.common.collect.Maps
 import com.google.inject.Binder
-import com.google.inject.Guice
 import com.google.inject.Inject
 import java.util.ArrayList
 import java.util.List
@@ -29,16 +27,17 @@ import org.eclipse.xtext.Grammar
 import org.eclipse.xtext.GrammarUtil
 import org.eclipse.xtext.Group
 import org.eclipse.xtext.Keyword
+import org.eclipse.xtext.Parameter
 import org.eclipse.xtext.ParserRule
 import org.eclipse.xtext.RuleCall
 import org.eclipse.xtext.TypeRef
 import org.eclipse.xtext.UnorderedGroup
 import org.eclipse.xtext.XtextRuntimeModule
 import org.eclipse.xtext.formatting.ILineSeparatorInformation
+import org.eclipse.xtext.nodemodel.util.NodeModelUtils
 import org.eclipse.xtext.resource.SaveOptions
 import org.eclipse.xtext.serializer.ISerializer
 import org.eclipse.xtext.xtext.RuleNames
-import org.eclipse.xtext.xtext.generator.CodeConfig
 import org.eclipse.xtext.xtext.generator.XtextGeneratorNaming
 import org.eclipse.xtext.xtext.generator.model.TypeReference
 import org.eclipse.xtext.xtext.generator.parser.antlr.AntlrGrammarGenUtil
@@ -48,7 +47,8 @@ import static extension java.lang.Character.*
 import static extension org.eclipse.xtext.EcoreUtil2.*
 import static extension org.eclipse.xtext.GrammarUtil.*
 import static extension org.eclipse.xtext.xtext.generator.parser.antlr.AntlrGrammarGenUtil.*
-import org.eclipse.xtext.Parameter
+import com.google.common.base.Strings
+import org.eclipse.xtext.xtext.generator.CodeConfig
 
 /**
  * This API can be used by other templates to generate code
@@ -66,11 +66,9 @@ class GrammarAccessExtensions {
 		"\t" -> "tab",
 		"\\" -> "backslash"
 	};
-	val Map<String, ISerializer> xtextSerializerByLineDelimiter = Maps.newHashMapWithExpectedSize(2)
-	
-	@Inject CodeConfig codeConfig
 	
 	@Inject extension XtextGeneratorNaming
+	@Inject CodeConfig codeConfig
 	
 	/**
 	 * Returns a reference to the GrammarAccess implementation for a grammar.
@@ -396,8 +394,59 @@ class GrammarAccessExtensions {
 		}
 	}
 	
-	def String grammarFragmentToString(EObject ele, String prefix) {
-		return serializer.grammarFragmentToString(ele, prefix)
+	def String grammarFragmentToString(EObject object, String prefix) {
+		val node = NodeModelUtils.findActualNodeFor(object)
+		if (node === null) {
+			if (object instanceof RuleCall)
+				if (object?.rule?.name !== null)
+					return process(object.rule.name, prefix)
+			return ""
+		} else {
+			return node.text.process(prefix)
+		}
+	}
+
+	private def process(String input, String prefix) {
+		// remove leading and trailing blank lines
+		var lines = input.split('\\s*(\\r?\\n)')
+		var first = 0
+		while (isBlank(lines.get(first)))
+			first++
+		var last = lines.length - 1
+		while (isBlank(lines.get(last)))
+			last--
+		lines = lines.subList(first, last + 1)
+		// just one line, trim it and be done
+		if (lines.size() == 1)
+			return prefix + lines.get(0).trim();
+		// remove common whitespace (e.g. leading blanks) and add prefix to each line
+		val commonWhitespace = commonLeadingWhitespace(lines)
+		for (var n=0; n<lines.length; n++)
+			lines.set(n, prefix + lines.get(n).replaceAll("\t", "    ").substring(commonWhitespace.length))
+		// generate result by joining lines
+		return lines.join(codeConfig.lineDelimiter)
+	}
+
+	private def isBlank(String line) {
+		return line.trim.empty
+	}
+	
+	private def String commonLeadingWhitespace(List<String> lines) {
+		if(lines.size() < 2) return "";
+		// determine common prefix while ignoring blank lines
+		var current = Strings.repeat(" ", lines.get(0).replaceAll("\t", "    ").length())
+		for (var i = 0; i < lines.length; i++) {
+			val next = lines.get(i).replaceAll("\t", "    ");
+			if (!isBlank(next))
+				current = Strings.commonPrefix(current, next);
+		}
+		// replace blank lines with prefix to handle empty lines gracefully
+		for (var i = 0; i < lines.length; i++) {
+			val next = lines.get(i).replaceAll("\t", "    ");
+			if (isBlank(next))
+				lines.set(i, current)
+		}
+		return current;
 	}
 
 	/**
@@ -544,18 +593,6 @@ class GrammarAccessExtensions {
 			default:
 				"null"
 		}
-	}
-
-	private def ISerializer getSerializer() {
-		val delimiter = codeConfig.lineDelimiter
-		var result = xtextSerializerByLineDelimiter.get(delimiter)
-		if (result !== null) {
-			return result
-		}
-		val injector = Guice.createInjector(new LineSeparatorModule[delimiter])
-		result = injector.getInstance(ISerializer)
-		xtextSerializerByLineDelimiter.put(delimiter, result)
-		return result;
 	}
 
 	@FinalFieldsConstructor
--- a/org.eclipse.xtext.xtext.generator/xtend-gen/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.java
+++ b/org.eclipse.xtext.xtext.generator/xtend-gen/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.java
@@ -9,12 +9,10 @@
 package org.eclipse.xtext.xtext.generator.grammarAccess;
 
 import com.google.common.base.Objects;
+import com.google.common.base.Strings;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Maps;
 import com.google.inject.Binder;
-import com.google.inject.Guice;
 import com.google.inject.Inject;
-import com.google.inject.Injector;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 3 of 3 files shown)


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -396,8 +391,61 @@ class GrammarAccessExtensions {
...
+				if (object?.rule?.name !== null)
+					return process(object.rule.name, prefix)
+			return ""
+		} else {
+			return node.text.process(prefix)
+		}
+	}
+
+	private def process(String input, String prefix) {
+		// remove leading and trailing blank lines
```
cdietrich, line 407: wonder if we have a util for this somewhere

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -396,8 +391,61 @@ class GrammarAccessExtensions {
...
+		lines = lines.subList(first, last + 1)
+		// remove common whitespace (e.g. leading blanks) and add prefix to each line
+		val commonWhitespace = commonLeadingWhitespace(lines)
+		for (var n=0; n<lines.length; n++)
+			lines.set(n, prefix + lines.get(n).substring(commonWhitespace.length))
+		// generate result by joining lines
+		return lines.join(System.lineSeparator)
+	}
+
+	private def isBlank(String line) {
```
cdietrich, line 424: am sure there is a util in guava or xtext this somewhere

ArneDeutsch: The only existing method that does more or less the same is org.apache.commons.lang.StringUtils#isBlank. But it is in a jar not currently in the classpath of the project. I would avoid adding a dependency just because of this tiny method.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
ArneDeutsch: I can not find one.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
cdietrich: @szarnekow do you have an idea?

cdietrich: https://github.com/eclipse/xtext-lib/issues/247

szarnekow: `CharMatcher.whitespace().matchesAllOf(..)`

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -396,8 +391,61 @@ class GrammarAccessExtensions {
...
+		return line.trim.empty
+	}
+	
+	private def String commonLeadingWhitespace(List<String> lines) {
```
szarnekow, line 428: `com.google.common.base.Strings.commonPrefix(CharSequence, CharSequence)` may be handy here, too along with `isBlank`

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
ArneDeutsch: I see but am not convinced that it would be an improvement.
"line.trim.empty" vs. "CharMatcher.whitespace().matchesAllOf(line)" ... for me it is not more clear and longer AND depend on third party code ... I would not change it ...

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
ArneDeutsch: Have thought about this a bit. As the existing method computes the prefix for only two strings I would have to iterate all the lines, compute the common prefix with the first line, then take the shortest prefix, remove all trailing non whitespaces, check that is blank and then return it. Do not believe it will be a noticeable improvement.

szarnekow: I didn't test this, but my gut feeling is, that a normal loop along with the utility method is also very readable:
```
String commonLeadingWS(List<String> lines) {
  if (lines.size() < 2) return "";
  // start at second line since the first is left-trimmed in any case
  String current = Strings.repeat(" ", tabsToSpaces(lines.get(1)).length())
  for(int i = 1; i < lines.length; i++) {
    String next = lines.get(i);
    if (!isBlank(next))) {
      current = Strings.commonPrefix(current, tabsToSpaces(lines.get(i));
    }
  }
  return current;
}
```

ArneDeutsch: Get your idea and guess it will work (we could NOT skip the first element but that is minor).

Modified it a bit to the following:

```
	private def String commonLeadingWhitespace(List<String> lines) {
		if(lines.size() < 2) return "";
		var current = Strings.repeat(" ", lines.get(0).replaceAll("\t", "    ").length())
		for (var i = 0; i < lines.length; i++) {
			val next = lines.get(i).replaceAll("\t", "    ");
			if (!isBlank(next))
				current = Strings.commonPrefix(current, next);
		}
		return current;
	}
```

Do you think a replacement of a tab with 4 spaces will work well? Or do we have to read some preferences or something like that? If the user uses tabs and spaces inconsistenly and has one tab for e.g. 2 spaces, the resulting formatting might be wrong ...

szarnekow: It was only an idea. We could use common prefix without that replace operation.

szarnekow: Why I excluded the first line from the common prefix:
```
SomeGrammarRule: 
  {NewThingy} ('{'
    elements+=Element;
  '}');
```

When you look at the group in parentheses, you'd want to strip the leading space from the last two lines but the element doesn't have a leading space in the first line. I didn't check the current approach carefully though it might indeed be flawed to exclude line zero.

ArneDeutsch: Ok, you are right. In case of inconsistent usage of blanks and tabs we will get wired results in any case. With or without the replace operation. In any case your solution is much more readable.

ArneDeutsch: Because of the skipping of the first line ... maybe I get you wrong, but in your example we do not want to remove any leading whitespace ... because of the first line. In this sense it is the counter example ... we need to consider the first line.

szarnekow: The grammar rule has a body of type `Group` - the group has two elements: one action and another group .
The nested group shall be put into the grammar access as a comment. It starts at the opening parenthesis and the keyword: `(`{``. So the raw text of that group is 
```
('{'
    elements+=Element;
  '}');
```
Note the indentation of the last two lines which is off by two chars. If the first line considered in the leading WS computation, there won't be any whitespace to be trimmed. That was my line of thought.

ArneDeutsch: I see. In that case the algorithm will not work. On the other hand the complete rule will be misformatted in case we skip the first line. Guess it is not possible to make every case 100% "correct" ... hence we have to try to make the most common cases good looking. And the complete rule is probably the most important case.

cdietrich:
might need rebasing for https://github.com/eclipse/xtext-core/pull/1667

ArneDeutsch:
Rebased.

cdietrich:
@szarnekow can you please reexamine

org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/generator/GrammarAccessExtensions2Test.java
```diff
@@ -70,19 +70,20 @@ public void testGrammarFragmentToString() throws Exception {
 				"	| '?:';\n";
 		String expected = 
 				"//OpOther:" + NL +
-				"//	'->'" + NL +
```
szarnekow, line 73: Nice nice!

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -66,9 +65,6 @@ class GrammarAccessExtensions {
 		"\t" -> "tab",
 		"\\" -> "backslash"
 	};
-	val Map<String, ISerializer> xtextSerializerByLineDelimiter = Maps.newHashMapWithExpectedSize(2)
```
szarnekow, line 69: 💪

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -66,9 +65,6 @@ class GrammarAccessExtensions {
...
-	
-	@Inject CodeConfig codeConfig
```
szarnekow, line 71: I thought we'd need that for the line ending?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -396,8 +392,49 @@ class GrammarAccessExtensions {
...
+		var last = lines.length - 1
+		while (isBlank(lines.get(last)))
+			last--
+		lines = lines.subList(first, last + 1)
+		// remove common whitespace (e.g. leading blanks) and add prefix to each line
+		val commonWhitespace = commonLeadingWhitespace(lines)
+		for (var n=0; n<lines.length; n++)
+			lines.set(n, prefix + lines.get(n).substring(commonWhitespace.length))
+		// generate result by joining lines
+		return lines.join(System.lineSeparator)
```
szarnekow, line 422: This should use the line ending from the code config

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -396,8 +392,49 @@ class GrammarAccessExtensions {
...
+		var lines = input.split('(\\r?\\n)')
+		var first = 0
+		while (isBlank(lines.get(first)))
+			first++
+		var last = lines.length - 1
+		while (isBlank(lines.get(last)))
+			last--
+		lines = lines.subList(first, last + 1)
+		// remove common whitespace (e.g. leading blanks) and add prefix to each line
+		val commonWhitespace = commonLeadingWhitespace(lines)
```
szarnekow, line 418: commonLeadingWhitespace ignores blank lines. Let's assume we do have three lines:
- `   text`
- ` `
- `   more text`

The commonWhitespace should be 3.
The second line has only a single char, though. I'd assume that the logic in line 420 will now fail: `lines.get(1).substring(3)` will throw an IOOBE.

I'd say blank lines should be replaced by the prefix itself.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -396,8 +392,49 @@ class GrammarAccessExtensions {
...
+					return process(object.rule.name, prefix)
+			return ""
+		} else {
+			return node.text.process(prefix)
+		}
+	}
+
+	private def process(String input, String prefix) {
+		// remove leading and trailing blank lines
+		var lines = input.split('(\\r?\\n)')
```
szarnekow, line 409: If we'd use `\\s*(\\r?\\n)` we'd automatically righttrim the lines and avoid dangling whitespace in the generated grammar access. What do you think?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -396,8 +392,49 @@ class GrammarAccessExtensions {
...
+	private def isBlank(String line) {
+		return line.trim.empty
+	}
+	
+	private def String commonLeadingWhitespace(List<String> lines) {
+		if(lines.size() < 2) return "";
+		var current = Strings.repeat(" ", lines.get(0).length())
+		for (var i = 0; i < lines.length; i++) {
+			val next = lines.get(i);
+			if (!isBlank(next))
```
szarnekow, line 434: Not sure if we should treat indentation by tabs specially here.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
ArneDeutsch: Yes, good idea.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
ArneDeutsch: Good catch. Will fix.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
ArneDeutsch: ok

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
ArneDeutsch: Ok.

cdietrich:
@szarnekow @ArneDeutsch what is the state of this PR?

ArneDeutsch:
From my point of view it is ready for merge. Would be good to either get a go or suggestions what to change.

szarnekow (approved):
Please go ahead and regenerate the test langs such that we can assess the fallout

miklossy:
> 
> 
> Please go ahead and regenerate the test langs such that we can assess the fallout

@ArneDeutsch Could you follow the [RegenerateLanguages.md](https://github.com/eclipse/xtext/blob/master/dev-doc/RegenerateLanguages.md) and give feedback, whether it is understandable?

ArneDeutsch:
Will do.

ArneDeutsch:
Because I have generated from windows I get windows line endings in some files, rendering them useless for review. But many are actually good. We have to redo this step on a linux machine, but probably the genereate is good enough to find remaining issues in comment generator code. Please use the files where line endings have NOT changed for review.

miklossy:
@ArneDeutsch you can give it a try using WSL on Windows. I also used that to regenerate the languages on Windows to avoid line endings changes.

ArneDeutsch:
Yes, I could redo it ... but I guess the tasks executed from eclipse will still be problematic.

ArneDeutsch:
Ok, after review myself, too, I would change two things:
1. Replace each tab by four spaces when doing the replace .... this will at least handle some cases better.
2. Just trim the stuff it it is exacly one line.
@szarnekow Do you think that will be good?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
```diff
@@ -396,8 +394,59 @@ class GrammarAccessExtensions {
...
+		// remove leading and trailing blank lines
+		var lines = input.split('\\s*(\\r?\\n)')
+		var first = 0
+		while (isBlank(lines.get(first)))
+			first++
+		var last = lines.length - 1
+		while (isBlank(lines.get(last)))
+			last--
+		lines = lines.subList(first, last + 1)
+		// just one line, trim it and be done
```
ArneDeutsch, line 419: Added this case for single lines.

szarnekow:
> Ok, after review myself, too, I would change two things:
> 
> 1. Replace each tab by four spaces when doing the replace .... this will at least handle some cases better ... e.g. the case where tabs are used for indentation.
> 2. Just trim the stuff it it is exacly one line. Obviously we do not handle that case at all right now.
>    @szarnekow Do you think that will be good?

Yes, I think both proposals make sense.

szarnekow:
> Because I have generated from windows I get windows line endings in some files, rendering them useless for review. But many are actually good. We have to redo this step on a linux machine, but probably the genereate is good enough to find remaining issues in comment generator code. Please use the files where line endings have NOT changed for review.

Is it better with an explicit NL configuration in the workflow?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/grammarAccess/GrammarAccessExtensions.xtend
ArneDeutsch: Now we do. Replace each tab with four space to handle the common case where we indent with tabs.

ArneDeutsch:
I fixed an IndexOutOfBoundsException in my recent code and regenerated some languages locally again. Looks quite good now, original format is mostly intact. Think we are really good to go now.

PR #1672, opened 2021-02-05
[#897] Folding Support for LSP

### LINKED ISSUE #897: [LSP] Implement Folding
- Implement Service
  - Would be nice if the actual impl would be free of LSP4J classes and similar to the eclipse api
  - so that it can be adapted in both eclipse and lsp4j
- Test Service (implement folding in lsp test lang & write a test for it, maybe adapt AbstractLanguageServerTest)

Darkgaja:
This solution should be easily adaptable for the eclipse/xtext-eclipse repo. The API is mostly based on the original implementation for Eclipse.

This should close #897

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/folding/AbstractFoldingRangeProviderTest.java
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/folding/AbstractFoldingRangeProviderTest.java
@@ -1,0 +1,55 @@
+/*******************************************************************************
+ * Copyright (c) 2021 TypeFox GmbH (http://www.typefox.io) and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.editor.folding;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.xtext.ide.editor.folding.FoldingRange;
+import org.eclipse.xtext.ide.editor.folding.IFoldingRangeProvider;
+import org.eclipse.xtext.resource.XtextResource;
+import org.eclipse.xtext.testing.util.ParseHelper;
+import org.eclipse.xtext.util.CancelIndicator;
+import org.junit.Assert;
+
+import com.google.inject.Inject;
+
+/**
+ * @author Mark Sujew - Initial contribution and API
+ */
+public abstract class AbstractFoldingRangeProviderTest<T extends EObject> {
+
+	@Inject
+	private IFoldingRangeProvider foldingRangeProvider;
+	@Inject
+	private ParseHelper<T> parseHelper;
+
+	protected void assertFoldingRanges(CharSequence content, FoldingRange... expectedRanges) {
+		XtextResource resource;
+		try {
+			resource = (XtextResource) parseHelper.parse(content).eResource();
+		} catch (Exception e) {
+			throw new RuntimeException(e);
+		}
+
+		List<FoldingRange> foldingRanges = foldingRangeProvider.getFoldingRanges(resource, CancelIndicator.NullImpl)
+				.stream().collect(Collectors.toList());
+
+		Assert.assertEquals(expectedRanges.length, foldingRanges.size());
+		
+		for (int i = 0; i < foldingRanges.size(); i++) {
+			FoldingRange range = foldingRanges.get(i);
+			FoldingRange expected = expectedRanges[i];
+			Assert.assertEquals(expected.getOffset(), range.getOffset());
+			Assert.assertEquals(expected.getLength(), range.getLength());
+			Assert.assertEquals(expected.getKind(), range.getKind());
+		}
+	}
+}
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/folding/FoldingRangeProviderTest.java
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/folding/FoldingRangeProviderTest.java
@@ -1,0 +1,51 @@
+/*******************************************************************************
+ * Copyright (c) 2021 TypeFox GmbH (http://www.typefox.io) and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.editor.folding;
+
+import org.eclipse.xtext.ide.editor.folding.FoldingRange;
+import org.eclipse.xtext.ide.editor.folding.FoldingRangeKind;
+import org.eclipse.xtext.ide.tests.testlanguage.TestLanguageIdeInjectorProvider;
+import org.eclipse.xtext.ide.tests.testlanguage.testLanguage.Model;
+import org.eclipse.xtext.testing.InjectWith;
+import org.eclipse.xtext.testing.XtextRunner;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * @author Mark Sujew - Initial contribution and API
+ */
+@RunWith(XtextRunner.class)
+@InjectWith(TestLanguageIdeInjectorProvider.class)
+public class FoldingRangeProviderTest extends AbstractFoldingRangeProviderTest<Model> {
+
+	@Test
+	public void foldingRangesProvider() {
+		String actual = "package testLanguage {"
+				+ "\n"
+				+ "\n";
+		int commentIndex = actual.length();
+		
+		actual += "/*\n"
+				+ " * This s a multiline comment"
+				+ " */";
+		int commentLength = actual.length() - commentIndex;
+		
+		actual += "\n";
+		int typeIndex = actual.length();
+		
+		actual += "type myType {"
+				+ "\n" 
+				+ "}";
+		int typeLength = actual.length() - typeIndex;
+		actual += "\n}";
+		
+		assertFoldingRanges(actual, new FoldingRange(0, actual.length()), new FoldingRange(commentIndex, commentLength, FoldingRangeKind.COMMENT), new FoldingRange(typeIndex, typeLength));
+	}
+	
+}
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeAcceptor.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeAcceptor.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2010, 2021 Michael Clay and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+import java.util.Collection;
+
+import org.eclipse.xtext.util.ITextRegion;
+
+/**
+ * @author Michael Clay - Initial contribution and API
+ * @author Sebastian Zarnekow - Introduced FoldedPosition
+ * @author Mark Sujew - Ported to IDE project
+ * 
+ * @since 2.26
+ */
+public class DefaultFoldingRangeAcceptor implements IFoldingRangeAcceptor {
+	private Collection<FoldingRange> result;
+
+	public DefaultFoldingRangeAcceptor(Collection<FoldingRange> result) {
+		this.result = result;
+	}
+
+	public Collection<FoldingRange> getFoldingRanges() {
+		return result;
+	}
+
+	@Override
+	public void accept(int offset, int length, FoldingRangeKind kind, boolean initiallyFolded,
+			ITextRegion visualPlaceholderRegion) {
+		result.add(createFoldingRange(offset, length, kind, initiallyFolded, visualPlaceholderRegion));
+	}
+
+	protected FoldingRange createFoldingRange(int offset, int length, FoldingRangeKind kind, boolean initiallyFolded,
+			ITextRegion visualPlaceholderRegion) {
+		return new FoldingRange(offset, length, kind, initiallyFolded, visualPlaceholderRegion);
+	}
+}
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
@@ -1,0 +1,202 @@
+/*******************************************************************************
+ * Copyright (c) 2010, 2021 Michael Clay and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+import java.util.Collection;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.emf.common.util.TreeIterator;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.xtext.TerminalRule;
+import org.eclipse.xtext.documentation.impl.AbstractMultiLineCommentProvider;
+import org.eclipse.xtext.nodemodel.ICompositeNode;
+import org.eclipse.xtext.nodemodel.INode;
+import org.eclipse.xtext.nodemodel.util.NodeModelUtils;
+import org.eclipse.xtext.parser.IParseResult;
+import org.eclipse.xtext.resource.ILocationInFileProvider;
+import org.eclipse.xtext.resource.XtextResource;
+import org.eclipse.xtext.util.CancelIndicator;
+import org.eclipse.xtext.util.ITextRegion;
+import org.eclipse.xtext.util.ITextRegionWithLineInformation;
+import org.eclipse.xtext.util.LineAndColumn;
+import org.eclipse.xtext.util.TextRegion;
+
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
+
+/**
+ * @author Michael Clay - Initial contribution and API
+ * @author Sebastian Zarnekow - Introduced FoldedRegion, use ILocationInFileProvider
+ * @author Mark Sujew - Ported to IDE project
+ * 
+ * @since 2.26
+ */
+public class DefaultFoldingRangeProvider implements IFoldingRangeProvider {
+
+	@Inject
+	private ILocationInFileProvider locationInFileProvider;
+
+	@Inject(optional = true)
+	@Named(AbstractMultiLineCommentProvider.RULE)
+	private String ruleName = "ML_COMMENT";
+
+	protected static final Pattern TEXT_PATTERN_IN_COMMENT = Pattern.compile("\\w");
+
+	@Override
+	public SortedSet<FoldingRange> getFoldingRanges(XtextResource xtextDocument, CancelIndicator cancelIndicator) {
+		SortedSet<FoldingRange> result = new TreeSet<>((a, b) -> Integer.compare(a.getOffset(), b.getOffset()));
+		IFoldingRangeAcceptor foldingRegionAcceptor = createAcceptor(xtextDocument, result);
+		computeObjectFolding(xtextDocument, foldingRegionAcceptor, cancelIndicator);
+		computeCommentFolding(xtextDocument, foldingRegionAcceptor);
+		return result;
+	}
+
+	protected void computeObjectFolding(XtextResource resource, IFoldingRangeAcceptor foldingRangeAcceptor,
+			CancelIndicator cancelIndicator) {
+		IParseResult parseResult = resource.getParseResult();
+		if (parseResult != null) {
+			EObject rootASTElement = parseResult.getRootASTElement();
+			if (rootASTElement != null) {
+				if (cancelIndicator.isCanceled())
+					throw new OperationCanceledException();
+				if (isHandled(rootASTElement)) {
+					acceptObjectFolding(rootASTElement, foldingRangeAcceptor);
+				}
+				if (shouldProcessContent(rootASTElement)) {
+					TreeIterator<EObject> allContents = rootASTElement.eAllContents();
+					while (allContents.hasNext()) {
+						if (cancelIndicator.isCanceled())
+							throw new OperationCanceledException();
+						EObject eObject = allContents.next();
+						if (isHandled(eObject)) {
+							acceptObjectFolding(eObject, foldingRangeAcceptor);
+						}
+						if (!shouldProcessContent(eObject)) {
+							allContents.prune();
+						}
+					}
+				}
+			}
+		}
+	}
+
+	protected void acceptObjectFolding(EObject eObject, IFoldingRangeAcceptor foldingRangeAcceptor) {
+		ITextRegion region = locationInFileProvider.getFullTextRegion(eObject);
+		if (region != null) {
+			INode eObjectNode = NodeModelUtils.getNode(eObject);
+			ITextRegion significant = buildSignificantRegion(locationInFileProvider.getSignificantTextRegion(eObject),
+					eObjectNode);
+			foldingRangeAcceptor.accept(region.getOffset(), region.getLength(), null, false, significant);
+		}
+	}
+
+	protected ITextRegion buildSignificantRegion(ITextRegion significantRegion, INode node) {
+		if (significantRegion == null || node == null) {
+			return null;
+		}
+		int offset = significantRegion.getOffset();
+		int endOffset = significantRegion.getOffset() + significantRegion.getLength();
+
+		int startLine;
+		int endLine;
+		if (significantRegion instanceof ITextRegionWithLineInformation) {
+			ITextRegionWithLineInformation lineInfoRegion = (ITextRegionWithLineInformation) significantRegion;
+			startLine = lineInfoRegion.getLineNumber();
+			endLine = lineInfoRegion.getEndLineNumber();
+		} else {
+			startLine = NodeModelUtils.getLineAndColumn(node, offset).getLine();
+			endLine = NodeModelUtils.getLineAndColumn(node, endOffset).getLine();
+		}
+
+		if (startLine != endLine) {
+			/*
+			 * The Eclipse IDE can only use the significant region if it starts and ends on the same line.
+			 * Therefore, we here calculate the index of the end of the line.
+			 */
+			for (int index = offset; index < endOffset; index++) {
+				LineAndColumn lineInfo = NodeModelUtils.getLineAndColumn(node, index);
+				if (lineInfo.getLine() != startLine) {
+					return new TextRegion(offset, index - offset - 1);
+				}
+			}
+		}
+		return significantRegion;
+	}
+
+	protected void computeCommentFolding(XtextResource resource, IFoldingRangeAcceptor foldingRangeAcceptor) {
+		IParseResult parseResult = resource.getParseResult();
+		if (parseResult != null) {
+			EObject rootASTElement = parseResult.getRootASTElement();
+			acceptCommentNodes(rootASTElement, foldingRangeAcceptor);
+		}
+	}
+
+	protected void acceptCommentNodes(EObject eObject, IFoldingRangeAcceptor foldingRangeAcceptor) {
+		ICompositeNode node = NodeModelUtils.getNode(eObject);
+		if (node != null) {
+			for (INode leafNode : node.getAsTreeIterable()) {
+				if (leafNode.getGrammarElement() instanceof TerminalRule
+						&& ruleName.equalsIgnoreCase(((TerminalRule) leafNode.getGrammarElement()).getName())) {
+					acceptCommentFolding(leafNode, foldingRangeAcceptor);
+				}
+			}
+		}
+	}
+
+	protected void acceptCommentFolding(INode commentNode, IFoldingRangeAcceptor foldingRangeAcceptor) {
+		int offset = commentNode.getOffset();
+		int length = commentNode.getLength();
+		Matcher matcher = getTextPatternInComment().matcher(commentNode.getText());
+		if (matcher.find()) {
+			TextRegion significant = new TextRegion(offset + matcher.start(), 0);
+			foldingRangeAcceptor.accept(offset, length, FoldingRangeKind.COMMENT, significant);
+		} else {
+			foldingRangeAcceptor.accept(offset, length, FoldingRangeKind.COMMENT);
+		}
+	}
+
+	protected IFoldingRangeAcceptor createAcceptor(XtextResource resource, Collection<FoldingRange> foldingRanges) {
+		return new DefaultFoldingRangeAcceptor(foldingRanges);
+	}
+
+	protected ILocationInFileProvider getLocationInFileProvider() {
+		return locationInFileProvider;
+	}
+
+	protected String getMultilineCommentRuleName() {
+		return ruleName;
+	}
+
+	/**
+	 * @return the regular expression that finds the first significant part of a multi line comment.
+	 */
+	protected Pattern getTextPatternInComment() {
+		return TEXT_PATTERN_IN_COMMENT;
+	}
+
+	/**
+	 * @return <code>true</code> if the object should be folded if it spans more than one line. Default is
+	 *         <code>false</code> if and only if the object is the root object of the resource.
+	 */
+	protected boolean isHandled(EObject eObject) {
+		return eObject.eContainer() != null;
+	}
+
+	/**
+	 * @return clients should return <code>false</code> to abort the traversal of the model.
+	 */
+	protected boolean shouldProcessContent(EObject object) {
+		return true;
+	}
+}
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
@@ -1,0 +1,93 @@
+/*******************************************************************************
+ * Copyright (c) 2021 TypeFox GmbH (http://www.typefox.io) and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+import org.eclipse.xtext.util.ITextRegion;
+
+import com.google.common.base.Objects;
+
+/**
+ * Represents an abstraction for different folding implementations.
+ * Specifically Eclipse's <code>FoldingRegion</code> and the <code>FoldingRange</code> of the LSP specification.
+ * 
+ * @author Mark Sujew - Initial contribution and API
+ * 
+ * @since 2.26
+ */
+public class FoldingRange {
+
+	private final int offset;
+	private final int length;
+	private final FoldingRangeKind kind;
+	private final boolean initiallyFolded;
+	private final ITextRegion visualPlaceholderRegion;
+
+	public FoldingRange(int offset, int length) {
+		this(offset, length, null);
+	}
+
+	public FoldingRange(int offset, int length, FoldingRangeKind kind) {
+		this(offset, length, kind, false, null);
+	}
+
+	public FoldingRange(int offset, int length, FoldingRangeKind kind, boolean initiallyFolded,
+			ITextRegion visualPlaceholderRegion) {
+		this.offset = offset;
+		this.length = length;
+		this.kind = kind;
+		this.initiallyFolded = initiallyFolded;
+		this.visualPlaceholderRegion = visualPlaceholderRegion;
+	}
+	
+	public boolean isInitiallyFolded() {
+		return initiallyFolded;
+	}
+
+	public ITextRegion getVisualPlaceholderRegion() {
+		return visualPlaceholderRegion;
+	}
+
+	public FoldingRangeKind getKind() {
+		return kind;
+	}
+
+	public int getOffset() {
+		return offset;
+	}
+
+	public int getLength() {
+		return length;
+	}
+
+	@Override
+	public boolean equals(Object other) {
+		if (other instanceof FoldingRange) {
+			FoldingRange range = (FoldingRange) other;
+			return offset == range.offset && length == range.length && initiallyFolded == range.initiallyFolded
+					&& Objects.equal(kind, range.kind)
+					&& Objects.equal(visualPlaceholderRegion, range.visualPlaceholderRegion);
+		}
+		return false;
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hashCode(offset, length, initiallyFolded, kind, visualPlaceholderRegion);
+	}
+	
+	@Override
+	public String toString() {
+		StringBuilder content = new StringBuilder();
+		content.append("offset=").append(offset);
+		content.append(", length=").append(length);
+		content.append(", kind=").append(kind);
+		content.append(", initiallyFolded=").append(initiallyFolded);
+		return content.toString();
+	}
+}
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRangeKind.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRangeKind.java
@@ -1,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2021 TypeFox GmbH (http://www.typefox.io) and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+/**
+ * Contains the possible kinds of folding ranges defined by the LSP spec.
+ * 
+ * @author Mark Sujew - Initial contribution and API
+ * 
+ * @since 2.26
+ */
+public class FoldingRangeKind {
+	public static final FoldingRangeKind COMMENT = new FoldingRangeKind("comment");
+	public static final FoldingRangeKind IMPORTS = new FoldingRangeKind("imports");
+	public static final FoldingRangeKind REGION = new FoldingRangeKind("region");
+
+	private String kind;
+
+	private FoldingRangeKind(String kind) {
+		this.kind = kind;
+	}
+
+	@Override
+	public String toString() {
+		return kind;
+	}
+
+	@Override
+	public boolean equals(Object other) {
+		if (other instanceof FoldingRangeKind) {
+			return this.kind.equals(((FoldingRangeKind) other).kind);
+		}
+		return false;
+	}
+
+	@Override
+	public int hashCode() {
+		return kind.hashCode();
+	}
+}
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
@@ -1,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2010, 2021 Michael Clay and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+import org.eclipse.xtext.util.ITextRegion;
+
+/**
+ * @author Michael Clay - Initial contribution and API
+ * @author Sebastian Zarnekow - Distinguish between total and identifying region
+ * @author Mark Sujew - Ported to IDE project
+ * 
+ * @since 2.26
+ */
+public interface IFoldingRangeAcceptor {
+
+	default void accept(int offset, int length) {
+		accept(offset, length, false);
+	}
+
+	default void accept(int offset, int length, FoldingRangeKind kind) {
+		accept(offset, length, kind, null);
+	}
+
+	default void accept(int offset, int length, boolean initiallyFolded) {
+		accept(offset, length, null, initiallyFolded, null);
+	}
+
+	default void accept(int offset, int length, ITextRegion visualPlaceholderRegion) {
+		accept(offset, length, null, false, visualPlaceholderRegion);
+	}
+
+	default void accept(int offset, int length, FoldingRangeKind kind, ITextRegion visualPlaceholderRegion) {
+		accept(offset, length, kind, false, visualPlaceholderRegion);
+	}
+
+	void accept(int offset, int length, FoldingRangeKind kind, boolean initiallyFolded,
+			ITextRegion visualPlaceholderRegion);
+}
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeProvider.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeProvider.java
@@ -1,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) 2010, 2021 Michael Clay and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+import java.util.SortedSet;
+
+import org.eclipse.xtext.resource.XtextResource;
+import org.eclipse.xtext.util.CancelIndicator;
+
+import com.google.inject.ImplementedBy;
+
+/**
+ * Compute the folding regions in the given document. This is a blocking action when opening the editor in e4, so
+ * clients should be careful to not resolve too many cross references in their implementation.
+ * 
+ * @author Michael Clay - Initial contribution and API
+ * @author Sebastian Zarnekow - Refactoring, introduced FoldedPosition
+ * @author Mark Sujew - Ported to IDE project
+ * 
+ * @since 2.26
+ */
+@ImplementedBy(DefaultFoldingRangeProvider.class)
+public interface IFoldingRangeProvider {
+
+	/**
+	 * @return the set of <code>FoldingRanges</code> for the given document
+	 */
+	SortedSet<FoldingRange> getFoldingRanges(XtextResource xtextDocument, CancelIndicator cancelIndicator);
+}
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
@@ -56,6 +56,9 @@ import org.eclipse.lsp4j.ExecuteCommandCapabilities;
 import org.eclipse.lsp4j.ExecuteCommandOptions;
 import org.eclipse.lsp4j.ExecuteCommandParams;
 import org.eclipse.lsp4j.FileChangeType;
+import org.eclipse.lsp4j.FoldingRange;
+import org.eclipse.lsp4j.FoldingRangeCapabilities;
+import org.eclipse.lsp4j.FoldingRangeRequestParams;
 import org.eclipse.lsp4j.Hover;
 import org.eclipse.lsp4j.HoverParams;
 import org.eclipse.lsp4j.InitializeParams;
@@ -111,6 +114,7 @@ import org.eclipse.xtext.ide.server.commands.ExecutableCommandRegistry;
 import org.eclipse.xtext.ide.server.concurrent.RequestManager;
 import org.eclipse.xtext.ide.server.contentassist.ContentAssistService;
 import org.eclipse.xtext.ide.server.findReferences.WorkspaceResourceAccess;
+import org.eclipse.xtext.ide.server.folding.FoldingRangeService;
 import org.eclipse.xtext.ide.server.formatting.FormattingService;
 import org.eclipse.xtext.ide.server.hover.IHoverService;
 import org.eclipse.xtext.ide.server.occurrences.IDocumentHighlightService;
@@ -135,6 +139,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.LinkedListMultimap;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
 import com.google.inject.Inject;
 
@@ -268,8 +273,14 @@ public class LanguageServerImpl implements LanguageServer, WorkspaceService, Tex
 			textDocument = clientCapabilities.getTextDocument();
 		}
 		RenameCapabilities rename = null;
+		FoldingRangeCapabilities folding = null;
 		if (textDocument != null) {
 			rename = textDocument.getRename();
+			folding = textDocument.getFoldingRange();
+		}
+		if (folding != null) {
+			serverCapabilities.setFoldingRangeProvider(allLanguages.stream()
+					.anyMatch(serviceProvider -> serviceProvider.get(FoldingRangeService.class) != null));
 		}
 		Boolean prepareSupport = null;
 		if (rename != null) {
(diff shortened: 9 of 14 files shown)


org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -91,6 +91,7 @@ import org.junit.jupiter.api.BeforeEach
 
 import static extension org.eclipse.lsp4j.util.Ranges.containsRange
 import static extension org.eclipse.xtext.util.Strings.*
+import org.eclipse.lsp4j.FoldingRangeRequestParams
```
miklossy, line 94: Could you please organize the imports?

cdietrich:
@miklossy could you give this a try with your sample dsl

miklossy:
> @miklossy could you give this a try with your sample dsl

Yes, I will try this out and give feedback.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeAcceptor.java
```diff
@@ -0,0 +1,59 @@
...
+package org.eclipse.xtext.ide.editor.folding;
+
+import java.util.Collection;
+
+/**
+ * @author Michael Clay - Initial contribution and API
+ * @author Sebastian Zarnekow - Introduced FoldedPosition
+ * @author Mark Sujew - Ported to IDE project
+ */
+public class DefaultFoldingRangeAcceptor implements IFoldingRangeAcceptor<Void> {
```
cdietrich, line 19: please add  `@since 2.25` to new API classes

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
```diff
@@ -0,0 +1,139 @@
...
+import com.google.common.collect.Sets;
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
+
+/**
+ * @author Michael Clay - Initial contribution and API
+ * @author Sebastian Zarnekow - Introduced FoldedRegion, use ILocationInFileProvider
+ * @author Mark Sujew - Ported to IDE project
+ */
+public class DefaultFoldingRangeProvider implements IFoldingRangeProvider {
```
cdietrich, line 37: please add  `@since 2.25` to new API classes

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
```diff
@@ -0,0 +1,53 @@
...
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+/**
+ * @author Mark Sujew - Initial contribution and API
+ */
+public class FoldingRange {
```
cdietrich, line 14: please add  `@since 2.25` to new API classes

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRangeKind.java
```diff
@@ -0,0 +1,18 @@
...
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+/**
+ * @author Mark Sujew - Initial contribution and API
+ */
+public class FoldingRangeKind {
```
cdietrich, line 14: please add  `@since 2.25` to new API classes

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+/**
+ * @author Michael Clay - Initial contribution and API
+ * @author Sebastian Zarnekow - Distinguish between total and identifying region
+ * @author Mark Sujew - Ported to IDE project
+ */
+public interface IFoldingRangeAcceptor<T> {
```
cdietrich, line 17: please add  `@since 2.25` to new API classes

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeProvider.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Compute the folding regions in the given document. This is a blocking action when opening the editor in e4, so
+ * clients should be careful to not resolve too many cross references in their implementation.
+ * 
+ * @author Michael Clay - Initial contribution and API
+ * @author Sebastian Zarnekow - Refactoring, introduced FoldedPosition
+ * @author Mark Sujew - Ported to IDE project
+ */
+@ImplementedBy(DefaultFoldingRangeProvider.class)
+public interface IFoldingRangeProvider {
```
cdietrich, line 28: please add  `@since 2.25` to new API classes

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
```diff
@@ -0,0 +1,52 @@
...
+import org.eclipse.xtext.util.CancelIndicator;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+
+/**
+ * @author Mark Sujew - Initial contribution and API
+ */
+@Singleton
+public class FoldingRangeService {
```
cdietrich, line 28: please add  `@since 2.25` to new API classes

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -726,6 +727,20 @@ abstract class AbstractLanguageServerTest implements Endpoint {
 		val result = new Document(1, fileInfo.contents).applyChanges(<TextEdit>newArrayList(changes.get()).reverse)
 		assertEqualsStricter(configuration.expectedText, result.contents)
 	}
+	
+	protected def testFolding((FoldingConfiguration)=>void configurator) {
```
cdietrich, line 731: please add  `@since 2.25` to new API method

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/FoldingConfiguration.java
```diff
@@ -0,0 +1,15 @@
+package org.eclipse.xtext.testing;
+
+public class FoldingConfiguration extends TextDocumentConfiguration {
```
cdietrich, line 3: please add  `@since 2.25` to new API classes

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeAcceptor.java
```diff
@@ -0,0 +1,61 @@
+/*******************************************************************************
+ * Copyright (c) 2010 Michael Clay and others.
```
miklossy, line 2: I think we should use `2021` in the copyright. @cdietrich what do you think?

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -42,6 +42,7 @@ import org.eclipse.lsp4j.DocumentSymbol
 import org.eclipse.lsp4j.DocumentSymbolParams
 import org.eclipse.lsp4j.FileChangeType
 import org.eclipse.lsp4j.FileEvent
+import org.eclipse.lsp4j.FoldingRangeRequestParams
```
miklossy, line 45: Please also update the year in the [copyright](https://github.com/eclipse/xtext-core/blob/bc72a8f56082ee1bc6ff9fc0f14f379b6a3c4318/org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend#L2) of the AbstractLanguageServerTest class.

org.eclipse.xtext.testing/xtend-gen/org/eclipse/xtext/testing/AbstractLanguageServerTest.java
```diff
@@ -98,6 +100,20 @@
...
+import org.eclipse.xtext.testing.FoldingConfiguration;
+import org.eclipse.xtext.testing.FormattingConfiguration;
+import org.eclipse.xtext.testing.HoverTestConfiguration;
+import org.eclipse.xtext.testing.RangeFormattingConfiguration;
+import org.eclipse.xtext.testing.ReferenceTestConfiguration;
+import org.eclipse.xtext.testing.SignatureHelpConfiguration;
+import org.eclipse.xtext.testing.TestCompletionConfiguration;
+import org.eclipse.xtext.testing.TextDocumentConfiguration;
+import org.eclipse.xtext.testing.TextDocumentPositionConfiguration;
+import org.eclipse.xtext.testing.WorkspaceSymbolConfiguration;
```
miklossy, line 116: These imports should not be necessary, since the imported classes reside in the same package as the class itself.

@Darkgaja Which Xtend version do you use? These imports should not be generated if you use Xtend 2.22 or above, see https://github.com/eclipse/xtext-xtend/issues/1014

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRangeKind.java
miklossy: Would it make sense to reuse the [org.eclipse.lsp4j.FoldingRangeKind](https://github.com/eclipse/lsp4j/blob/master/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/FoldingRangeKind.java) class instead of this class?

cdietrich: @miklossy i assume the idea is to make the impl independent from LSP4J APIs. I like this approach a lot.

org.eclipse.xtext.testing/xtend-gen/org/eclipse/xtext/testing/AbstractLanguageServerTest.java
Darkgaja: That's weird. I'm using Eclipse DSL 2020-12 with Xtend 2.24. Even when I'm deleting them manually, they get regenerated everytime I edit the Xtend file. However, using Gradle with the `-PcompileXtend=true` option generates this file correctly.

cdietrich: the fix for import gen is in M1 and the build bootstraps against M2, so it has it. The 2.24 in your IDE wont have it.

cdietrich: @miklossy the version is 2.25.0.M1 not 2.22.0

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2020 TypeFox GmbH (http://www.typefox.io) and others.
+ * Copyright (c) 2016, 2020, 2021 TypeFox GmbH (http://www.typefox.io) and others.
```
miklossy, line 2: Please modify `2016, 2020, 2021` to `2016, 2021`.
From the [eclipse handbook](https://www.eclipse.org/projects/handbook/#ip-copyright-headers):

> The year is expressed as a single year or a range spanning the year of the very initial assertion and that of the most recent change.

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
miklossy: Thank for fixing the copyright header in the `AbstractLanguageServerTest.xtend` file. Please also fix it in the generated `AbstractLanguageServerTest.java` file.

Darkgaja: Oops, looks like I didn't pay attention there. Thanks for the heads-up.

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -726,6 +727,23 @@ abstract class AbstractLanguageServerTest implements Endpoint {
...
+		configuration.filePath = 'MyModel.' + fileExtension
+		configurator.apply(configuration)
+		
+		val fileInfo = initializeContext(configuration)
+		
+		val foldings = languageServer.foldingRange(new FoldingRangeRequestParams => [
+			textDocument = new TextDocumentIdentifier(fileInfo.uri)
+		]).get()
+		
+		assertEqualsStricter(configuration.expectedFoldings, String.join("", foldings.map[toString]));
```
miklossy, line 745: Instead of using `String.join("", foldings.map[toString])`, I would prefer using `foldings.toExpectation` and adding the following dispatch method:
```Xtend
protected dispatch def String toExpectation(FoldingRange foldingRange) {
	foldingRange.toString	
}
```
In that case, the code will remain consistent to the already existing actual value calculations.

Let's see what the other colleagues think about this suggestion.

miklossy:
I added the following folding test cases to the Domainmodel language available in the [Xtext - 15 Minutes Tutorial](https://www.eclipse.org/Xtext/documentation/102_domainmodelwalkthrough.html):

```Xtend
class DomainmodelFoldingRangeLanguageServerTest extends AbstractLanguageServerTest {

	new() {
		super("dmodel")
	}

	@Test def folding_multiline_comment() {
		testFolding[
			model = '''
				/**
				 * block comment
				 */
				 entity E {}
			'''
			expectedFoldings = '''
				FoldingRange [
				  startLine = 0
				  endLine = 2
				  startCharacter = 0
				  endCharacter = 24
				  kind = "comment"
				]'''
		]
	}

	@Test def folding_one_entity() {
		testFolding[
			model = '''
				datatype DataType
				entity Entity1 {
					f : DataType
				}
			'''
			expectedFoldings = '''
				FoldingRange [
				  startLine = 1
				  endLine = 3
				  startCharacter = 18
				  endCharacter = 50
				  kind = null
				]'''
		]
	}

	@Test def folding_two_entities() {
		testFolding[
			model = '''
				datatype DataType
				entity Entity1 {
					f : DataType
				}
				entity Entity2 {
					f : DataType
				}
			'''
			expectedFoldings = '''
				FoldingRange [
				  startLine = 1
				  endLine = 3
				  startCharacter = 18
				  endCharacter = 50
				  kind = null
				]FoldingRange [
				  startLine = 4
				  endLine = 6
				  startCharacter = 51
				  endCharacter = 83
				  kind = null
				]'''
		]
	}
	
	@Test def folding_all_in() {
		testFolding[
			model = '''
				/**
				 * block comment
				 */
				datatype DataType
				entity Entity1 {
					f : DataType
				}
				entity Entity2 {
					f : DataType
				}
			'''
			expectedFoldings = '''
				FoldingRange [
				  startLine = 4
				  endLine = 6
				  startCharacter = 43
				  endCharacter = 75
				  kind = null
				]FoldingRange [
				  startLine = 7
				  endLine = 9
				  startCharacter = 76
				  endCharacter = 108
				  kind = null
				]FoldingRange [
				  startLine = 0
				  endLine = 2
				  startCharacter = 0
				  endCharacter = 24
				  kind = "comment"
				]'''
		]
	}
}
```

All test cases succeed, so the calculation of block comment and entity folding seems to be correct.
screenshot

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/folding/AbstractFoldingRangeProviderTest.java
```diff
@@ -0,0 +1,75 @@
...
+			FoldingRange range = foldingRanges.get(i);
+			FoldingRange expected = expectedRanges[i];
+			Assert.assertEquals(expected.getOffset(), range.getOffset());
+			Assert.assertEquals(expected.getLength(), range.getLength());
+			Assert.assertEquals(expected.getKind(), range.getKind());
+		}
+	}
+
+	private XtextResource parse(String doc) throws Exception {
+		URI uri = URI.createURI("dummy." + fileExtension.getPrimaryFileExtension());
```
szarnekow, line 68: You can probably use an injected ParseHelper to simplify this code a little bit.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/folding/FoldingRangeProviderTest.java
```diff
@@ -0,0 +1,56 @@
...
+public class FoldingRangeProviderTest extends AbstractFoldingRangeProviderTest {
+
+	@Test
+	public void foldingRangesProvider() {
+		String actual = "package testLanguage {"
+				+ "\n"
+				+ "\n";
+		
+		int commentIndex = actual.length();
+
```
szarnekow, line 33: Nit: A lot a blank lines here. Not sure why. Not important to adjust, though.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/folding/FoldingRangeProviderTest.java
```diff
@@ -0,0 +1,56 @@
...
+	@Test
+	public void foldingRangesProvider() {
+		String actual = "package testLanguage {"
+				+ "\n"
+				+ "\n";
+		
+		int commentIndex = actual.length();
+		
+		actual += "/*\n"
```
szarnekow, line 34: Generally speaking (also not important in this context here): A StringBuilder may be handy when it comes to String concatenation.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * @author Mark Sujew - Initial contribution and API
+ * 
+ * @since 2.25
+ */
+public class FoldingRange {
+
+	private int offset;
+	private int length;
+	private String kind;
```
szarnekow, line 20: If we agree on moving this up as API, please add a few words of JavaDoc, e.g. what is the purpose of `kind`. Should the constants from FoldingRangeKind be moved here to avoid the otherwise useless type FoldingRangeKind? Or should kind be of type FoldingRangeKind and that one be a wrapper around String?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
```diff
@@ -0,0 +1,27 @@
...
+ * @author Mark Sujew - Ported to IDE project
+ * 
+ * @since 2.25
+ */
+public interface IFoldingRangeAcceptor<T> {
+
+	void accept(int offset, int length);
+	void accept(int offset, int length, String kind);
+	void accept(int offset, int length, boolean initiallyFolded);
+	void accept(int offset, int length, T addedParams);
```
szarnekow, line 24: What is the purpose of the type parameter T?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
```diff
@@ -0,0 +1,141 @@
...
+		Collection<FoldingRange> result = Sets.newLinkedHashSet();
+		IFoldingRangeAcceptor<?> foldingRegionAcceptor = createAcceptor(xtextDocument, result);
+		computeObjectFolding(xtextDocument, foldingRegionAcceptor, cancelIndicator);
+		computeCommentFolding(xtextDocument, foldingRegionAcceptor);
+		return result;
+	}
+
+	protected IFoldingRangeAcceptor<?> createAcceptor(XtextResource xtextDocument,
+			Collection<FoldingRange> foldedPositions) {
+		return new DefaultFoldingRangeAcceptor(foldedPositions);
```
szarnekow, line 60: The returned acceptor is of type FoldingRangeAcceptor<?> which implies that the accept methods that take a `T additionalParam` cannot be called. What is their purpose?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
```diff
@@ -0,0 +1,55 @@
...
+package org.eclipse.xtext.ide.editor.folding;
+
+/**
+ * @author Mark Sujew - Initial contribution and API
+ * 
+ * @since 2.25
+ */
+public class FoldingRange {
+
+	private int offset;
```
szarnekow, line 18: We collect the foldings in a HashSet by default - should this implement hashCode and equals?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ * 
+ * @since 2.25
+ */
+@ImplementedBy(DefaultFoldingRangeProvider.class)
+public interface IFoldingRangeProvider {
+
+	/**
+	 * @return the list of <code>FoldedPosition</code> for the given document
+	 */
+	Collection<FoldingRange> getFoldingRanges(XtextResource xtextDocument, CancelIndicator cancelIndicator);
```
szarnekow, line 35: Since we are effectively creating a new API, can this return a more specific type, e.g. Set<FoldingRange> or SortedSet<FoldingRange> to assert the ordering properties? Would that be useful?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
```diff
@@ -1019,6 +1023,25 @@ protected IResourceServiceProvider getResourceServiceProvider(URI uri) {
 		options.setCancelIndicator(cancelIndicator);
 		return renameService.prepareRename(options);
 	}
+	
+	/**
+	 * @since 2.25
+	 */
+	public CompletableFuture<List<FoldingRange>> foldingRange(FoldingRangeRequestParams params) {
```
szarnekow, line 1030: Here we use a List of FoldingRange - does it make sense to document if we expect multiple folding ranges on the same line?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
```diff
@@ -263,6 +266,7 @@ protected ServerCapabilities createServerCapabilities(InitializeParams params) {
 		serverCapabilities.setDocumentFormattingProvider(true);
 		serverCapabilities.setDocumentRangeFormattingProvider(true);
 		serverCapabilities.setDocumentHighlightProvider(true);
+		serverCapabilities.setFoldingRangeProvider(true);
```
szarnekow, line 269: The server capabilities should not be set if we expect that we may see an UnsupportedOperationException. Maybe check if any of the available languages has a folding provider?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
```diff
@@ -1019,6 +1023,25 @@ protected IResourceServiceProvider getResourceServiceProvider(URI uri) {
...
+	}
+	
+	/**
+	 * @since 2.25
+	 */
+	protected List<FoldingRange> foldingRange(FoldingRangeRequestParams params, CancelIndicator cancelIndicator) {
+		URI uri = getURI(params.getTextDocument());
+		FoldingRangeService foldingRangeService = getService(uri, FoldingRangeService.class);
+		if (foldingRangeService == null) {
+			throw new UnsupportedOperationException();
```
szarnekow, line 1041: I think an empty list would be more user friendly if the service is not implemented

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
```diff
@@ -0,0 +1,52 @@
...
+
+	@Inject
+	private IFoldingRangeProvider foldingRangeProvider;
+
+	public List<FoldingRange> createFoldingRanges(Document document, XtextResource resource,
+			CancelIndicator cancelIndicator) {
+		return foldingRangeProvider.getFoldingRanges(resource, cancelIndicator).stream()
+				.map(range -> toFoldingRange(document, range))
+				.filter(range -> range.getStartLine() < range.getEndLine()).collect(Collectors.toList());
```
szarnekow, line 37: Question: The filter is used to avoid folding ranges that are on a single line?
Should we have a method `isValidRange` that can be specialised?

Should this use `FoldingRangeClientCapabilities.lineFoldingsOnly`?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	protected FoldingRange toFoldingRange(Document document, org.eclipse.xtext.ide.editor.folding.FoldingRange range) {
+		int offset = range.getOffset();
+		int length = range.getLength();
+		int endOffset = offset + length;
+		Position start = document.getPosition(offset);
+		Position end = document.getPosition(endOffset);
+		FoldingRange result = new FoldingRange(start.getLine(), end.getLine());
+		result.setStartCharacter(offset);
```
szarnekow, line 47: I may be wrong here, but I understand the spec as these chars being columns in the folded line. Can you confirm that these are supposed to be relative to the document?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
```diff
@@ -0,0 +1,52 @@
...
+				.map(range -> toFoldingRange(document, range))
+				.filter(range -> range.getStartLine() < range.getEndLine()).collect(Collectors.toList());
+	}
+
+	protected FoldingRange toFoldingRange(Document document, org.eclipse.xtext.ide.editor.folding.FoldingRange range) {
+		int offset = range.getOffset();
+		int length = range.getLength();
+		int endOffset = offset + length;
+		Position start = document.getPosition(offset);
+		Position end = document.getPosition(endOffset);
```
szarnekow, line 45: Since the endOffset is exclusive, it may be on the next line in the document. I think this needs refinements as the endline is by default fully folded (e.g. if endCharacter is omitted). Do we need to provide the last folded character instead?

szarnekow (changes requested):
Nice work! I've read through the change and annotated the things that caught my attention. Not all of these are strictly actionable items but more like curious findings. Please feel free to explicitly reject what you feel to be not necessary.

What I'd like you to double-check though is the contract of the folding ranges in LSP. I read the spec differently from what I understand to be implemented here. Please add clarifying comments where I misread the spec and fix the things that do not match the protocol, if any.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
Darkgaja: Yes, you're right. I somehow got this wrong.

Darkgaja: Wait, on second rereading, it sounds different again: "If not defined, defaults to the length of the start line." Is the *length* here meant as the length of the whole document string or as the length of the line?

szarnekow: To me, `defaults to the length of the start line` sounds like the length of the (start) line ;)

miklossy: > defaults to the length of the start line
I understand it as the length of the line the `FoldingRange.startLine` variable refers to.

Darkgaja: I've looked into different language server implementations ([java](https://github.com/eclipse/eclipse.jdt.ls/blob/3a52306c1eda1c409c68f5d80bb0d64e518c2233/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/FoldingRangeHandler.java#L139), [ts](https://github.com/theia-ide/typescript-language-server/blob/3a8ea0f04bde047e2f60b9f68ef7816e3a41aa10/server/src/lsp-server.ts#L887-L912), [R](https://github.com/REditorSupport/languageserver/blob/21d796ba4167ab457cace13b6096121adde3e6d9/R/folding.R#L73-L77)) and neither actually use the `startCharacter` or `endCharacter` properties. So I don't have any reference how everything is actually meant.

Is there an actual benefit of implementing it with this PR? Not even vscode can actually handle these properties (returns `lineFoldingsOnly=true`)

cdietrich: would you please be so kind and ask it as a clarifining question issue in the protocol github repo?

Darkgaja: Alright https://github.com/microsoft/language-server-protocol/issues/1198

szarnekow: > Is there an actual benefit of implementing it

When in doubt leave it out

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
Darkgaja: The purpose is to keep compatibility with the original folding implementation of [xtext-eclipse](https://github.com/eclipse/xtext-eclipse/tree/master/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/folding), which as well contains a type parameter.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
Darkgaja: True, that doesn't make a lot of sense. I'll refactor some methods of the `DefaultFoldingRangeProvider` into an `AbstractFoldingRangeProvider<T>` that makes more use of the type parameter.

miklossy:
> @miklossy could you give this a try with your sample dsl

I added folding capabilities to the Domainmodel language, the calculation of block comment and entity folding seems to be correct:

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
szarnekow: In Eclipse, a folding region can carry information about the text that is to be shown if that region is collapsed. This capability does not seem to be present in LSP.
OTOH the LSP protocol has a notion of a `kind` which is not at all relevant in Eclipse.
Am I right in the assumption that you are implementing the superset of both frontend's capabilities? Would it make sense to use ITextRegion directly instead of the type parameter? It would also allow to give it a better name compared to `addedParams`

Darkgaja: Yes, implementing a superset of both capabilities was the idea behind that. I thought about using ITextRegion as well. However, I tried not to break compatibility for future would-be implementors of the new API that didn't use ITextRegion in the past.

I thought about the intersection based solution as well. However, I found it to be quite bothersome to build acceptors for each frontend solution. Alternatively, an `IFoldingRangeAcceptor` could accept any FoldingRange implementation, instead of accepting the individual parts of a FoldingRange. What do you think of this?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
Darkgaja: I removed the start- and endCharacter setters. I've created #1675 as a reminder.

@miklossy It may be necessary to update your tests, as the ordering of the FoldingRanges may have changed as well (using a `SortedSet` to hold them now)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
szarnekow: I've a weak tendency towards an *explicit* superset of both known worlds, too.

In Eclipse, we have the `org.eclipse.jface.text.source.projection.IProjectionPosition` and in LSP the `FoldingRange`.
Xtext simplifies the IProjectionPosition by chosing exactly one text region that is supposed to stay whereas the are before and after that region is the part of the document that will not be visible if the range is collapsed.
Apart from that, Xtext offers a flag `initiallyCollapsed`.
LSP does not offer these but has a notion of a `kind` instead.
So our superset could be modelled as `offset, length, initiallyFolded, kind, visualPlaceholderRegion(=ITextRegion)`. A little annoying since the intersection of both frontends is only the tuple of `offset, length`. 

Let's assume we use `FoldingRange` in the acceptor. We'd still want to be able to implement folding only once and benefit the most on both platforms. So FoldingRange would support the superset. Essentially we'd replace the accept(..) methods by overloaded constructors or factory methods. That would only add another layer of indirection. It'd render the acceptor obsolete since the API could be based on a SortedSet directly. The disadvantage: The acceptor pattern allows to create a pseudo-smart acceptor, e.g. one that will only accept regions larger than X, that will drop regions after n other regions have already been registered or that will even do cancellation checks internally and drop out automatically. We probably want to stick to the acceptor based solution with the plenty of overloaded accept methods. Maybe it's sufficient to add reasonable JavaDoc about the limitations of both platforms on the fully fledged overload of accept? And dropping the type parameter in favour of a concrete ITextRegion should be possible if I'm not mistaken.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
szarnekow: Just saw that we have JavaDoc on the interface. I think it's vague enough and good enough at the same time.

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -742,7 +747,7 @@ abstract class AbstractLanguageServerTest implements Endpoint {
 			textDocument = new TextDocumentIdentifier(fileInfo.uri)
 		]).get()
 		
-		assertEqualsStricter(configuration.expectedFoldings, String.join("", foldings.map[toString]));
+		assertEqualsStricter(configuration.expectedFoldings, String.join("", foldings.map[toExpectation]));
```
miklossy, line 750: Instead of using `String.join("", foldings.map[toExpectation])`, I would prefer using `foldings.toExpectation`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
miklossy: ok, thanks for the reminder. I will try it out now.

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
miklossy: Thanks for fixing it @Darkgaja , it looks really good now!

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
Darkgaja: Ok, so I now implemented everything as a superset. Sadly, I couldn't reproduce the exact behavior of [this code](https://github.com/eclipse/xtext-eclipse/blob/39c7b77e5100dc15284cd30c20f20cdf15b110e5/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/folding/DefaultFoldingRegionAcceptor.java#L43-L50), since I couldn't figure out how to get the length of a certain line without access to `IXtextDocument` or `Document`. Is it possible to figure this out via the NodeModel?

miklossy:
I have just tested the PR once again based on the Xtext Domainmodel Tutorial.

I implemented the following test cases, they all succeed.

```Xtend
class DomainmodelFoldingRangeLanguageServerTest extends AbstractLanguageServerTest {

	new() {
		super("dmodel")
	}

	@Test def folding_multiline_comment() {
		testFolding[
			model = '''
				/**
				 * block comment
				 */
				 entity E {}
			'''
			expectedFoldings = '''
				FoldingRange [
				  startLine = 0
				  endLine = 2
				  startCharacter = null
				  endCharacter = null
				  kind = "comment"
				]
			'''
		]
	}

	@Test def folding_one_entity() {
		testFolding[
			model = '''
				datatype DataType
				entity Entity1 {
					f : DataType
				}
			'''
			expectedFoldings = '''
				FoldingRange [
				  startLine = 1
				  endLine = 3
				  startCharacter = null
				  endCharacter = null
				  kind = null
				]
			'''
		]
	}

	@Test def folding_two_entities() {
		testFolding[
			model = '''
				datatype DataType
				entity Entity1 {
					f : DataType
				}
				entity Entity2 {
					f : DataType
				}
			'''
			expectedFoldings = '''
				FoldingRange [
				  startLine = 1
				  endLine = 3
				  startCharacter = null
				  endCharacter = null
				  kind = null
				]
				FoldingRange [
				  startLine = 4
				  endLine = 6
				  startCharacter = null
				  endCharacter = null
				  kind = null
				]
			'''
		]
	}
	
	@Test def folding_all_in() {
		testFolding[
			model = '''
				/**
				 * block comment
				 */
				datatype DataType
				entity Entity1 {
					f : DataType
				}
				entity Entity2 {
					f : DataType
				}
			'''
			expectedFoldings = '''
				FoldingRange [
				  startLine = 0
				  endLine = 2
				  startCharacter = null
				  endCharacter = null
				  kind = "comment"
				]
				FoldingRange [
				  startLine = 4
				  endLine = 6
				  startCharacter = null
				  endCharacter = null
				  kind = null
				]
				FoldingRange [
				  startLine = 7
				  endLine = 9
				  startCharacter = null
				  endCharacter = null
				  kind = null
				]
			'''
		]
	}
}
```

I also add an LSP Client based on LSP4E and the Generic Text Editor, the folding works there as well (currently LSP4E does not initialize the folding capabilities properly, [bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=571162) has been reported).
screenshot

The folding functionality seems to be working on both test side and client side.

Two additional remarks:

1. consider marking this new API as Beta (with the `com.google.common.annotations.Beta` annotation) to be able to modify the API later in the future with no risk.
2. squash your commits into a single one (or use Squash and Merge when merging) to keep the history a little bit cleaner

I am approving this PR now, @Darkgaja thanks a lot for your contribution!

miklossy (approved):

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
szarnekow: @Darkgaja  IIUIC the code will alter invalid significant regions such that only the first line of the region is taken. So basically it'll start from the significant regions offset until it finds the first line-break or the end-offset of the significant region. It is possible to achieve that based on the root node's text along with NodeModelUtils.getLineAndColumn(INode, int) to compare the line numbers.

szarnekow:
@Darkgaja Your work on this is greatly appreciated and I hope that the feedback loops are ok for you.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeAcceptor.java
Darkgaja: Great, thanks for the info. I implemented it via the getLineAndColumn method. Could you take a look at it? I think I addressed every issue you had.

Darkgaja:
@szarnekow Thank you. I greatly appreciate your feedback on the topic. :)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
szarnekow: @Darkgaja Do you have a response to this question?

Darkgaja: While endOffset is exclusive, if I've understood [this code](https://github.com/eclipse/xtext-core/blob/2b25ff301e6730277b2f270baf29bed2ac490445/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.java#L73-L84) correctly, it shouldn't be a problem. Since the endOffset is then on the newline character, we should still receive the correct line. I am wrong there?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/AbstractFoldingRangeAcceptor.java
```diff
@@ -0,0 +1,47 @@
...
+import org.eclipse.xtext.util.ITextRegion;
+
+/**
+ * @author Mark Sujew - Initial contribution and API
+ */
+public abstract class AbstractFoldingRangeAcceptor implements IFoldingRangeAcceptor {
+
+	@Override
+	public void accept(int offset, int length) {
+		accept(offset, length, false);
```
szarnekow, line 20: Can these all be default methods on the interface?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeAcceptor.java
```diff
@@ -0,0 +1,44 @@
...
+		return result;
+	}
+
+	@Override
+	public void accept(int offset, int length, FoldingRangeKind kind, boolean initiallyFolded,
+			ITextRegion visualPlaceholderRegion) {
+		result.add(getFoldingRange(offset, length, kind, initiallyFolded, visualPlaceholderRegion));
+	}
+
+	protected FoldingRange getFoldingRange(int offset, int length, FoldingRangeKind kind, boolean initiallyFolded,
```
szarnekow, line 40: Nit: createFoldingRange instead of getFoldingRange?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
```diff
@@ -0,0 +1,197 @@
...
+			}
+		}
+	}
+
+	protected void acceptCommentFolding(INode commentNode, IFoldingRangeAcceptor foldingRangeAcceptor) {
+		int offset = commentNode.getOffset();
+		int length = commentNode.getLength();
+		Matcher matcher = getTextPatternInComment().matcher(commentNode.getText());
+		if (matcher.find()) {
+			TextRegion significant = new TextRegion(offset + matcher.start(), 0);
```
szarnekow, line 156: I'm not sure the significant part of a comment is the empty region. Shouldn't this be the first line?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
```diff
@@ -0,0 +1,113 @@
...
+
+	public int getLength() {
+		return length;
+	}
+
+	public void setLength(int length) {
+		this.length = length;
+	}
+
+	public FoldingRange(int offset, int length) {
```
szarnekow, line 68: Nit: Please move constructors to the beginning of the file.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
```diff
@@ -0,0 +1,113 @@
...
+		hash = prime * hash + length;
+		hash = prime * hash + (initiallyFolded ? 1 : 0);
+		if (kind != null) {
+			hash = prime * hash + kind.hashCode();
+		}
+		if (visualPlaceholderRegion != null) {
+			hash = prime * hash + visualPlaceholderRegion.hashCode();
+		}
+		return hash;
+	}
```
szarnekow, line 112: Please add a meaningful toString, too.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * @author Mark Sujew - Initial contribution and API
+ * 
+ * @since 2.25
+ */
+public class FoldingRange {
+
+	private int offset;
+	private int length;
```
szarnekow, line 23: Since the fields are used in equals and hashCode, they should be final.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ * @author Sebastian Zarnekow - Refactoring, introduced FoldedPosition
+ * @author Mark Sujew - Ported to IDE project
+ * 
+ * @since 2.25
+ */
+@ImplementedBy(DefaultFoldingRangeProvider.class)
+public interface IFoldingRangeProvider {
+
+	/**
+	 * @return the list of <code>FoldedPosition</code> for the given document
```
szarnekow, line 33: JavaDoc is sort-of off with the signature.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
szarnekow: Sounds correct to me. How does the folding look like in VS Code e.g. if there is a meaningful char after a closed ML_COMMENT?

```
/*
 * fold me
 */identifier
```

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -407,6 +409,10 @@ abstract class AbstractLanguageServerTest implements Endpoint {
 	protected dispatch def String toExpectation(CodeLens it) {
 		return command.title + " " +range.toExpectation
 	}
+	
+	protected dispatch def String toExpectation(FoldingRange it) {
+		return toString
```
szarnekow, line 414: Question: For CodeLens above we do invoke toExpectation on the range. Should we do something similar for the positions in the range?

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/FoldingConfiguration.java
```diff
@@ -0,0 +1,15 @@
...
+
+	private String expectedFoldings;
+
+	public String getExpectedFoldings() {
+		return expectedFoldings;
+	}
+
+	public void setExpectedFoldings(String expectedFoldings) {
+		this.expectedFoldings = expectedFoldings;
```
szarnekow, line 12: Question: Don't we expect multiple foldings?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
```diff
@@ -0,0 +1,113 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.editor.folding;
+
+import org.eclipse.xtext.util.ITextRegion;
+
+import com.google.common.base.Objects;
+
+/**
+ * @author Mark Sujew - Initial contribution and API
```
szarnekow, line 16: Please add a comment indicating the reason why this type is present, e.g. abstraction over FoldedPosition and LSP FoldingRange

szarnekow (commented):
We are getting there though I feel it's too late for the next release.

Some minor things are still pending. Please mark all open discussion that you don't want to address with a short comment.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
Darkgaja: I just took the original code from [here](https://github.com/eclipse/xtext-eclipse/blob/aa0272eb77c5476e2a245812c5362628bc10d2d1/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/folding/DefaultFoldingRegionProvider.java#L183-L194).

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
Darkgaja: The `identifier` part gets folded as well:
grafik

szarnekow: Interesting. What happens if we also send an end char?

cdietrich:
we will move this to 2.26 to give some more time to shape this PR.
thx for your work so far @Darkgaja  @szarnekow

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/FoldingConfiguration.java
Darkgaja: We do, but I just took the same approach as the [completion testing](https://github.com/eclipse/xtext-core/blob/2027567c0115cda8a6dd071f44ebb7615640991e/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/CompletionTest.java#L52-L63).

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/AbstractFoldingRangeAcceptor.java
Darkgaja: You're right. I removed the abstract acceptor.

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
Darkgaja: Right, that simplified folding range testing a lot :D

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeAcceptor.java
tivervac: Agriid

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
tivervac: Well, this means there is no text in the comment. The significant region for folding is probably used to show some text upon hovering, if there's no text in the comment there's not much to be shown on hover IMO so creating an empty significant region might be ok.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRange.java
```diff
@@ -0,0 +1,104 @@
...
+			return offset == range.offset && length == range.length && initiallyFolded == range.initiallyFolded
+					&& Objects.equal(kind, range.kind)
+					&& Objects.equal(visualPlaceholderRegion, range.visualPlaceholderRegion);
+		}
+		return false;
+	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
```
tivervac, line 81: FYI, there's also Objects.hashCode which would likely result in a similarly strong hashcode

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/FoldingRangeKind.java
```diff
@@ -0,0 +1,46 @@
...
+
+	@Override
+	public String toString() {
+		return kind;
+	}
+
+	@Override
+	public boolean equals(Object other) {
+		if (other instanceof FoldingRangeKind) {
+			return this.kind == ((FoldingRangeKind) other).kind;
```
tivervac, line 37: Sure this shouldn't be .equals?

tivervac (changes requested):

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
```diff
@@ -0,0 +1,197 @@
...
+			acceptCommentNodes(rootASTElement, foldingRangeAcceptor);
+		}
+	}
+
+	protected void acceptCommentNodes(EObject eObject, IFoldingRangeAcceptor foldingRangeAcceptor) {
+		ICompositeNode node = NodeModelUtils.getNode(eObject);
+		if (node != null) {
+			for (INode leafNode : node.getAsTreeIterable()) {
+				if (leafNode.getGrammarElement() instanceof TerminalRule
+						&& ruleName.equalsIgnoreCase(((TerminalRule) leafNode.getGrammarElement()).getName())) {
```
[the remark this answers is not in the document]
Darkgaja: As per the [documentation](https://google.github.io/guice/api-docs/4.2/javadoc/com/google/inject/Inject.html), the injector will competely ignore the field and the default value will remain:

>When applied to a field, any default value already assigned to the field will remain (guice will not actively null out the field).

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
```diff
@@ -0,0 +1,197 @@
...
+				}
+			}
+		}
+	}
+
+	protected void acceptObjectFolding(EObject eObject, IFoldingRangeAcceptor foldingRangeAcceptor) {
+		ITextRegion region = locationInFileProvider.getFullTextRegion(eObject);
+		if (region != null) {
+			INode eObjectNode = NodeModelUtils.getNode(eObject);
+			ITextRegion significant = buildSignificantRegion(locationInFileProvider.getSignificantTextRegion(eObject),
```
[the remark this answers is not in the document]
Darkgaja: I added a guard in the `buildSignificantRegion` method.

cdietrich:
any update here (also needs a rebase)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
Darkgaja: Good question. I'd imagine the `identifier` part would still be visible. However, I currently don't have a test setup available where I could easily test this. Can we postpone this question for #1675? (which I currently don't have the time for)

szarnekow: Sure

szarnekow (approved):
Thank you for working on this! Nicely done!

cdietrich:
@tivervac can you please re-review. thx

tivervac:
@cdietrich on holiday now, will look at it monday if it hasn’t been merged yet.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeAcceptor.java
tivervac: @Darkgaja Could you fix this?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
```diff
@@ -0,0 +1,202 @@
+/*******************************************************************************
+ * Copyright (c) 2010 Michael Clay and others.
```
tivervac, line 2: @Darkgaja Same for this copyright

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeProvider.java
```diff
@@ -0,0 +1,202 @@
...
+	/**
+	 * @return <code>true</code> if the object should be folded if it spans more than one line. Default is
+	 *         <code>false</code> if and only if the object is the root object of the resource.
+	 */
+	protected boolean isHandled(EObject eObject) {
+		return eObject.eContainer() != null;
+	}
+
+	/**
+	 * @return clients should <code>false</code> to abort the traversal of the model.
```
tivervac, line 197: s/should/should return/

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/IFoldingRangeProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ * @author Sebastian Zarnekow - Refactoring, introduced FoldedPosition
+ * @author Mark Sujew - Ported to IDE project
+ * 
+ * @since 2.26
+ */
+@ImplementedBy(DefaultFoldingRangeProvider.class)
+public interface IFoldingRangeProvider {
+
+	/**
+	 * @return the set of <code>FoldingRange</code> for the given document
```
tivervac, line 33: s/<code>FoldingRange</code>/<code>FoldingRange</code>s/

tivervac (approved):
Few typos and old licenses but looks good other that that

Darkgaja:
@tivervac Addressed the issues, thanks for all the feedback, everyone.

cdietrich:
many thanks for the contribution @Darkgaja

PR #1674, opened 2021-02-10
Allow for custom modifications in rename service

msujew:
For the reason behind this PR, see eclipse/xtext#2474

The proposed changes allow for users to customize their renaming service further by operating directly on the `WorkspaceEdit` without overriding the whole `rename` method. This is necessary for cases where changes aren't picked up by the `ChangeSerializer` and related services.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.java
@@ -125,7 +125,7 @@ public class RenameService2 implements IRenameService2 {
 						issueAcceptor.add(RefactoringIssueAcceptor.Severity.FATAL,
 								"No element found at " + toPositionFragment(position, uri));
 					} else {
-						applyModifications(element, workspaceEdit, issueAcceptor, options);
+						applyModifications(element, workspaceEdit, issueAcceptor, options, context);
 					}
 				} else {
 					issueAcceptor.add(RefactoringIssueAcceptor.Severity.FATAL,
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.java
```diff
@@ -197,6 +198,20 @@ protected boolean isIdentifier(ILeafNode leafNode) {
...
+	 * Override this method to supplement custom modifications on a {@link WorkspaceEdit}. 
+	 * This method is called after every other modification has been processed.
+	 * 
+	 * @param eObject Originally renamed EObject
+	 * @param edit Allows for additional changes to be added 
+	 * @param languageServerAccess Access to the language server
+	 * 
+	 * @since 2.25
+	 */
+	protected void applyCustomModifications(EObject eObject, WorkspaceEdit edit, ILanguageServerAccess languageServerAccess) {
```
szarnekow, line 212: The rename options themselves would be handy.
It'd also make sense to move the usage of the ChangeSerializer (lines 128-141) into an own method that can be specialized for languages, that do not have other means to perform renames. This would probably render the applyCustomModifications obsolete since custom stuff could be done by overriding + calling super

Darkgaja: Thanks for the feedback. I'd agree, that's definitely a cleaner solution. I've implemented it this way.

szarnekow (approved):
Apart from the missing `@since` tag on the new method, this looks ok to me.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.java
```diff
@@ -163,6 +150,24 @@ public WorkspaceEdit rename(IRenameService2.Options options) {
 			throw Exceptions.sneakyThrow(e);
 		}
 	}
+	
+	protected void applyModifications(EObject element, WorkspaceEdit workspaceEdit, ServerRefactoringIssueAcceptor issueAcceptor, IRenameService2.Options options) {
```
szarnekow, line 154: I know this does already have an aweful signature, but the languageserver.context could also be useful to implement a reasonable rename strategy. Also this method deserves some JavaDoc that explains, that the element is from a live-scope-RS and this entire thing runs in a read-transaction on the index

cdietrich:
thx @Darkgaja

Issue #1684, opened 2021-03-01
[LSP] META-INF/services/org.eclipse.xtext.ISetup overridden when using multiple languages


andreasmuelder:
When using more than one language and bundling the plugins as a fat jar, the maven-shade-plugin overrides the contents of 'org.eclipse.xtext.ISetup' an thus only one language service is started properly. This can be fixed by adding the ServicesResourceTransformer to the pom.xml:
```
<transformer 
  implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
</transformers>
```

cdietrich:
Gradle is also affected. with provide a PR for the gradle part

cdietrich:
fixed in 2.26

PR #1688, opened 2021-03-05
[#1684] fix services merging in maven lsp fat jars

cdietrich:
[#1684] fix services merging in maven lsp fat jars


kthoms:
Looks good, but is there an example that I could use for proving this?

kthoms (approved):

Issue #1652, opened 2020-12-18
Update to LSP4J.next (final 3.16 version)


cdietrich:
once there is a version of LSP4J available that reflects the final 3.16 protocol version,
we should update

see also https://github.com/eclipse/xtext/issues/1917

cdietrich:
current semantic highlighting will be dropped with that version. cc @szarnekow

cdietrich:
likely not 2021-03/2.25

cdietrich:
ServerCapabilities might be changed

cdietrich:
lsp4j 0.12.0 is coming

cdietrich:
done

PR #1694, opened 2021-03-22
[eclipse/xtext-core#1652] adapt to lsp4j 0.11.0

cdietrich:
[eclipse/xtext-core#1652] adapt to lsp4j 0.11.0


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
```diff
@@ -1144,18 +1131,6 @@ public InitializeResult getInitializeResult() {
 
 	@Override
 	public void afterBuild(List<IResourceDescription.Delta> deltas) {
```
szarnekow, line 1133: We should remove the implements `IBuildListener` from the server, too.

cdietrich: i wanted to keep it here so that you can simply kick in then you provide the new impl

szarnekow: The final protocol is based on a regular request from the client. I don't the afterBuild will be used for that.

cdietrich: ok. what about subclassing clients?

szarnekow: Hmmm maybe mark it at least as deprecated?

cdietrich: better so?

szarnekow: I think so, yes.

cdietrich:
replaced by https://github.com/eclipse/xtext-core/pull/1696

Issue #1703, opened 2021-04-26
org.eclipse.xtext.preferences is not api, but testing api uses it


cdietrich:
org.eclipse.xtext.preferences (e.g. MapBasedPreferenceValues) is not api (it is provisional), but testing api uses it

is there anything that speaks against making it api?

kthoms:
What is the purpose of making it API? Which kind of clients would need to access these classes?

The only classes I would consider to possibly be API are `IPreferenceValues`, `IPreferenceValuesProvider` and `PreferenceKey`.

How about moving the others to an internal package. However, we would risk that clients break which are using those restricted classes on purpose.

cdietrich:
see org.eclipse.xtext.testing.formatter.FormatterTestRequest.preferences(Procedure1<MapBasedPreferenceValues>)

kthoms:
Yes, quite unfortunate. The method is no good, but why exposing the dirt even more?
The method getOrCreateMapBasedPreferences might get deprecated and another one provided that expose only the interface ITypedPreferences. The clients could assume to cast if really needed. I doubt there are some that really rely on having MapBasedPreferenceValues.

cdietrich:
all tests that use it do and would break

szarnekow:
>  I doubt there are some that really rely on having MapBasedPreferenceValues.

Quite the contrary. MapBasedPreferenceValues is the only API that allows to *set* preferences in the runtime case.

szarnekow:
> What is the purpose of making it API? Which kind of clients would need to access these classes?

Putting it the other way round: Why keeping this internal? There haven't been any changes in ages.

kthoms:
I have raised the questions that came in my mind, but if you feel comfortable with exposing this, then I don't care much. Please decide.

cdietrich:
done

PR #1704, opened 2021-04-26
[#1703] expose org.eclipse.xtext.preferences as api

cdietrich:
[#1703] expose org.eclipse.xtext.preferences as api


szarnekow:
Just checked the code in that package. Looks good in general. One thing that bugs me is the error handling in the concrete incarnations of `TypedPreferenceKey`, e.g. `IntegerKey` will log a warning in case of `NumberFormatException` and return the default value, `BooleanKey` will use the conversion rules of `Boolean.valueOf` thus often not using the default value even though the value string may be not a boolean representation. StringKey will happily return null instead of the default value if the given string is null. It's not really consistent.

Long story short: Should we try to make this a little better before we publish it as API in its current shape?

cdietrich:
good question. i did not really look into it.

szarnekow (approved):

PR #1713, opened 2021-06-21
Fixed typo, NPE and small cleanup

tivervac:
comment.getGrammarElement() can be null thus toString should be able
to handle nulls

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
@@ -275,14 +275,6 @@ public class TextRegionAccessToString {
 	}
 
 	/**
-	 * @deprecated Use {@link #highlightOrigin()} instead
-	 */
-	@Deprecated
-	public TextRegionAccessToString hightlightOrigin() {
-		return highlightOrigin();
-	}
-
-	/**
 	 * @since 2.26
 	 */
 	public TextRegionAccessToString highlightOrigin() {
@@ -302,14 +294,6 @@ public class TextRegionAccessToString {
 
 	public boolean isHideIndentation() {
 		return hideIndentation;
-	}
-
-	/**
-	 * @deprecated use {@link #isHighlightOrigin()} instead
-	 */
-	@Deprecated
-	public boolean isHightlightOrigin() {
-		return isHighlightOrigin();
 	}
 
 	/**
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
@@ -8,6 +8,7 @@
  *******************************************************************************/
 package org.eclipse.xtext.formatting2.debug;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -128,7 +129,7 @@ public class TextRegionAccessToString {
 			for (ITextSegmentDiff diff : this.access.getRegionDifferences()) {
 				ISequentialRegion current = toSequential(diff.getOriginalFirstRegion());
 				ISequentialRegion last = toSequential(diff.getOriginalLastRegion());
-				List<ITextSegment> regions = Lists.newArrayList();
+				List<ITextSegment> regions = new ArrayList<>();
 				while (current != null) {
 					regions.add(current);
 					if (current.getOffset() >= last.getOffset()) {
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
```diff
@@ -273,8 +274,19 @@ public TextRegionAccessToString hideIndentation() {
...
+	 * @deprecated Use {@link #highlightOrigin()} instead
+	 */
+	@Deprecated
 	public TextRegionAccessToString hightlightOrigin() {
-		this.hightlightOrigin = true;
+		return highlightOrigin();
+	}
+
+	/**
+	 * @since 2.25
```
cdietrich, line 286: `2.26` ?

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
```diff
@@ -293,7 +305,7 @@ public boolean isHideIndentation() {
 	}
 
 	public boolean isHightlightOrigin() {
-		return hightlightOrigin;
+		return highlightOrigin;
```
cdietrich, line 308: do we need a new version deprecation here too?

cdietrich:
where is the actual fix for the NPE?

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
tivervac: Fixed

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
```diff
@@ -349,14 +361,17 @@ protected String toString(EObject ele) {
 			return grammarToString.apply((AbstractElement) ele);
 		if (ele instanceof AbstractRule)
 			return ele.eClass().getName() + "'" + ((AbstractRule) ele).getName() + "'";
+		if (ele == null) {
```
tivervac, line 364: @cdietrich This is the NPE fix

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
tivervac: We do, didn't notice that one

tivervac: Done

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
```diff
@@ -292,8 +304,19 @@ public boolean isHideIndentation() {
 		return hideIndentation;
 	}
 
+	/**
+	 * @deprecated use {@link #isHighlightOrigin()} instead
+	 */
+	@Deprecated
 	public boolean isHightlightOrigin() {
```
cdietrich, line 311: as this is provisional api: i wonder if we need the deprecated. @szarnekow what do you think?

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
cdietrich: @tivervac  any hint on how to get an IComment without grammar element?

tivervac: In our case it was a syntax error. We have a hidden terminal rule that looks somewhat like this
```
terminal MACRO: '`' ('a'..'z');
```

When you then have code that only has the backtick but no body we would have an IComment without grammar element because the node behind the IComment was a `HiddenLeafNodeWithSyntaxError`.

```
package p;
  `
endpackage
```

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
szarnekow: I think we can remove the method from this provisional debug util.

szarnekow (approved):
Approved as is, but the deprecation may be not worth it. OTOH it's not a lot of dancing to get it right.

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/debug/TextRegionAccessToString.java
tivervac: Removed the deprecated method

cdietrich (approved):

cdietrich (approved):

PR #1717, opened 2021-07-07
[eclipse/xtext#1976] workaround for platform version range issue in gradle

cdietrich:
[eclipse/xtext#1976] workaround for platform version range issue in gradle


cdietrich:
fix someone may review this in the next 200 years and finds a problem, please create a new issue

PR #1720, opened 2021-07-08
Improve performance of StateBasedContainer

rubenporras:
Improve performance of StateBasedContainer by calling
state.getContents() only once.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
@@ -53,16 +53,12 @@ public class StateBasedContainer extends ResourceDescriptionsBasedContainer {
 	@Override
 	protected Iterable<IEObjectDescription> filterByURI(Iterable<IEObjectDescription> unfiltered) {
 		return Iterables.filter(unfiltered, new Predicate<IEObjectDescription>() {
-			private Collection<URI> contents = null;
+			private Collection<URI> contents = getState().getContents();
 
 			@Override
 			public boolean apply(IEObjectDescription input) {
-				if(contents == null) {
-					contents = getState().getContents();
-				}
 				URI resourceURI = input.getEObjectURI().trimFragment();
-				final boolean contains = contents.contains(resourceURI);
-				return contains;
+				return contents.contains(resourceURI);
 			}
 		});
 	}


cdietrich:
can you make sure you have ECA signed (am not sure if this works with PRs done from Github Webinterface at all (meanwhile)

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/StateBasedContainer.java
```diff
@@ -53,16 +53,12 @@ protected final IContainerState getState() {
 	@Override
 	protected Iterable<IEObjectDescription> filterByURI(Iterable<IEObjectDescription> unfiltered) {
 		return Iterables.filter(unfiltered, new Predicate<IEObjectDescription>() {
-			private Collection<URI> contents = null;
+			private Collection<URI> contents = getState().getContents();
```
cdietrich, line 56: i wonder how this should be any faster.

rubenporras: will it not eliminate the null check on each call of the apply method? I understand that a null check is not expensive, but we can do the call once to getState().getContents() once when initializing the predicate and eliminate it.

rubenporras: In case it helps to answer the question, we have had this code in producton see the override in (https://github.com/dsldevkit/dsl-devkit/blob/master/com.avaloq.tools.ddk.xtext/src/com/avaloq/tools/ddk/xtext/resource/StateBasedContainerWithHandle.java) since years without any problems.

I would like to start commiting upstream some of the improvements we did over the years, and I decided to start with this one, as it looks like a very small commit.

szarnekow: It may have a subtle impact on the timing. Also getContents() may not always be super cheap and doing it later might (in few cases) avoid it at all if the iterable is not traversed so far.
Can you provide any numbers with the impact of the proposed change?

szarnekow: @rubenporras I understand and greatly appreciate the initiative to upstream some of the goodness.

I'm aware of closed source projects that do have specialications on IContainerState.getContents and in the past I've often been bitten by seemingly small changes.. Do you have battle tested this with other IContainerState implementations?

rubenporras: Hi Sebastian and Christian,

I will run some test with and without the override in the DDK and see if I can gather some numbers, I will come back once I have them.

Thanks

rubenporras: Thanks for the feedback.

My measurements cannot confirm this makes any difference on real case scenarios, so I will close this pull request and remove the override from our specialization.
Regards

szarnekow: Thank you for coming to this so swiftly, Ruben. Sorry to hear that the tuning does not make a big impact (anymore?).

I'm looking forward to your next contribution.

PR #1722, opened 2021-07-21
update to tycho 2.4

cdietrich:
update to tycho 2.4


miklossy (approved):

LorenzoBettini:
For projects using Tycho 2.4.0, shouldn't we remove the dependencies from the `exec-maven-plugin` configuration and rely on the project dependencies (since additional bundles in build.properties are now correctly used)?

cdietrich:
yes but i want to postpone this until @kthoms and @szarnekow also have looked at the problem.

cdietrich:
created https://github.com/eclipse/xtext-core/pull/1723 for discussion

PR #1723, opened 2021-07-21
dont duplicate deps. that are already part of add. bundles in build.properties

cdietrich:
dont duplicate deps. that are already part of add. bundles in build.properties


cdietrich:
are we ok with this change? is the tp specific enhough?

cdietrich:
for testing: https://github.com/itemis/xtext-reference-projects/tree/cd_addBundlesOnlyTest

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
```diff
@@ -36,26 +36,6 @@
...
-					<dependency>
-						<groupId>org.eclipse.xtext</groupId>
-						<artifactId>org.eclipse.xtext.xtext.generator</artifactId>
-						<version>${xtextVersion}</version>
-					</dependency>
-					<dependency>
-						<groupId>org.eclipse.xtext</groupId>
-						<artifactId>org.eclipse.xtext.xbase</artifactId>
-						<version>${xtextVersion}</version>
-					</dependency>
```
cdietrich, line 58: in the log i still see

``` 
2021-07-21T19:14:20.5006666Z 82   WARN  StandaloneSetup    - Skipping conflicting project org.apache.commons.io at 'archive:file:/home/runner/.m2/repository/commons-io/commons-io/2.5/commons-io-2.5.jar!/' and using 'archive:file:/home/runner/.m2/repository/p2/osgi/bundle/org.apache.commons.io/2.8.0.v20210415-0900/org.apache.commons.io-2.8.0.v20210415-0900.jar!/' instead.
```
wonder where this is coming from

cdietrich: maybe from the exec plugin itself?

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
```diff
@@ -36,26 +36,6 @@
 					<cleanupDaemonThreads>false</cleanupDaemonThreads><!-- see https://bugs.eclipse.org/bugs/show_bug.cgi?id=475098#c3 -->
```
LorenzoBettini, line 36: shouldn't we also remove

```xml
<includePluginDependencies>true</includePluginDependencies>
```

?

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
LorenzoBettini: That's not the only occurrence, is it? I've seen these warnings for ages myself, I'm wondering if they are due now to

```xml
<includePluginDependencies>true</includePluginDependencies>
```

which is still there (see my previous comment)

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
```diff
@@ -36,26 +36,6 @@
...
-						<version>${xtextVersion}</version>
-					</dependency>
-					<dependency>
-						<groupId>org.eclipse.xtext</groupId>
-						<artifactId>org.eclipse.xtext.xbase</artifactId>
-						<version>${xtextVersion}</version>
-					</dependency>
 					<dependency>
 						<groupId>org.eclipse.xtext</groupId>
 						<artifactId>xtext-antlr-generator</artifactId>
```
LorenzoBettini, line 41: Oh... we still have this one :(
So maybe we can't get rid of

```xml
<includePluginDependencies>true</includePluginDependencies>
```

couldn't we remove this one as well and put in the TP a reference to the itemis update site?
After all this won't violate anything and the xtext.generator already has an optional dependency on this bundle anyway (IIRC)

LorenzoBettini: I mean something like

```xml
<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
<unit id="de.itemis.xtext.antlr.feature.feature.group" version="2.1.1.v201405091103"/>
<repository location="https://download.itemis.com/updates/releases/2.1.1/"/>
</location>
```

LorenzoBettini (changes requested):

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
cdietrich: if i had any clue why this is set to true in the first place

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
cdietrich: as i have mentioned before i have zero idea on legal impacts. at that time i was zero involved in Xtext

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
cdietrich: looks like this is not working in all cases

cdietrich: i also dont understand if this also referrs to the "extra" plugin dependencies

cdietrich: yes it does. so it wont find the antlr one and use the facade instead
https://pipelines.actions.githubusercontent.com/VGbhG7MqInqDfxY0bKv0qAuPeX3d22owq0UvtAsHMpYjQuu8bd/_apis/pipelines/1/runs/556/signedlogcontent/27?urlExpires=2021-07-23T05%3A09%3A55.7351874Z&urlSigningMethod=HMACV1&urlSignature=lnqMg85bgGGidvw73PGx1a8O5X4CiUjH12Uk2%2BjKvUE%3D
again something i have ZERO idea if it is any good idea

cdietrich: executableDependency s dont seem to support versions

cdietrich: can we postpone this issue?

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
szarnekow: I think for the code we generate with the wizard it is ok to have a dependency to xtext-antlr-generator - I'd favor a dependency to maven central though - for qos reasons.

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
szarnekow: I think it was originally put there to simplify the workflow (scanClasspath=true) which is no longer the default (since a very long time).

LorenzoBettini: Maybe I confused your idea, but I didn't mean `<cleanupDaemonThreads>false</cleanupDaemonThreads>`, I actually meant

```xml
<includePluginDependencies>true</includePluginDependencies>
```

That is, https://www.mojohaus.org/exec-maven-plugin/java-mojo.html#includePluginDependencies

I think it was added because the additional.bundles was not not taken into consideration, and, from what I understand, we don't want things like xtext.generator in the required bundles, since they are used only for running MWE2. So xtext.generator was added to additional.bundles. However, Tycho did not consider that (due to a bug, now fixed), and so we put them in the dependencies of the plugin (but then we had to tell maven exec to also consider plugin dependencies or MWE2 would not find the classes of xtext.generator and friends). If you try to remove that without Tycho 2.4.0 you get failure when running maven exec (it cannot find classes). With Tycho 2.4.0 that's not needed anymore, and I guess it avoids mixing classes from different classpath...

cdietrich: i tried and removing does not work

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
LorenzoBettini: in that case we have to leave with `<includePluginDependencies>true</includePluginDependencies>`

I'm wondering why not simply removing also that dependency: the facade will download the bundle when running MWE2... after all, if someone doesn't want the jar to be downloaded by the facade they already know they must add the lines above to their TP (in that case, since xtext.generator has an optional dependency on the antlr generator, Tycho should take it from the TP when running maven exec).

I'm pretty sure I checked that in my previous experiments when dealing with the original problem. If you want I can double-check that, but only if you're willing to remove `<includePluginDependencies>true</includePluginDependencies>` and the additional maven dep on `xtext-antlr-generator`. If you think that's not worthwhile it's fine with me (I always tweak the Maven POM and the TP as the first thing anyway).

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
LorenzoBettini: I think I lost you here... what is the problem?

cdietrich: solve the antlr problem separately.

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
LorenzoBettini: sorry I lost the workflow of the conversation ;)

removing what? and what does not work?

cdietrich: the pluginDependencies thing.
i also lost overview.
thus i want to solve these stuff separately

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
cdietrich: i still dont trust our (itemis) admins

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
LorenzoBettini: I think in the end it's really your call.
If you want to keep `<includePluginDependencies>true</includePluginDependencies>` then you can keep also the Maven dep on xtext-antlr... for the time being or forever ;)

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
cdietrich: sebastian proposed to look at the new tycho use maven stuff feature and consume the antlr generator this way

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
cdietrich: yes but this still has the commons io problem

LorenzoBettini (approved):
I have already stated my additional proposals for further cleaning, but I'm fine with the current changes

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
LorenzoBettini: You mean the fact that a TP can now refer to a Maven artifact? That could a cool solution (or a legal trick ;) O:)

cdietrich: yes @szarnekow  do you have a hint which tycho version brought that? and how it looks like

cdietrich: https://läubisoft.gmbh/en/articles/using-maven-artifacts-in-pde-rcp-and-tycho-builds/

LorenzoBettini: Yes, but please keep in mind that this only works with recent versions of Eclipse, if a user wants to activate the TP from Eclipse (of course, if the TP is used only during the Maven build, using Tycho 2.4.0 will be enough).

cdietrich: somehow this does not show up my runtimeXtext workspace from dev env although we use m2e 1.18

cdietrich: ahh its a separate feature

org.eclipse.xtext.tests/testdata/wizard-expectations/mavenTychoJ11/mavenTychoJ11.parent/mavenTychoJ11/pom.xml
LorenzoBettini: > in the log i still see
> 
> ```
> 2021-07-21T19:14:20.5006666Z 82   WARN  StandaloneSetup    - Skipping conflicting project org.apache.commons.io at 'archive:file:/home/runner/.m2/repository/commons-io/commons-io/2.5/commons-io-2.5.jar!/' and using 'archive:file:/home/runner/.m2/repository/p2/osgi/bundle/org.apache.commons.io/2.8.0.v20210415-0900/org.apache.commons.io-2.8.0.v20210415-0900.jar!/' instead.
> ```
> 
> wonder where this is coming from

@cdietrich @szarnekow 
Just for the record, after additional experiments, these bad warnings go away by removing

```xml
<includePluginDependencies>true</includePluginDependencies>
```

and by relying only on additional.bundles

cdietrich: yes. but that needs another source for the antlr dependency.
i dont trust our it department and want to avoid to it being ddosed by maven builds

LorenzoBettini: Yes, I remember that. I just wanted to confirm that the warnings are definitely due to mixing plugin dependencies (Maven deps) with Eclipse bundles. From what I recall, we were not completely sure about that

PR #1724, opened 2021-07-21
[eclipse/xtext#1976] workaround for maven tycho + j8 + platform artifacts issues

cdietrich:
[eclipse/xtext#1976] workaround for maven tycho + j8 + platform artifacts issues


cdietrich:
follow up for / depends on https://github.com/eclipse/xtext-core/pull/1723

LorenzoBettini:
@cdietrich so you don't take my proposal into consideration?
(specify and pin deps in xtext.generator)?

cdietrich:
@LorenzoBettini yes and no. we would also have to do this for other plugin.
as i have **nobody** to discuss this issue i am clueless if the manual "ausmultiplizieren" of the the transtive dependencies is a good idea. thus i am hesitating to decide. i am not a maven expert and dont know what a good design would look like
i am also not sure for what plugins we have to do this.

e.g. xtext-maven-plugin, xtend-maven-plugin?
do we need to do it for the old generator too?
or the xtend.core.

please discuss on the main issues
i still hope @szarnekow and @kthoms  will find some minutes to express there expertise and opinion

LorenzoBettini:
Sorry in advance for the outburst...

Why, is manual multiplication (if that's what "ausmultiplizieren" means) of exclusion a good idea?
To be honest I can say that I am a Maven expert, or at least, very close to.
I also spent almost my whole spare time in 3 days to fix the problem with a careful analysis of causes and outputs.
The cause, if it's still not clear, is the dependency on EMF artifacts that have version ranges, depending on Eclipse Platform and JDT artifacts that have version ranges. This is the case for the Xtext generator (both the new one and the old one, if I remember correctly; however, that's deprecated, isn't it). So, as soon as an Xtext bundle, which dependens on EMF or Platform/JDT, is used in a Maven dependency setting you'll face the problem again. Currently, that's the case of xtext generator, used as a Maven dependency in the exec-maven-plugin.
A lot of job has been done to pin versions in the Xtext BOM; my solution relies on those explicit versions. As long as those versions in the BOM are not touched, my solution lives on. As soon as a newer EMF Maven artifact is released that depends on newer Platform/JDT versions (which will surely be valid for the version ranges, since, as we know now, Eclipse doesn't care about semantic versioning) your solution might break.
As far as I know, xtend-maven-plugin has been fixed long ago using the same strategy (relying on the pinned versions in the Xtext BOM).

At the cost of sounding argumentative and arrogant, if you feel like my help and proposal are not worthwhile or that my experties are not enough, I'm more than happy to spend my spare time doing something else. I had the impression that Xtext needed some love and help, but the ones I provide (also in other fixes and proposals) don't seem to be taken into consideration.

Of course, I'm eager to know what @szarnekow and @kthoms have to say on my proposals, I'd really love to, but I don't understand why quick and dirty solutions like yours are applied like that, while my cleaner solutions are not even worthwhile be further discussed.

cdietrich:
@LorenzoBettini  please discuss in the issue. not in this pr
and as already stated: this is a workaround and not a solution.
as i dont know if we will have a solution in 1 week or 5 years.
i went with the workaround

szarnekow:
For the record: I'm unfortunately not an expert on maven either that's one reason why I didn't immediately jump into the discussion.

First of all: I agree that we should move the discussion to the ticket. Second of all: I agree that Lorenzo brings a great deal of maven knowledge to the table.

Now shifting to the ticket eclipse/xtext#1976

LorenzoBettini:
PING once again, since from what I understand 2.26 is going to be released soon... why isn't my proposal being considered at all?

cdietrich:
cause there is NOBODY that can assist me with assising if the thing makes sense :( :( :(

LorenzoBettini:
because I'm "nobody"...

cdietrich:
no, but i am the stupid guy that always gets all the fallout if something breaks

cdietrich:
do we have a list of stuff that would need to be pinned in the generator?

LorenzoBettini:
> do we have a list of stuff that would need to be pinned in the generator?

I'm not sure I understand

cdietrich:
do we have a complete list of all transitive ependencies yu want to put somewhere

PR #1735, opened 2021-08-12
update to gradle 7

cdietrich:
update to gradle 7
https://github.com/eclipse/xtext/issues/1548


cdietrich:
we still get

```
Execution optimizations have been disabled for task ':org.eclipse.xtext.testlanguages:compileJava' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/Users/cdietrich/xtext-master/git/xtext-core/org.eclipse.xtext.testlanguages/src-gen'. Reason: Task ':org.eclipse.xtext.testlanguages:compileJava' uses this output of task ':org.eclipse.xtext.testlanguages:generateTestLanguages' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.1.1/userguide/validation_problems.html#implicit_dependency for more details about this problem.
``` 

need a solution fo this.

oehme:
`compileJava.dependsOn(generateTestLanguages)` would be enough.

oehme (approved):

cdietrich:
@oehme that does not work. generateTestLanguages should stay an optional task

oehme:
Oh because you check in generated code, right? In that case, I'd remove the `output` declaration from `generateTestLanguages` so that Gradle won't complain. Essentially we don't want Gradle to control this output dir for us.

szarnekow (approved):

PR #1750, opened 2021-08-30
Fix port parsing in SocketServerLauncher

### LINKED ISSUE #1749: SocketServerLauncher can not recover integer argument "-port"
When launching an LSP with SocketServerLauncher with arguments "-port 8559", the server is launched on default port, not on 8559.
The problem lies in the org.eclipse.xtext.ide.server.SocketServerLauncher.getPort(String...) method, which uses `Integer.getInteger(getValue(args, PORT))`, which returns `null` in this case. `Integer.getInteger` gets value from a system property name and should be replaced by `Integer.parseInt`.

If the intention really is to use a system property (e.g. "-port MY_PORT" and the user must set MY_PORT property to the adequate value), there should be more indications in the javadoc, but I highly doubt it is intentional.

cdietrich:
Fix port parsing in SocketServerLauncher
Fixes #1749


vhemery:
LGTM

Issue #1753, opened 2021-09-14
Semantic quickfix for generic IDEs: Command interface


imhotep82:
Step backwards: What do you think about keeping the logic as is in this PR and discuss a potential change of the design in a subsequent, isolated ticket / pull request? @cdietrich @imhotep82

_Originally posted by @szarnekow in https://github.com/eclipse/xtext-core/issues/1597#issuecomment-718175071_

szarnekow:
Unfortunately, no discussion happened. Please open a ticket in https://github.com/eclipse/xtext if you want to pick this up.

PR #1754, opened 2021-09-14
Implement quickfixes with commands

imhotep82:
#1597 
#1753


imhotep82:
This is WIP, and point of discussion: One downside of this could be that the "bulb" icon will be missing for commands (afaik).

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixExecutableCommandService.java
```diff
@@ -0,0 +1,71 @@
+/*******************************************************************************
+ * Copyright (c) 2020 itemis AG (http://www.itemis.eu) and others.
```
cdietrich, line 2: 2021

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolutionInfo.java
```diff
@@ -0,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2020 TypeFox GmbH (http://www.typefox.io) and others.
```
cdietrich, line 2: 2021

cdietrich: you can also add your name/company there

cdietrich:
PR misses a change to AbstractLanguageServerTest

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixExecutableCommandService.java
```diff
@@ -0,0 +1,71 @@
...
+		return null;
+	}
+
+	private void executeQuickFix(ExecuteCommandParams params) {
+		DiagnosticResolutionInfo resolution = (DiagnosticResolutionInfo) params.getArguments().get(0);
+		Options options = resolution.getOptions();
+		DiagnosticResolution quickfix = quickfixes.resolveResolution(resolution);
+		WorkspaceEdit edit = quickfix.apply();
+		for (Either<TextDocumentEdit, ResourceOperation> change : edit.getDocumentChanges()) {
+			options.getDocument().applyChanges(change.getLeft().getEdits());
```
imhotep82, line 67: I don't know if this is the right way to do this.

imhotep82:
> PR misses a change to AbstractLanguageServerTest

What do you mean by that @cdietrich ?

cdietrich:
You pushed the change to AbstractLanguageServerTest in the meantime (build is green now)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -13,6 +13,7 @@
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.resource.ResourceSet;
 import org.eclipse.lsp4j.CodeActionParams;
+import org.eclipse.lsp4j.Diagnostic;
```
szarnekow, line 16: Is this an unused import?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolutionAcceptor.java
```diff
@@ -38,6 +38,17 @@ public void accept(String label, IModification<EObject> modification) {
 		issues.add(new DiagnosticResolution(label, modificationContextFactory, modification));
 	}
 
+	public List<DiagnosticResolutionInfo> getDiagnosticResolutionInfos() {
+		List<DiagnosticResolutionInfo> infos = new ArrayList<>();
+		int id = 0;
+		for (DiagnosticResolution issue : issues) {
```
szarnekow, line 44: Can we come up with a better name for the field `issues` and the local variable `issue`?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolutionAcceptor.java
```diff
@@ -38,6 +38,17 @@ public void accept(String label, IModification<EObject> modification) {
 		issues.add(new DiagnosticResolution(label, modificationContextFactory, modification));
 	}
 
+	public List<DiagnosticResolutionInfo> getDiagnosticResolutionInfos() {
+		List<DiagnosticResolutionInfo> infos = new ArrayList<>();
+		int id = 0;
```
szarnekow, line 43: Is it guaranteed that all resolutions will always be found in the same order?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
```diff
@@ -35,8 +35,13 @@
...
 	 *            EMF diagnostic issue
+	 * @param options 
 	 * @return 0..n resolutions for the given issue
 	 */
-	List<DiagnosticResolution> getResolutions(Options options, Diagnostic issue);
+	List<DiagnosticResolutionInfo> getResolutions(Diagnostic issue, Options options);
+
+	DiagnosticResolution resolveResolution(DiagnosticResolutionInfo info);
+
+	String getCommandProviderId();
```
szarnekow, line 45: This would need some JavaDoc

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixExecutableCommandService.java
```diff
@@ -0,0 +1,71 @@
...
+	@Override
+	public Object execute(ExecuteCommandParams params, ILanguageServerAccess access, CancelIndicator cancelIndicator) {
+		if (!cancelIndicator.isCanceled() && quickfixes.getCommandProviderId().equals(params.getCommand())) {
+			executeQuickFix(params);
+		}
+		return null;
+	}
+
+	private void executeQuickFix(ExecuteCommandParams params) {
+		DiagnosticResolutionInfo resolution = (DiagnosticResolutionInfo) params.getArguments().get(0);
```
szarnekow, line 62: I'm not sure this will be correctly typed at runtime

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixExecutableCommandService.java
```diff
@@ -0,0 +1,71 @@
...
+		Options options = resolution.getOptions();
+		DiagnosticResolution quickfix = quickfixes.resolveResolution(resolution);
+		WorkspaceEdit edit = quickfix.apply();
```
szarnekow, line 65: I think these changes would need to be sent back to the client since the server should not perform modifications that have not been requested by the client. At least that is my understanding.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
```diff
@@ -49,7 +58,20 @@ private boolean getFixMethodPredicate(Method input, String issueCode) {
 		return false;
 	}
 
-	private List<DiagnosticResolution> getResolutions(Options options, Iterable<Method> fixMethods) {
+	private DiagnosticResolution getResolution(Options options, Iterable<Method> fixMethods, int id) {
+		DiagnosticResolutionAcceptor issueResolutionAcceptor = issueResolutionAcceptorProvider.get();
+		List<Method> methods = Lists.newArrayList(fixMethods);
```
szarnekow, line 63: This is based on the assumption that the methods will be returned in the correct order always.
Given that the order in Class.getMethods is not defined, we might see flaky tests where the expectations contains the int id.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
```diff
@@ -49,7 +58,20 @@ private boolean getFixMethodPredicate(Method input, String issueCode) {
...
+		Method method = methods.get(id);
+		try {
+			// will throw if this is not a public method, but it should be
+			method.invoke(this, issueResolutionAcceptor);
+		} catch (Exception e) {
+			LOG.error("Error executing fix method", e);
+		}
+		return issueResolutionAcceptor.getDiagnosticResolutions(options).get(0);
```
szarnekow, line 71: A single fix method might contribute multiple resolutions. The hardcoded access to the first item is probably bogus.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
```diff
@@ -9,18 +9,23 @@
...
  *******************************************************************************/
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
+import javax.inject.Named;
```
szarnekow, line 19: Usually we'd use the Named annotation from Guice

szarnekow:
I like the direction of this. The most crucial thing will be the task of identifying the applied resolution after the roundtrip to the client.
Maybe we should require a custom identifier when a command is created and pass that identifier as an Optional into the getResolutions fix methods?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
heinrichWeichert: you're right, it would be better to tag the methods with ids as suggested. i'll consider this.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
heinrichWeichert: yes

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolutionAcceptor.java
heinrichWeichert: issueResolution : resourceIssueResolutions

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixExecutableCommandService.java
imhotep82: I know this is a hard cast. I guess GSON is used to deserialize this in the background? Any idea how this could be done another way?

imhotep82:
> I like the direction of this. The most crucial thing will be the task of identifying the applied resolution after the roundtrip to the client. Maybe we should require a custom identifier when a command is created and pass that identifier as an Optional into the getResolutions fix methods?

Good idea, but this is perhaps error-prone, as there is no check whether the same identifier exists more than once.
My second safest bet is the latest push, the method name. As we currently only accept methods with an annotation and a a single parameter that is typed to `DiagnosticResolutionAcceptor`, this should work.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixExecutableCommandService.java
imhotep82: Is this `org.eclipse.lsp4j.CodeActionCapabilities.dataSupport` ?

szarnekow: > I know this is a hard cast. I guess GSON is used to deserialize this in the background? 

Yes, it is. Since the list `params.getArguments()` is of type `List<Object>` the GSON deserializer doesn't have any clue about the target type.

cdietrich:
any update here?

heinrichWeichert:
> any update here?

Unfortunately not.
Should I close the PR in the meantime while it is stale?

szarnekow:
> Should I close the PR in the meantime while it is stale?

I'm closing this now. Please don't hesitate to reopen it if you find the time to pick it up again.

Issue #1522, opened 2020-06-10
Enhancements to AntlrToolFacade


cdietrich:
* Make AntlrToolFacade https aware 
* and capable to handle http redirects
* allow proxy settings

(also affects the one in xtext-extras)

cdietrich:
https://square.github.io/okhttp/

cdietrich:
move to 2.24

cdietrich:
will provide a partial fix
- use ssl by default
- fall back to http
wont care about redirects as of now

cdietrich:
created https://github.com/eclipse/xtext/issues/2464 for "the rest"

cdietrich:
Fixed in 2.26

PR #1756, opened 2021-09-16
[#1522] download via https first and try http on error only in AntlrToolFacade

cdietrich:
[#1522] download via https first and try http on error only in AntlrToolFacade

### DIFF
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AntlrToolFacade.java
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AntlrToolFacade.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2009, 2021 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -33,6 +33,7 @@ public class AntlrToolFacade {
 	}
 
 	private String downloadURL = "http://download.itemis.com/antlr-generator-3.2.0-patch.jar";
+	private String secureDownloadURL = "https://download.itemis.com/antlr-generator-3.2.0-patch.jar";
 	private boolean askBeforeDownload = false;
 
 	public void setAskBeforeDownload(boolean shouldAsk) {
@@ -42,7 +43,11 @@ public class AntlrToolFacade {
 	public void setDownloadFrom(String downloadURL) {
 		this.downloadURL = downloadURL;
 	}
-	
+
+	public void setSecureDownloadFrom(String secureDownloadURL) {
+		this.secureDownloadURL = secureDownloadURL;
+	}
+
 	private String downloadTo = "./.antlr-generator-3.2.0-patch.jar";
 
 	public void setDownloadTo(String path) {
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AntlrToolFacade.java
```diff
@@ -33,6 +33,7 @@ public void setLoader(ClassLoader loader) {
 	}
 
 	private String downloadURL = "http://download.itemis.com/antlr-generator-3.2.0-patch.jar";
+	private String secureDownloadURL = "https://download.itemis.com/antlr-generator-3.2.0-patch.jar";
```
cdietrich, line 36: tested with https://untrusted-root.badssl.com/

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AntlrToolFacade.java
```diff
@@ -122,13 +127,34 @@ public boolean download() {
...
 		} catch (IOException e) {
-			System.err.println("Downloading ANTLR parser generator failed: " + e.getMessage());
-			System.err.println("Please install the feature 'Xtext Antlr SDK' manually using the external updatesite:");
-			System.err.println();
-			System.err.println("\t\t'http://download.itemis.com/updates/'."); 
-			System.err.println();
-			System.err.println("(see http://www.eclipse.org/Xtext/download.html for details)");
-			return false;
+			try {
+				log.warn("Downloading file from '"+secureDownloadURL+"' failed.", e);
```
cdietrich, line 131: should this become an error instead?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AntlrToolFacade.java
```diff
@@ -122,13 +127,34 @@ public boolean download() {
...
+				log.info("Downloading file from '"+downloadURL+"' ...");
+				BufferedInputStream in = new BufferedInputStream(new URL(downloadURL).openStream());
+				try {
```
cdietrich, line 134: copy & paste is ugly but i dont see a nice option for reuse here

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AntlrToolFacade.java
cdietrich: should we log the exception or skip it/log the message only?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AntlrToolFacade.java
szarnekow: I think the fallback to http if https failed for some reason is not something that should be done transparently.
Can we assume that https will usually work? Should clients that cannot use https change the download location in their mwe2 file instead?

cdietrich: there is no api to store anything in mwe yet

cdietrich: i also want to keep the fallback in case people dont test early

szarnekow: Ok

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AntlrToolFacade.java
szarnekow: I think it's fine.

szarnekow (approved):

PR #1765, opened 2021-10-08
added ability to configure download url from workflow

cdietrich:
added ability to configure download url from workflow


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/AbstractAntlrGeneratorFragment2.java
```diff
@@ -61,6 +60,16 @@
 
 	private ArrayList<String> antlrParams = new ArrayList<>();
 
+	private String dowloadUrl;
+	
+	@Inject
+	private void setAntlrToolFacade(AntlrToolFacade antlrTool) {
```
szarnekow, line 66: I think the setter injection is reasonable.

szarnekow (approved):

PR #1768, opened 2021-10-21
ported xtend code 2 java

cdietrich:
ported xtend code 2 java


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGeneratorTemplates.xtend
```diff
@@ -575,4 +576,15 @@ class XtextGeneratorTemplates {
 		return file
 	}
 	
+	def TextFileAccess createMetaInfServices(List<? extends IXtextGeneratorLanguage> languageConfigs) {
```
cdietrich, line 579: wonder why this was not here in the first place

Issue #1781, opened 2021-11-27
org.eclipse.xtext.testing does not import org.junit.jupiter.api


cdietrich:
org.eclipse.xtext.testing does not import org.junit.jupiter.api
so AbstractLanguageServerTest subclasses dont run the BeforeEach/AfterEach code
https://www.eclipse.org/forums/index.php/t/1109520/

cdietrich:
repo with fix https://ci.eclipse.org/xtext/job/xtext-umbrella/job/cd_issue1781/lastSuccessfulBuild/artifact/build/p2-repository/

cdietrich:
FIxed in 2.26.0
should be available in snapshots soon

PR #1782, opened 2021-11-27
[#1781] fix package import for org.junit.jupiter.api

cdietrich:
[#1781] fix package import for org.junit.jupiter.api

### DIFF
--- a/org.eclipse.xtext.testing/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.testing/META-INF/MANIFEST.MF
@@ -28,6 +28,7 @@ Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.eclipse.xtend.lib
 Import-Package: org.apache.log4j;version="1.2.15",
  org.apache.log4j.spi;version="1.2.15",
+ org.junit.jupiter.api;version="[5.0.0,6.0.0)";resolution:=optional,
  org.junit.jupiter.api.extension;version="[5.0.0,6.0.0)";resolution:=optional
 Bundle-ActivationPolicy: lazy
 Automatic-Module-Name: org.eclipse.xtext.testing


org.eclipse.xtext.testing/META-INF/MANIFEST.MF
```diff
@@ -28,6 +28,7 @@ Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.eclipse.xtend.lib
 Import-Package: org.apache.log4j;version="1.2.15",
  org.apache.log4j.spi;version="1.2.15",
+ org.junit.jupiter.api;version="[5.0.0,6.0.0)";resolution:=optional,
```
szarnekow, line 31: Interesting find. Why did it even compile without the import?

szarnekow (approved):
Thank you for investigating this!

org.eclipse.xtext.testing/META-INF/MANIFEST.MF
cdietrich: Cause gradle does not check manifest

szarnekow: Of course... why would we compile an OSGI module with a tools that understands OSGI manifests ...

LorenzoBettini:
As @szarnekow pointed out (https://github.com/eclipse/xtext-core/pull/1782/files#r757780619) "Of course... why would we compile an OSGI module with a tools that understands OSGI manifests ..."
I'd still like to understand why we (Xtext) keep on using (read "banging our head against") Gradle instead of doing everything with the good ol' Maven, which still works like a charm and does not break things with each new release (or if it does, it's quickly fixed or it's easy to fix things).
In particular, from what I understand, none of the committers are really confident with Gradle (at least, that's the impression I had -- I seem to recall that once we completely bricked the metadata on Maven Central... something that I don't even think it's possible when releasing with Maven instead of Gradle).

PR #1788, opened 2021-12-03
Add possibiliy to use custom capacity for cache adapters

cdietrich:
Add possibiliy to use custom capacity for cache adapters


szarnekow:
LGTM.

Since you're on it: Would it makes sense to have a protected constructor that takes the map itself, such that clients can do more fancy map types if desired?

szarnekow (approved):

cdietrich:
dont think we need it as of now

PR #1790, opened 2021-12-10
Make @InjectWith meta-annotation capable (fix #1789)

### LINKED ISSUE #1789: @InjectWith is not meta-annotation capable
JUnit 5 built-in annotations are all [meta-annotation](https://junit.org/junit5/docs/current/user-guide/#writing-tests-meta-annotations) capable. That means you can create own annotations annotated with the existing ones, and use your own annotation instead (e.g. to group multiple existing annotations into a single new one).
The Xtext @InjectWith annotation doesn't support this, because its InjectionExtension [only looks for the annotation at the class under test](https://github.com/eclipse/xtext-core/blob/c6daeb75a5399903ac982a75f22a25305ac94afe/org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/extensions/InjectionExtension.java#L150) and ignores any super interfaces and similar. To fix this, the extension needs to use the JUnit Platform AnnotationSupport to query the annotation.

Will provide a PR.

Bananeweizen:
Use the JUnit Platform annotation retrieval to also look at super
interfaces, composed interfaces etc. of the test class.

JUnit Platform 1.y.z corresponds to JUnit Jupiter 5.y.z, that's why the
platform range [1,2) should fit well to the existing Jupiter range [5,6)
in the manifest.

Signed-off-by: Michael Keppler <Michael.Keppler@gmx.de>

### DIFF
--- a/org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/extensions/InjectionExtension.java
+++ b/org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/extensions/InjectionExtension.java
@@ -21,6 +21,7 @@ import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.TestInstancePostProcessor;
+import org.junit.platform.commons.support.AnnotationSupport;
 import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 import org.junit.jupiter.api.extension.ExtensionContext.Store.CloseableResource;
 
@@ -38,7 +39,7 @@ import com.google.inject.Injector;
  * It takes care about JUnit5 {@link Nested nested} test classes. They are inner classes that might
  * be annotated as well and must be handled for injection.
  * </p>
- * 
+ *
  * @author Karsten Thoms - Initial contribution and API
  * @since 2.14
  */
@@ -52,7 +53,7 @@ public class InjectionExtension implements BeforeEachCallback, AfterEachCallback
 	 * </p>
 	 */
 	private static ClassToInstanceMap<IInjectorProvider> injectorProviderClassCache = MutableClassToInstanceMap.create();
-	
+
 	/**
 	 * @since 2.24
 	 */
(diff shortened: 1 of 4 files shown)


cdietrich:
have no idea how to test this and make sure it does not bad have any terrible side effects.
e.g. we dont generate a dependency for it to target platforms

cdietrich:
simple downstream test looks good
https://github.com/itemis/xtext-reference-projects/actions/runs/1563163148
but non of it makes serious use of junit 5

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/extensions/InjectionExtension.java
```diff
@@ -140,15 +141,16 @@ protected void restoreRegistry(IInjectorProvider injectorProvider, ExtensionCont
...
 	 * If the injector provider can be found, it will be reflectively instantiated and cached. Only one instance of any
 	 * given injector provider will be created.
-	 * 
+	 *
 	 * @since 2.24
 	 */
 	protected static IInjectorProvider getOrCreateInjectorProvider(ExtensionContext context) {
-		InjectWith injectWith = context.getRequiredTestClass().getAnnotation(InjectWith.class);
-		if (injectWith != null) {
+		Optional<InjectWith> injectWithOpt = AnnotationSupport.findAnnotation(context.getRequiredTestClass(), InjectWith.class);
```
szarnekow, line 151: Could you please add a new test case to the `InjectionExtensionTest` ?

cdietrich:
@Bananeweizen  any update here?

Bananeweizen:
Sorry for the delay. I've added a new test which uses a composed annotation instead of the original InjectWith annotation. It fails on master, it fails without the annotation, but it passes with the composed annotation. I guess that should be sufficient to prove the functionality of the change.

Bananeweizen:
The failed build yesterday was probably caused by the colocation changes at the Eclipse foundation: https://twitter.com/Eclipse_Status/status/1472584444607275015. I've rebased to force a new build. Not sure if contributors can retrigger builds here with other means.

szarnekow (approved):
Not sure if you feel it's important, but the copyright headers do not mention you as an original author. Nevertheless, the change looks good.

Thank you for adding the test.

PR #1799, opened 2021-12-17
Fix hidden region formatting priorities (fix #1798)

### LINKED ISSUE #1798: Suspicious code in HiddenRegionFormatting's mergeValuesFrom
The `mergeValuesFrom` method never updates the priority of `this`. This could cause especially weird results in `HiddenRegionFormattingMerger`'s merge method, where the resulting priority could depend on the number of conflicts, and the priority of 3 or more merged conflicts will always be 0. I think fixing this would also fix the `// TODO: don't do this` in `HiddenRegionFormattingMerger`.

LawrenceGoossens:
mergeValuesFrom in HiddenRegionFormatting now updates the priority
The calculation of strategy can no longer cause overflows

Signed-off-by: Lawrence Goossens <lawrence.goossens@sigasi.com>

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/HiddenRegionFormattingMerger.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/HiddenRegionFormattingMerger.java
@@ -14,6 +14,8 @@ import org.eclipse.xtext.formatting2.AbstractFormatter2;
 import org.eclipse.xtext.formatting2.IHiddenRegionFormatting;
 import org.eclipse.xtext.formatting2.IMerger;
 
+import com.google.common.collect.Lists;
+
 /**
  * @author Moritz Eysholdt - Initial contribution and API
  */
@@ -28,13 +30,16 @@ public class HiddenRegionFormattingMerger implements IMerger<IHiddenRegionFormat
 
 	@Override
 	public IHiddenRegionFormatting merge(List<? extends IHiddenRegionFormatting> conflicting) {
+		// If there are only 2 conflicts,
+		// usages of this method expect the second value to be updated to the merge result
+		// TODO: Fix those usages so they no longer expect this method to edit its input
 		if (conflicting.size() == 2) {
-			// TODO: don't do this
 			conflicting.get(1).mergeValuesFrom(conflicting.get(0));
 			return conflicting.get(1);
 		}
 		IHiddenRegionFormatting result = formatter.createHiddenRegionFormatting();
-		for (IHiddenRegionFormatting conflict : conflicting)
+		// Reversed so the merging order is consistent with the special case above
+		for (IHiddenRegionFormatting conflict : Lists.reverse(conflicting))
 			result.mergeValuesFrom(conflict);
 		return result;
 	}
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/HiddenRegionFormatting.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/HiddenRegionFormatting.java
@@ -107,7 +107,7 @@ public class HiddenRegionFormatting implements IHiddenRegionFormatting {
 
 	@Override
 	public void mergeValuesFrom(IHiddenRegionFormatting other) throws ConflictingFormattingException {
-		int strategy = other.getPriority() - getPriority();
+		int strategy = Integer.compare(other.getPriority(), getPriority());
 		setSpace(merge(getSpace(), other.getSpace(), strategy, "space"));
 		setNewLinesMin(merge(getNewLineMin(), other.getNewLineMin(), strategy, "newLineMin"));
 		setNewLinesDefault(merge(getNewLineDefault(), other.getNewLineDefault(), strategy, "newLineDefault"));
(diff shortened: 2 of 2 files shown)


cdietrich:
wonder how to test this. have you a usecase?

cdietrich:
breaks tests in extras 
https://ci.eclipse.org/xtext/job/xtext-extras/job/LawrenceGoossens-fix-hidden-region-formatting-priorities/lastCompletedBuild/testReport/

cdietrich:
also in xtend
https://ci.eclipse.org/xtext/job/xtext-xtend/job/LawrenceGoossens-fix-hidden-region-formatting-priorities/1/testReport/junit/org.eclipse.xtend.core.tests.formatting/XtendClassFormatterTest/formatMethod3/

LawrenceGoossens:
All tests succeeded locally, there must be something wrong with my setup. I'll put his pr in draft for now untill I fix it.

cdietrich:
no code is split over x repos. you just changed xtext-core.
which projects did you setup in your eclipse?
did you use oomph an import all projects from all repos?

cdietrich:
in xtext-umbrella there is also a grade composite build at 
https://github.com/eclipse/xtext-umbrella/tree/master/releng/gradle-composite
this one has a generatePrefs task
that processes 
the location.properties in xtext-umbrella
and setups buildship in other projects to use upstream stuff
```
cat locations.properties
git.clone.xtext.lib.location=/Users/cdietrich/xtext-master/git/xtext-lib
git.clone.xtext.core.location=/Users/cdietrich/xtext-master/git/xtext-core
git.clone.xtext.extras.location=/Users/cdietrich/xtext-master/git/xtext-extras
git.clone.xtext.xtend.location=/Users/cdietrich/xtext-master/git/xtext-xtend
git.clone.xtext.web.location=/Users/cdietrich/xtext-master/git/xtext-web%
```
for gradle you might need to update 
https://github.com/eclipse/xtext-extras/blob/master/gradle/upstream-repositories.gradle
to pick the branch you created in core
https://ci.eclipse.org/xtext/job/xtext-core/view/change-requests/job/PR-1799/lastSuccessfulBuild/artifact/build/maven-repository/

unfortunately somebody thought repo split was a good idea which leads to this mess, specially for non committer contributions

LawrenceGoossens:
Locally I still have some test that give errors with or without changes, but all previously failing tests now succeed.

LawrenceGoossens:
It turned out that the code with the todo in `HiddenRegionFormattingMerger` is still needed after the change.

LawrenceGoossens:
As for testing, I got here based on a stacktrace that I think was caused by this, but I haven't been able to reproduce it.

cdietrich:
downstream tests are green again

szarnekow (approved):
The change looks good to me. Mind to remove the confusing comment or even updating it to reflect why the special casing for 2 elements is required?

LorenzoBettini:
> unfortunately somebody thought repo split was a good idea which leads to this mess, specially for non committer contributions

I take this opportunity to once again raise the question: why don't we go back to a monolithic repo? I guess this would be better not only for non-committers but even more for we committers ourselves. I seem to understand that the promoters of split repos are not actively working anymore on this project.

szarnekow:
@LorenzoBettini Thank you for chiming in. I think all active committers are on the same page in that regard. Unfortunately a repo merge is not a trivial exercise. Besides somehow merging the histories of multiple git repos, we have to fix the build. A while back, some effort was done by Arne, but it never landed. I'm afraid the amount of necessary work is easily underestimated and so far no sponsor was stepping up to fund a repo-merge.

LorenzoBettini:
Hi @szarnekow 
Probably it's not that easy, but not even that hard.

If you have a git repo A and an utterly unrelated git repo B, you can add B as a remote of A (I did that a few times by mistake O:) and you end up with two separate streams in the same repo. Then you merge the second stream into the first one. Of course, there'll be a few conflicts to merge for the duplicate files at the root, but all the subdirectories with different names will be seamlessly merged. The two histories will be kept, and from the merge point on, you get a single stream.

For example,

```
git clone git@github.com:eclipse/xtext-core.git
cd xtext-core
git remote add xtext-lib git@github.com:eclipse/xtext-lib.git
git fetch xtext-lib
gitk --all &
```

And you get the two separate streams ready to be merged (I know, there will be conflicts for root files).

image

And the two repos' histories will survive.

Merging all the builds into a single one shouldn't be that hard.

LawrenceGoossens:
> The change looks good to me. Mind to remove the confusing comment or even updating it to reflect why the special casing for 2 elements is required?

I updated the comment. I also reversed the direction of the loop so the merge direction is consistent, in case `mergeValuesFrom` ever gets overridden with something asymmetric.

org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/HiddenRegionFormattingMerger.java
```diff
@@ -28,13 +30,16 @@ public HiddenRegionFormattingMerger(AbstractFormatter2 formatter) {
...
+		// usages of this method expect the second value to be updated to the merge result
+		// TODO: Fix those usages so they no longer expect this method to edit its input
 		if (conflicting.size() == 2) {
-			// TODO: don't do this
 			conflicting.get(1).mergeValuesFrom(conflicting.get(0));
 			return conflicting.get(1);
 		}
 		IHiddenRegionFormatting result = formatter.createHiddenRegionFormatting();
-		for (IHiddenRegionFormatting conflict : conflicting)
+		// Reversed so the merging order is consistent with the above if
```
szarnekow, line 41: There seems to be something missing from the comment.

szarnekow: Ah got it now, `with the above if condition` or `with the special case above` might be a better wording.

LawrenceGoossens: I fixed it

szarnekow (approved):
Thank you!

PR #1801, opened 2022-01-10
Make QualifiedName.USE_INTERNING writable (using reflection)

### LINKED ISSUE #1787: xtext.qn.interning should be evaluated later
Xtext provides the system property `xtext.qn.interning` for interning the Strings used in `org.eclipse.xtext.naming.QualifiedName`.

Unfortunately this property is evaluated when the class is loaded as it's a static final constant.

The only way how to set this property is using a -D JVM parameter. I would like to use System.setProperty or I would like to be able to configure it by any other way like binding some other implementation.

We ship our application with an Xtext interpreter and we have some communication efforts and error sources, if the application's operators have to pass the JVM property externally.

cdietrich:
Make QualifiedName.USE_INTERNING writable (using reflection)
Fixes #1787


szarnekow (approved):
I think it's a fair compromise

PR #1803, opened 2022-01-10
Introduction of QualifiedNameInterning in order to configure QualifiedName.USE_INTERNING without using reflection

OLibutzki:
Signed-off-by: Oliver Libutzki <oliver@libutzki.de>

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
@@ -36,7 +36,7 @@ public class QualifiedName implements Comparable<QualifiedName> {
 
 	private QualifiedName lowerCase;
 
-	private static boolean USE_INTERNING = Boolean.getBoolean("xtext.qn.interning");
+	static boolean USE_INTERNING = Boolean.getBoolean("xtext.qn.interning");
 
 	/**
 	 * The single existing empty QualifiedName.
@@ -128,18 +128,15 @@ public class QualifiedName implements Comparable<QualifiedName> {
 	}
 
 	/**
-	 * Returns string internal instance from string pool, if a system property {@code xtext.qn.interning} is set to {@code true}, or the
-	 * same object otherwise.
+	 * Returns string internal instance from string pool, if {@link #USE_INTERNING} is true, or the same object
+	 * otherwise.
 	 * <p>
-	 * Implementation notes:
-	 * <ol>
-	 * <li>Interning {@link String} objects may affect performance, see bug 484215.
-	 * <li>Interning {@link String} objects is not recommended for older JVM's, because of possible perm gen memory explosion, see
-	 * http://java-performance.info/string-intern-in-java-6-7-8/.
-	 * </ol>
+	 * {@link #USE_INTERNING} can be set by using the system property {@code xtext.qn.interning} or by calling
+	 * {@link QualifiedNameInterning#enable()} / {@link QualifiedNameInterning#disable()}.<br/>
+	 * The default value is <code>false</code>.
 	 */
 	private static String intern(String string) {
-		return USE_INTERNING? CommonUtil.intern(string) : string;
+		return USE_INTERNING ? CommonUtil.intern(string) : string;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
```diff
@@ -36,7 +36,7 @@
 
 	private QualifiedName lowerCase;
 
-	private static boolean USE_INTERNING = Boolean.getBoolean("xtext.qn.interning");
+	public static boolean USE_INTERNING = Boolean.getBoolean("xtext.qn.interning");
```
szarnekow, line 39: I'm not a big fan of the undocumented public boolean flag here, but since the command line switch wasn't documented either, it's probably ok.

In other cases, where we added internal API via statics, we tried to hide it a little more, e.g. right now every code completion with the goal to show `QualifiedName.create` after the dot would also show `USE_INTERNING`.

Can we make this a protected field, such that the rare use case can be achieved by:
```
class QualifiedNameAccess extends QualifiedName {
  static void useInterning() {
    USE_INTERNING = true;
  }
}
```

OLibutzki: What about package private visibility without using inheritance?

szarnekow: I had in mind that `QualifiedNameAccess` would not be in the Xtext codebase but part of the user code.

szarnekow: I'd also be ok with a class next to QualifiedName, e.g. `QualifiedNameInterning` with public methods `enable()` and `disable()` and `USE_INTERNING` being package visible.

OLibutzki: Got it. I'm going to adjust the PR.

szarnekow (commented):
Thank you also for adding the comments. Could you please rephrase the comments such that they match the implementation? Bonus points for updating the comments on the original QualifiedName.intern method, too.

cdietrich:
can you please sqash the commits into one

org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedNameInterning.java
```diff
@@ -0,0 +1,51 @@
+/*******************************************************************************
+ * Copyright (c) 2022 itemis AG (http://www.itemis.eu) and others.
```
szarnekow, line 2: The copyright header is probably not the correct one.

org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedNameInterning.java
```diff
@@ -0,0 +1,51 @@
...
+ * <p>
+ * String interning is disabled by default. It can be enabled by setting the system property {@code xtext.qn.interning} to <code>true</code> or by calling {@link #enable()}.
+ * <p>
+ * Caution: Setting the system property at runtime does not have an effect. Use {@link #enable()} to enable String interning at runtime.
+ * 
+ *  @see QualifiedName#USE_INTERNING
+ * 
+ * @author Oliver Libutzki - Initial contribution and API
+ */
+public final class QualifiedNameInterning {
```
szarnekow, line 22: An `@since` tag would be nice.

szarnekow (approved):
@OLibutzki Thank you for the patch. I've two minor comments that I'd be also ok with in the current shape.

org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedNameInterning.java
OLibutzki: Don't know... I just left it as-is.

org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedNameInterning.java
OLibutzki: I added it.

OLibutzki:
I did not format the code using the jd formatter settings. I applied that formatter in my latest squashed commit, too.

szarnekow (approved):

cdietrich (approved):

szarnekow:
It's merged and will be available with 2.26. Thank you @OLibutzki for driving this home.

OLibutzki:
You're welcome. @szarnekow, thanks for reviewing quickly so it was it short way home.

PR #1804, opened 2022-01-12
[eclipse/xtext#1976] pin explicit transitive deps in xtext.xtext.generator

cdietrich:
[eclipse/xtext#1976] pin explicit transitive deps in xtext.xtext.generator


LorenzoBettini (approved):
If I understand correctly, that's basically my original proposal apart from using api instead of compile (I don't know the difference in gradle). Besides that, I'm in favor of approval!

cdietrich:
@LorenzoBettini  yes this is just a version of your patch for gradle 7

LorenzoBettini:
@cdietrich OK, thanks for handling this.
I guess then we'll also have to change what the wizard generates in the POM, right?

cdietrich:
@LorenzoBettini  what do you want to change?

cdietrich:
yes unfortunately testing it is pain. so i would do it in 2nd pr

cdietrich:
https://github.com/itemis/xtext-reference-projects/pull/227

szarnekow (approved):

PR #1805, opened 2022-01-12
[eclipse/xtext#1976] remove java 8 workaround

cdietrich:
[eclipse/xtext#1976] remove java 8 workaround


LorenzoBettini:
@cdietrich, I bet you also have to update the test output expectations.

cdietrich:
yes and no.i ran the update main method and had no changes. need to check why

LorenzoBettini:
> yes and no.i ran the update main method and had no changes. need to check why

But now I see you also updated the expectations...

PR #1807, opened 2022-01-17
Downgrade error to debug log

rubenporras:


org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
```diff
@@ -156,15 +156,19 @@ public ResolvedResourceDescription(IResourceDescription original) {
 
 		@Override
 		public Iterable<QualifiedName> getImportedNames() {
-			IllegalStateException exception = new IllegalStateException("getImportedNames" + getURI());
-			LOG.error(exception, exception);
+			if (LOG.isDebugEnabled()) {
```
szarnekow, line 159: @rubenporras I appreciate the effort to align this logging with the flow, but in fact this is a highly problematic situation. If these descriptions end up somewhere unexpected, at least a warning is appropriate.

cdietrich:
can you please rebase and squash the commits into one.
can you also give some hints on the reasoning behind the changes

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
rubenporras: @szarnekow, thanks. If it is that problematic, then I will need to take a closer look why with my customization of the Language Server vor de DDK I end up processing these.
May I ask in the meantime why is that problematic?

szarnekow: Either you have code paths that attempt to use reference descriptions or imported names in an unexpected phase, or the flow below https://github.com/eclipse/xtext-core/blob/4e87c8eca9bd88088f0207cf9ed2a4661737228c/org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java#L245 is buggy.

It's noteworthy that these descriptions have been created for resources that have not been fully processed. If they remain in the index after the build, subsequent changes cannot correctly detect the affected resources and the index state will gradually erode and become different from the actual project contents on disk.
The ResolvedDescriptions should be local to a single build run and never survive that single build cycle. Within the build loop, there imported names and reference descriptions should never be necessary.

rubenporras: Thanks for the explanation @szarnekow. I have closed the PR by mistake while trying to rebase and squash my commit, but I will leave it closed until I have done the analysis on my side.

PR #1814, opened 2022-01-26
Add protected method for subclassing Indexer.ResolvedResourceDescription

rubenporras:
As a followup of https://github.com/eclipse/xtext-core/pull/1807, do not generally downgrade the error to debug log in the Indexer, but allow subclassing of Indexer.ResolvedResourceDescription to remove the error logs if desired (or extend it in any other way) without the need to reimplement the full addToIndex method.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
@@ -245,6 +245,18 @@ public class Indexer {
 	}
 
 	/**
+	 * Return a new resolved resource description from a {@link IResourceDescription}.
+	 *
+	 * @param description
+	 *            the resource description, must not be {@code null}
+	 * @return the new resolved resource description, never {@code null}
+	 * @since 2.26
+	 */
+	protected IResourceDescription getResolvedResourceDescription(IResourceDescription description) {
+		return new Indexer.ResolvedResourceDescription(description);
+	}
+
+	/**
 	 * Index the given resource.
 	 *
 	 * @param isPreIndexing
@@ -257,10 +269,8 @@ public class Indexer {
 		IResourceServiceProvider serviceProvider = context.getResourceServiceProvider(uri);
 		IResourceDescription.Manager manager = serviceProvider.getResourceDescriptionManager();
 		IResourceDescription newDescription = manager.getResourceDescription(resource);
-		IResourceDescription toBeAdded = new Indexer.ResolvedResourceDescription(newDescription);
-		IResourceDescription.Delta delta = manager
-				.createDelta(oldIndex != null ? oldIndex.getResourceDescription(uri) : null, toBeAdded);
-		return delta;
+		IResourceDescription toBeAdded = getResolvedResourceDescription(newDescription);
+		return manager.createDelta(oldIndex != null ? oldIndex.getResourceDescription(uri) : null, toBeAdded);
 	}
 
 	/**


org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
```diff
@@ -244,6 +244,17 @@ public URI getURI() {
...
 	}
 
+	/**
+	 * Return a new {@link ResolvedResourceDescription} from a {@link IResourceDescription}.
+	 *
+	 * @param description
+	 *            the resource description, must not be {@code null}
+	 * @return the new resolved resource description, never {@code null}
+	 */
+	protected Indexer.ResolvedResourceDescription getResolvedResourceDescription(IResourceDescription description) {
```
cdietrich, line 254: misses `@since 2.26`

cdietrich:
@rubenporras i still wonder what your usecase is and WHY it fails

cdietrich:
can you please squash the commits

rubenporras:
Hi @cdietrich , 

we will need a subclass like the one in https://github.com/dsldevkit/dsl-devkit/blob/2722e8120d35e48cde28ecde1fd86ea59ec92631/com.avaloq.tools.ddk.xtext.builder/src/com/avaloq/tools/ddk/xtext/builder/FixedCopiedResourceDescription.java (that one has a slightly different behaviour than CopiedResourceDescription and implements an interface IResourceDescription2 that we use in the dsl-devkit).

For the ResolvedResourceDescription we would like as well to implement IResourceDescription2. I could not yet understand why we need to call getImportedNames()  on the CopiedResourceDescription and on the ResolvedResourceDescription, during our builds, I am still on it, but adding the interface makes sense for us, and this customization point is quite easy to do.

I will comment when I found out why we call getImportedNames in our builds.

cdietrich:
i still have the feeling you do somthing wrong and this pr would make it easier to do so
lets see what @szarnekow thinks

rubenporras:
Hi @cdietrich ,

The reason we access the information of this resource description is that we have our custom index which is persisted in a database. All the deltas registered in the index are persisted on the database, even if a description does not have any imported names or referenced descriptions, we still run the persist on it. You might be right that we should not do persist this descriptions, but for that we would need to filter them out in the code of the custom database index (maybe with instanceof or similar) and that would not be very nice. I can image that is the reason the CopiedResourceDescriptor was customized to suppress the errors. This customization is very old, and I was not involved in it, so I can only guess, the only thing I can tell for sure is that in all these years nothing bad come out of it.

Independent on that, for performance reasons we have an extra interface IResourceDescription2, which has not only the imported names but also the corresponding EClass. That allows us to have more efficient invalidation.

Regards

cdietrich:
without looking at the code, i would expect that after the build we have full index entries

rubenporras:
I think that is right, but in between the the copied description has been added to the index (even though it is replaced later), and our database-backed implementation index process it in between, even though if it is going to be replaced shortly after, the index does not know it. 

Our index implementation, besides keeping elements in memory, flushes them to the database, and this flush access the data. The fact that the data in the database might contain a state which should not outlive the build is not a problem, because as you say the descriptor will be replaced, and we commit the index in oun transaction, so the copied index is not visible from outside.

Does it make sense now?

cdietrich:
but index will be incomplete at this point in time wont it?

rubenporras:
Yes, you understand correctly, the index is incomplete at this point incomplete. 

We replicate in a background thread the memory index to a database, like in a logger pattern, so if the index is in intermediate state, that is replicated as well. A logger cannot help that.

It does not matter to our use case that we replicate an intermediate/incomplete state because the whole build correspond to a transaction on the database, so at the end of the build we wait for the background thread to flush the entries and  commited them or we just rollback whatever was written to the database. By then this copied descriptor has been replaced, as you said.

Since the logger cannot help to replicate all descriptions, it needs to call methods on  them, and if they log errors, that is something we need to fix.

Does it makes sense?

cdietrich:
@szarnekow what do you think?

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
szarnekow: Is it possible to make the API slightly more generic, e.g. return IResourceDescription instead of an implementation class. Something like `protected IResourceDescription toIndexingPhaseDescription(IResourceDescription description) {}` ?

szarnekow:
> @szarnekow what do you think?

I think the protected method is a fair compromise.

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
rubenporras: done

szarnekow (approved):

szarnekow:
I'm ok with the change. What do you think about it, @cdietrich ?

rubenporras:
@szarnekow , @cdietrich , thanks a lot for approving the changes. Could you merge it? I do not have the right to merge it.

cdietrich (approved):

cdietrich:
thx @szarnekow @rubenporras

PR #1815, opened 2022-01-27
Expose serialization methods on AbstractNode

msujew:
Signed-off-by: Mark Sujew <mark.sujew@typefox.io>

In one of our projects, we override the `NodeModelBuilder` to create custom leaf nodes (which don't inherit from the default `LeafNode` class). However, since the `NodeType` enum, and the `write`/`readData` methods were package private, we were not able to override them, even though the serializer needs them to be overriden.

This change just changes them to be accessible by downstream classes which directly inherit `AbstractNode`.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/AbstractNode.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/AbstractNode.java
@@ -360,13 +360,13 @@ public abstract class AbstractNode implements INode, BidiTreeIterable<INode> {
 		return prev != this;
 	}
 
-	enum NodeType {
+	protected enum NodeType {
 		CompositeNode, LeafNode, CompositeNodeWithSemanticElement, CompositeNodeWithSyntaxError, CompositeNodeWithSemanticElementAndSyntaxError, RootNode, HiddenLeafNode, HiddenLeafNodeWithSyntaxError, LeafNodeWithSyntaxError
 	}
 
-	abstract NodeType getNodeId();
+	protected abstract NodeType getNodeId();
 	
-	void readData(DataInputStream in, DeserializationConversionContext context) throws IOException {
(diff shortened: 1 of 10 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/RootNode.java
```diff
@@ -153,7 +153,7 @@ protected void basicSetParent(CompositeNode parent) {
 	}
 	
 	@Override
-	void readData(DataInputStream in, DeserializationConversionContext context) throws IOException {
+	public void readData(DataInputStream in, DeserializationConversionContext context) throws IOException {
```
szarnekow, line 156: I'd prefer to make this protected rather than public.

Other protected implementation details are made accessible via the InternalNodeModelUtils. You probably want to make readData accessible there such that clients can extend InternalNodeModelUtils to call readData directly.

I'm asking for this since it's strictly an implementation detail how the AbstractNode hierarchy is designed. The indirection via InternalNodeModelUtils gives us at least a little wiggle room for optimizations / refactorings - which didn't happen in the recent past, but there exist ideas on how to reduce the memory footprint of the node model.

The more methods are directly available on that type hierarchy, the harder it gets to make changes.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/AbstractNode.java
```diff
@@ -360,13 +360,13 @@ protected boolean basicHasSiblings() {
 		return prev != this;
 	}
 
-	enum NodeType {
+	public enum NodeType {
```
szarnekow, line 363: It should be possible to keep this enum protected.

msujew:
@szarnekow Thanks for the feedback, both of cases should've been `protected` anyway 👍

PR #1819, opened 2022-02-01
Always write binary models

rubenporras:
Write binary models even if no GeneratorDelegate is provided. The binary models can be used when demand-loading resources during the incremental build, regardless of if there is a GeneratorDelegate, thus, persist them at the beginnig of the generate method.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
@@ -311,10 +311,6 @@ public class IncrementalBuilder {
 		 */
 		protected void generate(Resource resource, BuildRequest request, Source2GeneratedMapping newMappings) {
 			IResourceServiceProvider serviceProvider = getResourceServiceProvider(resource);
-			GeneratorDelegate generator = serviceProvider.get(GeneratorDelegate.class);
-			if (generator == null) {
-				return;
-			}
 			Set<URI> previous = newMappings.deleteSource(resource.getURI());
 			URIBasedFileSystemAccess fileSystemAccess = createFileSystemAccess(serviceProvider, resource);
 			fileSystemAccess.setBeforeWrite((uri, outputCfgName, contents) -> {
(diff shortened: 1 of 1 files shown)


cdietrich:
am not sure if this is a good idea.

rubenporras:
Sorry I ask something that I should know already, but is the usage of binary models to speed up subsequent incremental builds not a normal use case? That is how we use them, and so far I have assumed that is the normal use case. I also assumed that if IResourceStorageFacade is provided, one should just persist them, but that is just how we do it.

In any case, yes, I can make it an opt-out flag if you like, no problem. Should I?

PD: I will squash the change later if we think it is a good idea to merge it in this or another version.

org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
```diff
@@ -311,12 +311,20 @@ protected boolean validate(Resource resource) {
...
+			fileSystemAccess.setContext(resource);
+			if (request.isWriteStorageResources() && resource instanceof StorageAwareResource) {
+				IResourceStorageFacade resourceStorageFacade = ((StorageAwareResource) resource)
+						.getResourceStorageFacade();
+				if (resourceStorageFacade != null) {
+					resourceStorageFacade.saveResource((StorageAwareResource) resource, fileSystemAccess);
+				}
+			}
 			GeneratorDelegate generator = serviceProvider.get(GeneratorDelegate.class);
 			if (generator == null) {
```
cdietrich, line 324: wonder how this can happen

cdietrich:
how can you end up with generatordelegate null?

org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
cdietrich: looks like https://github.com/eclipse/xtext/commit/d1d683218487a85a968ec472a13fe40dc65f5af1
before it was IGenerator2, now its GeneratorDelegate, first can be null, but latter?

cdietrich: => simply remove null check?

szarnekow (approved):

cdietrich (approved):

cdietrich:
thx @rubenporras

PR #1822, opened 2022-02-04
XbaseUsageDetector for overriden rules

rubenporras:
Fix XbaseUsageDetector for overriden rules, where the gramar is a different one than org.eclipse.xtext.xbase.Xtype. Instead of the EPackage URI of the rule.

### DIFF
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/xbase/XbaseUsageDetector.java
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/xbase/XbaseUsageDetector.java
@@ -11,9 +11,11 @@ package org.eclipse.xtext.xtext.generator.xbase;
 import java.util.HashSet;
 import java.util.Set;
 
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.ecore.EClass;
 import org.eclipse.xtext.AbstractRule;
+import org.eclipse.xtext.EcoreUtil2;
 import org.eclipse.xtext.Grammar;
-import org.eclipse.xtext.GrammarUtil;
 import org.eclipse.xtext.xtext.UsedRulesFinder;
 import org.eclipse.xtext.xtext.generator.util.GrammarUtil2;
 
@@ -35,7 +37,7 @@ public class XbaseUsageDetector {
 		new UsedRulesFinder(usedRules).compute(grammar);
 		for (AbstractRule it : usedRules) {
 			if ("XImportSection".equals(it.getName())
-					&& "org.eclipse.xtext.xbase.Xtype".equals(GrammarUtil.getGrammar(it).getName())) {
+					&& EcoreUtil2.isAssignableFrom((EClass) it.eResource().getResourceSet().getEObject(URI.createURI("http://www.eclipse.org/xtext/xbase/Xtype#//XImportSection"), true), (EClass) it.getType().getClassifier())) {
 				return true;
 			}
 		}


rubenporras:
A very similar fix has been present in the DDK since very long (https://github.com/dsldevkit/dsl-devkit/blob/351f81c0a5a7d78f0ae0bfe78ab19bc42cc46c62/com.avaloq.tools.ddk.xtext.generator/src/com/avaloq/tools/ddk/xtext/generator/xbase/FixedXbaseUsageDetector.xtend) for grammars like https://github.com/dsldevkit/dsl-devkit/blob/351f81c0a5a7d78f0ae0bfe78ab19bc42cc46c62/com.avaloq.tools.ddk.check.core/metamodel/com/avaloq/tools/ddk/check/Check.xtext, which override the rule XImportSection.

I have just improved it to have an URI that will match only http://www.eclipse.org/xtext/xbase/Xtype and not any grammar with Xtype in the name.

cdietrich:
will try to test with generating test langs in xtext-extras and xtext-eclipse

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/xbase/XbaseUsageDetector.java
```diff
@@ -35,7 +35,7 @@ public boolean usesXImportSection(Grammar grammar) {
 		new UsedRulesFinder(usedRules).compute(grammar);
 		for (AbstractRule it : usedRules) {
 			if ("XImportSection".equals(it.getName())
-					&& "org.eclipse.xtext.xbase.Xtype".equals(GrammarUtil.getGrammar(it).getName())) {
+					&& "http://www.eclipse.org/xtext/xbase/Xtype".equals(it.getType().getMetamodel().getEPackage().getNsURI()) {
```
cdietrich, line 38: am not sure if this is null safe enhough

cdietrich: can you also fix the compile error?

rubenporras: Fixed the compile error.
I am also not sure if the chain is safe enough. I think getType can return null, getMetamodel and getEPackage will not return null, but can return a proxy. Shall I modify the code to check for proxies and null?

cdietrich: can you test with something broken. e.g

XImportSection returns does::NotExist

rubenporras: I am sorry, but I do not know how to do that, If I run the workflow which such a grammar, I get

XtextLinkingDiagnostic: null:23 Couldn't resolve reference to EClassifier 'NotExist'.
TransformationDiagnostic: null:23 Cannot create type 'NotExist' in alias null (ErrorCode: CannotCreateTypeInSealedMetamodel)
	at org.eclipse.xtext.xtext.generator.XtextGeneratorLanguage.initialize(XtextGeneratorLanguage.java:223)
	at org.eclipse.xtext.xtext.generator.XtextGenerator.initialize(XtextGenerator.java:175)
	at org.eclipse.xtext.xtext.generator.XtextGenerator.checkConfigurationInternal(XtextGenerator.java:129)
	at org.eclipse.emf.mwe.core.lib.AbstractWorkflowComponent2.checkConfiguration(AbstractWorkflowComponent2.java:23)
	at org.eclipse.emf.mwe.core.lib.Mwe2Bridge.preInvoke(Mwe2Bridge.java:69)
	at org.eclipse.emf.mwe.core.lib.AbstractWorkflowComponent.preInvoke(AbstractWorkflowComponent.java:208)
	at org.eclipse.emf.mwe2.runtime.workflow.AbstractCompositeWorkflowComponent.preInvoke(AbstractCompositeWorkflowComponent.java:31)
	at org.eclipse.emf.mwe2.runtime.workflow.Workflow.run(Workflow.java:20)
	at org.eclipse.emf.mwe2.launch.runtime.Mwe2Runner.run(Mwe2Runner.java:103)
	... 4 more

so I do not get to the point where the is run and XbaseGeneratorFragment2 is run as well.

cdietrich: so it looks like it practially cant get null
@szarnekow what do you think

rubenporras: Yes, I think in practice it cannot. 

I could also contribute our test for the fix, which would be in 
https://github.com/dsldevkit/dsl-devkit/blob/351f81c0a5a7d78f0ae0bfe78ab19bc42cc46c62/com.avaloq.tools.ddk.xtext.generator.test/src/com/avaloq/tools/ddk/xtext/generator/xbase/test/FixedXbaseGeneratorFragmentTest.xtend

but I am unsure how your test setup works. Could I put a similar class in org.eclipse.xtext.tests/generator/src/org/eclipse/xtext)/tests and it will be executed by your lunch config [xtext.tests (xtend).launch](https://github.com/eclipse/xtext-core/blob/master/org.eclipse.xtext.tests/xtext.tests%20(xtend).launch)?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/xbase/XbaseUsageDetector.java
```diff
@@ -35,7 +35,7 @@ public boolean usesXImportSection(Grammar grammar) {
 		new UsedRulesFinder(usedRules).compute(grammar);
 		for (AbstractRule it : usedRules) {
 			if ("XImportSection".equals(it.getName())
-					&& "org.eclipse.xtext.xbase.Xtype".equals(GrammarUtil.getGrammar(it).getName())) {
+					&& "http://www.eclipse.org/xtext/xbase/Xtype".equals(it.getType().getMetamodel().getEPackage().getNsURI())) {
```
szarnekow, line 38: If any of these would be null, it would have been reported before.

It is possible to override rules and return a specialization of XImportSection.
You might want to try something like `EcoreUtil2.isAssignableFrom(XtypePackage.Literals.XIMPORT_SECTION, (EClass)it.getType().getClassifier())`

rubenporras: HI @szarnekow ,
no, that version does not work, the it.getType().getClassifier() returns an empty list for getEAllSupperTypes, so EcoreUtil2.isAssignableFrom returns false.

szarnekow: Can you confirm that the Classifier is an instanceof EClass and is resolved, e.g. not a proxy?

szarnekow: Ah, this might suffer from a duality of EPackages:
XtypePackage.Literals.XIMPORT_SECTION must likely be replaced by something like `it.eResource().getResourceSet().getEObject(EcoreUtil.getURI(XtypePackage.Literals.XIMPORT_SECTION))`.

rubenporras: @szarnekow ,
Thanks, that version works.
What is the duality of EPackages, by the way?

rubenporras: I do not understand why the Xtype class cannot be found, do I need to use ;x-installation:=greedy in the manifest together with org.eclipse.xtext.xbase?

rubenporras: I have read about the builds in https://github.com/eclipse/xtext/blob/master/Builds.md, but it is still not clear to me why the type cannot be resolved. Do you know what I need to adapt in addition to the MANIFEST.MF?

cdietrich: we have a splitted repo unfortunetly. xtext-core does not know anything about downstream xbase/common.types which are in xtext-extras. we also build with gradle which does not know anything about manifest.mf
so i fear you cannot access any xbase/common.types java types by class literal

cdietrich: i guess also @szarnekow has forgotten that

szarnekow: True. I wrongfully assumed that the generator has a dependency on Xbase, but that’s only fake by means of the XbaseGeneratorFragment.

rubenporras: OK, then I go back to the version based on the URI. That is the best we can do, right?

szarnekow: If we use a hardcoded URI, I'd be in favor of hardcoding the URI for the XImportSection EClass and attempt to find that one in the resource set + do the isAssignableFrom.

rubenporras: sure, no problem. I have sent that version for review.

szarnekow (approved):
Thank you @rubenporras

rubenporras:
@cdietrich , could you merge the change? Thanks

cdietrich:
i first want to test this locally with downstream projects

szarnekow:
> i first want to test this locally with downstream projects

Thank you for testing this!

PR #1823, opened 2022-02-07
Replace and transfer look ahead with offset

rubenporras:
NodeModelBuilder when adjusting offsets performs it for the whole node
model starting from the root. This is necessary for partial reparse as
the offsets of the following nodes could change.

For parser delegation, that a node corresponding to a terminal rule is
is replaced by new root node (where the node corresponding to the
terminal has been parsed by a different language inside a value
converter), doing the full adjustments leads to a big overhead with
large number of delegations, since the initial offset is know, the
replacement is done directly in the node which needs reparsing and does
not start from the root.

For this use case, we create a new method which keeps the offset of the
old node and iterates over just the tree under from the new node.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
@@ -273,16 +273,17 @@ public class NodeModelBuilder {
 		}
 	}
 	
-	public void replaceAndTransferLookAhead(INode oldNode, INode newRootNode) {
+	private AbstractNode doReplaceAndTransferLookAhead(INode oldNode, INode newRootNode) {
 		AbstractNode newNode = ((CompositeNode) newRootNode).basicGetFirstChild();
 		replaceWithoutChildren((AbstractNode) oldNode, newNode);
 		if (oldNode instanceof ICompositeNode && newNode instanceof CompositeNode) {
 			CompositeNode newCompositeNode = (CompositeNode) newNode;
 			newCompositeNode.basicSetLookAhead(((ICompositeNode) oldNode).getLookAhead());
 		}
-		ICompositeNode root = newNode.getRootNode();
-		BidiTreeIterator<AbstractNode> iterator = ((AbstractNode) root).basicIterator();
-		int offset = 0;
+		return newNode;
+	}
+
+	private void setLeafNodesTotalOffset(int offset, BidiTreeIterator<AbstractNode> iterator) {
 		while(iterator.hasNext()) {
 			AbstractNode node = iterator.next();
 			if (node instanceof LeafNode) {
@@ -290,6 +291,20 @@ public class NodeModelBuilder {
 				offset += node.getTotalLength();
 			}
 		}
+	}
+
+	public void replaceAndTransferLookAhead(INode oldNode, INode newRootNode) {
+		AbstractNode newNode = doReplaceAndTransferLookAhead(oldNode, newRootNode);
+		ICompositeNode root = newNode.getRootNode();
+		setLeafNodesTotalOffset(0, ((AbstractNode) root).basicIterator());
+	}
+
+	/**
+	 * @since 2.26
+	 */
+	 public void replaceAndTransferLookAheadKeepingOffset(INode oldNode, INode newRootNode) {
+		AbstractNode newNode = doReplaceAndTransferLookAhead(oldNode, newRootNode);
+		setLeafNodesTotalOffset(oldNode.getTotalOffset(), newNode.basicIterator());
 	}
 
 	protected void replaceWithoutChildren(AbstractNode oldNode, AbstractNode newNode) {


org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -272,24 +272,30 @@ protected void replaceChildren(AbstractNode oldNode, AbstractNode newNode) {
...
 			}
 		}
+
+	}
+	public void replaceAndTransferLookAhead(INode oldNode, INode newRootNode) {
+		replaceAndTransferLookAhead(oldNode, ((CompositeNode) newRootNode).basicGetFirstChild(), 0);
+	}
+
+	public void replaceAndTransferLookAheadKeepingOffset(INode oldNode, INode newRootNode) {
+		replaceAndTransferLookAhead(oldNode, ((CompositeNode) newRootNode).basicGetFirstChild(), oldNode.getTotalOffset());
```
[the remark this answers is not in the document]
szarnekow: Unit tests that prove that the node model is consistent afterwards, e.g. by means of the InvariantChecker.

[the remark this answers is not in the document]
rubenporras: You were right, my version had a defect. I sent I new one but know tests are failing. I will take a look at it.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -273,16 +273,17 @@ protected void replaceChildren(AbstractNode oldNode, AbstractNode newNode) {
...
 			CompositeNode newCompositeNode = (CompositeNode) newNode;
 			newCompositeNode.basicSetLookAhead(((ICompositeNode) oldNode).getLookAhead());
 		}
-		ICompositeNode root = newNode.getRootNode();
-		BidiTreeIterator<AbstractNode> iterator = ((AbstractNode) root).basicIterator();
-		int offset = 0;
+		return newNode;
+	}
+
+	private void transferLookAhead(int offset, BidiTreeIterator<AbstractNode> iterator) {
```
szarnekow, line 286: This method is not transfering any lookAhead.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -273,16 +273,17 @@ protected void replaceChildren(AbstractNode oldNode, AbstractNode newNode) {
 		}
 	}
 	
-	public void replaceAndTransferLookAhead(INode oldNode, INode newRootNode) {
+	private AbstractNode replace(INode oldNode, INode newRootNode) {
 		AbstractNode newNode = ((CompositeNode) newRootNode).basicGetFirstChild();
 		replaceWithoutChildren((AbstractNode) oldNode, newNode);
 		if (oldNode instanceof ICompositeNode && newNode instanceof CompositeNode) {
 			CompositeNode newCompositeNode = (CompositeNode) newNode;
 			newCompositeNode.basicSetLookAhead(((ICompositeNode) oldNode).getLookAhead());
```
szarnekow, line 281: Here is where the lookAhead is transfered.

rubenporras: Thanks. I have renamed the methods

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -273,16 +273,17 @@ protected void replaceChildren(AbstractNode oldNode, AbstractNode newNode) {
...
 		}
-		ICompositeNode root = newNode.getRootNode();
-		BidiTreeIterator<AbstractNode> iterator = ((AbstractNode) root).basicIterator();
-		int offset = 0;
+		return newNode;
+	}
+
+	private void adjustLookAhead(int offset, BidiTreeIterator<AbstractNode> iterator) {
```
szarnekow, line 286: This method sets the total offsets into the leaf nodes. It's not related to look ahead

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
```diff
@@ -292,6 +293,20 @@ public void replaceAndTransferLookAhead(INode oldNode, INode newRootNode) {
...
+	public void replaceAndTransferLookAhead(INode oldNode, INode newRootNode) {
+		AbstractNode newNode = doReplaceAndTransferLookAhead(oldNode, newRootNode);
+		ICompositeNode root = newNode.getRootNode();
+		adjustLookAhead(0, ((AbstractNode) root).basicIterator());
+	}
+
+	/**
+	 * @since 2.26
+	 */
+	 public void replaceAndTransferLookAheadKeepingOffset(INode oldNode, INode newRootNode) {
```
szarnekow, line 305: I'm afraid this cannot be accepted without a test that proves the following scenario impossible:

Assuming we have a model with only three leaf nodes ( Name(totalOffset, totalEndOffset) ):
- L1 (0, 10)
- L2 (10, 20)
- L3 (20, 30)

`oldNode` encloses L2, `newRootNode` has a single leaf beneath of length 5 ( L4(0, 5) at this point ).
Now we are calling `replaceAndTransferLookAheadKeepingOffset` effectively ripping out `oldNode` with its L2 and replacing that one by the first child of newRootNode with its L4.
According to the logic, L4 would be adjusted towards L4(10, 15). Afterwards, the model is inconsistent since there's a gap of 5 characters before L3 starts at offset 20.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
rubenporras: Thanks for your patience. It only highlights my bad understanding of the code.

Lucky for me, I am only submitting working code that someone else did many years ago, I only need to extract the changes  in a way that they are mergeable in sensible and usable by others in chucks to xtext-core :)

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/NodeModelBuilder.java
rubenporras: I will check this.

szarnekow: @rubenporras Thank you very much for upstreaming the fixes you use in production since quite some time. But please keep in mind that we cannot accept this patch without at least some test coverage.

rubenporras: Hi @szarnekow ,

yes, I understand that I need to add test coverage, it will take a while for me to find time to understand how to submit a test for it, but I will come to it eventually. Is it fine for you if I keep the PR open for the time being?

szarnekow: Sure, that works for me.

rubenporras: Not yet, I promised to write some unit tests, but I did not have time for it so far.

rubenporras:
yes, let us close it until I provide the tests. We have a solution in place since years that uses reflection to access the internals of `CompositeNode` and `AbstractNode` (Using reflection is not nice but does the trick), so we do not depend on this PR. The PR is more part of my attempt to submit upstream whatever improvements we have done in the past to Xtext that could be meaningful to others.

PR #1828, opened 2022-02-09
Add public getter for the language name

rubenporras:
Add public getter for the language name so that suclasses can access the
language name.


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGeneratorLanguage.java
```diff
@@ -138,6 +138,10 @@ public void setName(String name) {
 		this.name = name;
 	}
 
+	public String getName() {
```
cdietrich, line 141: should this be protected?

cdietrich: @szarnekow what do you think?

szarnekow: Judging from the code we have a few inconsistencies here:
- There is a public accessor getGrammarUri() but no chance to directly read the name.
- Either grammarUri or name need to be set from the workflow
- getGrammar() would allow to access the name as well as the URI after initialize.. was called
- JavaDoc on setters but non on getters :(
- Some getters (e.g. getFileExtensions, getGrammar) do only make sense after initialize was called.

Not all of these can / should be fixed.

About the issue at hand:
I'd think a `public final getName()` would make sense or we replace all usages of the field name by usages of the getter. In that case, an overridable getter is ok, too. In any case, please add an `@since` tag and repeat the JavaDoc of of the setter on the new getter.

szarnekow (approved):

cdietrich:
thx @rubenporras @szarnekow

PR #1832, opened 2022-02-15
Fix NPE if there is no resource provider for an URI

rubenporras:
Fix NPE if there is no resource provider for an URI by checking for null
and defaulting to false if null, the null check is equivalent to the one
already present in getDeltasForDeletedResources. In addition mark the
URI for removal.

The resource provider can be null in our development environments or if
a persisted index is being reused after some language was de-installed.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
@@ -198,10 +198,18 @@ public class Indexer {
 				.transform(IResourceDescription::getURI).copyInto(new HashSet<>());
 		remainingURIs.removeAll(FluentIterable.from(deltas).transform(Delta::getUri).toSet());
 		List<URI> allAffected = FluentIterable.from(remainingURIs).filter(it -> {
-			IResourceDescription.Manager manager = context.getResourceServiceProvider(it)
-					.getResourceDescriptionManager();
-			IResourceDescription resourceDescription = previousIndex.getResourceDescription(it);
-			return isAffected(resourceDescription, manager, allDeltas, allDeltas, newIndex);
+			IResourceServiceProvider resourceServiceProvider = context.getResourceServiceProvider(it);
+			if (resourceServiceProvider != null) {
+				IResourceDescription.Manager manager = resourceServiceProvider.getResourceDescriptionManager();
+				IResourceDescription resourceDescription = previousIndex.getResourceDescription(it);
+				return isAffected(resourceDescription, manager, allDeltas, allDeltas, newIndex);
+			} else {
+				IResourceDescription.Delta delta = getDeltaForDeletedResource(it, previousIndex);
+				if (delta != null) {
+					deltas.add(delta);
+				}
+				return false;
+			}
 		}).toList();
 		deltas.addAll(getDeltasForChangedResources(allAffected, previousIndex, context));
 		return new Indexer.IndexResult(deltas, newIndex);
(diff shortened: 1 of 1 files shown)


cdietrich:
how do you manage to have index entries for stuff with no resourceserviceprovider
i feel you workaround a contract breach somewhere else

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
```diff
@@ -198,10 +198,13 @@ public URI getURI() {
 				.transform(IResourceDescription::getURI).copyInto(new HashSet<>());
 		remainingURIs.removeAll(FluentIterable.from(deltas).transform(Delta::getUri).toSet());
 		List<URI> allAffected = FluentIterable.from(remainingURIs).filter(it -> {
-			IResourceDescription.Manager manager = context.getResourceServiceProvider(it)
-					.getResourceDescriptionManager();
-			IResourceDescription resourceDescription = previousIndex.getResourceDescription(it);
-			return isAffected(resourceDescription, manager, allDeltas, allDeltas, newIndex);
+			IResourceServiceProvider resourceServiceProvider = context.getResourceServiceProvider(it);
```
cdietrich, line 201: how do you end up with non rsp resources in remainingURIs

rubenporras:
Hi @cdietrich ,

you are right that it will not happen in fully setup environment. 

However in our development setups, sometimes we manually test with our pre-calculated index stored in a database and with a subset of the languages that where used to calculate that index (that is because we have additional languages in separate repositories, that where used to calculate the index but we do not always check out to manually test). On this combination there are URIs in the index which are not built (since the extension is not registered), yet we process them to check if they are affected or not.

I think the null check is good compromise. I think that you also have the same pattern in org.eclipse.xtext.builder.clustering.ClusteringBuilderState.getResourceDescriptionManager(URI) which is used in org.eclipse.xtext.builder.clustering.ClusteringBuilderState.queueAffectedResources(Set<URI>, IResourceDescriptions, CurrentDescriptions, Collection<Delta>, Collection<Delta>, BuildData, IProgressMonitor) and will skip uris without an rsp.

Does it make sense now?

cdietrich:
yes but i am still not sure if we should implement such bug hiding enhancements

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
szarnekow: I can imagine a persisted index that is being reused after some language was deinstalled.
In any case, this should at least mark the URI for removal.

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
```diff
@@ -213,21 +221,32 @@ public URI getURI() {
...
-				operationCanceledManager.checkCanceled(context.getCancelIndicator());
-				IResourceDescription oldDescription = oldIndex != null ? oldIndex.getResourceDescription(deleted)
-						: null;
-				if (oldDescription != null) {
-					DefaultResourceDescriptionDelta delta = new DefaultResourceDescriptionDelta(oldDescription, null);
-					deltas.add(delta);
+		if (oldIndex != null) {
+			for (URI deleted : request.getDeletedFiles()) {
+				IResourceServiceProvider resourceServiceProvider = context.getResourceServiceProvider(deleted);
+				if (resourceServiceProvider != null) {
```
rubenporras, line 227: I have kept the check, even though I does not know what it is for. The resourceServiceProvider is not used, so from my naive point of view, it could be removed, the get and the null check could be removed, right?

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
```diff
@@ -213,21 +221,32 @@ public URI getURI() {
...
-		for (URI deleted : request.getDeletedFiles()) {
-			IResourceServiceProvider resourceServiceProvider = context.getResourceServiceProvider(deleted);
-			if (resourceServiceProvider != null) {
-				operationCanceledManager.checkCanceled(context.getCancelIndicator());
-				IResourceDescription oldDescription = oldIndex != null ? oldIndex.getResourceDescription(deleted)
-						: null;
-				if (oldDescription != null) {
-					DefaultResourceDescriptionDelta delta = new DefaultResourceDescriptionDelta(oldDescription, null);
-					deltas.add(delta);
+		if (oldIndex != null) {
```
rubenporras, line 224: I have pushed the null check on the oldIndex outside of the loop, but I am not sure it can ever happen that it will be null, if it would be null, the loop I have modified (line 204 on this PR) would throw an NPE. So I think we could remove it as well

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
rubenporras: Thanks, I have done that as, now, the URI will  be marked for removal. I also left two comments for two existing null checks in the code, which I think could be removed, in case you want to take a look at them.

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
cdietrich: the request might be filled by outside tool

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
```diff
@@ -213,21 +221,32 @@ public URI getURI() {
...
 				}
 			}
 		}
 		return deltas;
 	}
 
+	/**
+	 * Gets a delta for a resource that shall be deleted.
+	 */
+	protected IResourceDescription.Delta getDeltaForURI(URI uri, ResourceDescriptionsData oldIndex) {
```
cdietrich, line 242: maybe add the deleted fact somewhere to method name?

rubenporras: done

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
rubenporras: @szarnekow , I have added code to mark the URI for removal. Is the code now good for you? Thanks

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
```diff
@@ -213,21 +221,32 @@ public URI getURI() {
...
 				}
 			}
 		}
 		return deltas;
 	}
 
+	/**
+	 * Gets a delta for a resource that shall be deleted.
+	 */
+	protected IResourceDescription.Delta getDeltaForResourceToDelete(URI uri, ResourceDescriptionsData oldIndex) {
```
szarnekow, line 242: Can we align the method names. Right now we have `getDeltasForDeletedResources` and `getDeltaForResourceToDelete`.
Also please add an `@Since` tag to the new method.

szarnekow:
I'm inclined to merge. @cdietrich objections?

cdietrich (approved):

PR #1877, opened 2022-04-12
Allow configuration by custom LS implementations of the live resourceSet

rubenporras:
used for text edits in the same way that we allow for configuration for
the other resource sets created by the Language Server


szarnekow (approved):
I think it's ok.
It doesn't necessarily improve the situation in the ProjectManager with `createFreshResourceSet` which claims in its JavaDoc to return an empty resource, `createNewResourceSet` which does indeed return an empty resource set and now we have a `createLiveScopeResourceSet`, too. But I don't see a great way to improve on that.

PR #1881, opened 2022-04-14
Add API for textual Code Actions

rubenporras:
In addition, improve naming of variables in the existing API to use
diagnostics for org.eclipse.lsp4j.Diagnostic instead of issues and
errors, and resolutions for DiagnosticResolution instead of issues in
the DiagnosticResolutionAcceptor

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
@@ -200,14 +200,14 @@ public abstract class AbstractIdeQuickfixTest {
 
 		options.setCodeActionParams(codeActionParams);
 
-		List<DiagnosticResolution> actualIssueResolutions = IterableExtensions.sortBy(quickFixProvider.getResolutions(options, issue),
-				DiagnosticResolution::getLabel);
-		assertEquals("The number of quickfixes does not match!", expectedSorted.size(),
-				actualIssueResolutions.size());
+		for (QuickfixExpectation expectedIssueResolution: expectedSorted) {
+			List<DiagnosticResolution> actualIssueResolutions = 
+					quickFixProvider.getResolutions(options, issue).stream()
+					.filter(r -> r.getLabel().equals(expectedIssueResolution.getLabel()))
+					.collect(Collectors.toList());
+			assertEquals("More than one quickfix available!", 1, actualIssueResolutions.size());
 
-		for (int i = 0; i < actualIssueResolutions.size(); i++) {
-			DiagnosticResolution actualIssueResolution = actualIssueResolutions.get(i);
-			QuickfixExpectation expectedIssueResolution = expectedSorted.get(i);
+			DiagnosticResolution actualIssueResolution = actualIssueResolutions.get(0);
 
 			assertEquals(expectedIssueResolution.label, actualIssueResolution.getLabel());
 			assertEquals(expectedIssueResolution.description, actualIssueResolution.getLabel());
--- a/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/quickfix/TestLanguageQuickFixProvider.java
+++ b/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/quickfix/TestLanguageQuickFixProvider.java
@@ -8,6 +8,9 @@
  *******************************************************************************/
 package org.eclipse.xtext.ide.tests.testlanguage.ide.quickfix;
 
+import java.util.Collections;
+
+import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.xtext.ide.editor.quickfix.AbstractDeclarativeIdeQuickfixProvider;
 import org.eclipse.xtext.ide.editor.quickfix.DiagnosticResolutionAcceptor;
 import org.eclipse.xtext.ide.editor.quickfix.QuickFix;
@@ -17,12 +20,26 @@ import org.eclipse.xtext.xbase.lib.StringExtensions;
 
 public class TestLanguageQuickFixProvider extends AbstractDeclarativeIdeQuickfixProvider {
 
-	@QuickFix(TestLanguageValidator.INVALID_NAME)
-	public void fixLowerCaseName(DiagnosticResolutionAcceptor acceptor) {
-		acceptor.accept("Change element name to first upper", obj -> {
-			final TypeDeclaration element = (TypeDeclaration) obj;
-			element.setName(StringExtensions.toFirstUpper(element.getName()));
-		});
+	public static String EMF_QF_LABEL = "Change element name to first upper using object modification";
+	public static String TEXT_QF_LABEL = "Change element name to first upper using text replacement";
+
+	private String fixedName(TypeDeclaration declaration) {
+		return StringExtensions.toFirstUpper(declaration.getName());
 	}
 
+	@QuickFix(TestLanguageValidator.INVALID_NAME)
+	public void fixLowerCaseName(DiagnosticResolutionAcceptor acceptor) {
+		acceptor.accept(EMF_QF_LABEL, obj -> {
+			TypeDeclaration declaration = (TypeDeclaration) obj;
+			declaration.setName(fixedName(declaration));
+		});
+	}
+	
+	@QuickFix(TestLanguageValidator.INVALID_NAME)
+	public void textFixLowerCaseName(DiagnosticResolutionAcceptor acceptor) {
+		acceptor.accept(TEXT_QF_LABEL,  (diagnostic, obj, document) -> {
+			TextEdit textEdit = new TextEdit(diagnostic.getRange(), "type " + fixedName((TypeDeclaration) obj) + " {\n}");
+			return Collections.singletonList(textEdit);
+		});
+	}
 }
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
@@ -3,7 +3,7 @@
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
 package org.eclipse.xtext.ide.editor.quickfix;
@@ -13,15 +13,20 @@ import org.eclipse.emf.common.util.WrappedException;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.resource.ResourceSet;
 import org.eclipse.lsp4j.CodeActionParams;
+import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.WorkspaceEdit;
 import org.eclipse.xtext.ide.editor.quickfix.DiagnosticModificationContext.Factory;
 import org.eclipse.xtext.ide.serializer.IChangeSerializer;
 import org.eclipse.xtext.ide.serializer.IChangeSerializer.IModification;
+import org.eclipse.xtext.ide.server.Document;
 import org.eclipse.xtext.ide.server.ILanguageServerAccess;
+import org.eclipse.xtext.ide.server.TextEditAcceptor;
 import org.eclipse.xtext.ide.server.codeActions.ICodeActionService2.Options;
 import org.eclipse.xtext.ide.server.rename.ChangeConverter2;
 import org.eclipse.xtext.resource.EObjectAtOffsetHelper;
 import org.eclipse.xtext.resource.XtextResource;
+
+import com.google.common.collect.MoreCollectors;
 
 /**
  * @author Heinrich Weichert
@@ -34,11 +39,31 @@ public class DiagnosticResolution {
 
 	private IModification<EObject> modification;
 
+	private final ITextModification textModification;
+
 	private Factory factory;
 
 	private int relevance;
 
 	private Options options;
+
+	/**
+	 * @since 2.27
+	 */
+	public DiagnosticResolution(String label, Factory modificationContextFactory, ITextModification textModification) {
+		this(label, modificationContextFactory, textModification, 0);
+	}
+
+	/**
+	 * @since 2.27
+	 */
+	public DiagnosticResolution(String label, Factory modificationContextFactory, ITextModification textModification, int relevance) {
+		this.label = label;
+		this.factory = modificationContextFactory;
+		this.modification = null;
+		this.textModification = textModification;
+		this.relevance = relevance;
+	}
 
 	public DiagnosticResolution(String label, Factory modificationContextFactory, IModification<EObject> modification) {
 		this(label, modificationContextFactory, modification, 0);
@@ -49,6 +74,7 @@ public class DiagnosticResolution {
 		this.label = label;
 		this.factory = modificationContext;
 		this.modification = modification;
+		this.textModification = null;
 		this.relevance = relevance;
 	}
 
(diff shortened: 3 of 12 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -34,12 +41,22 @@ public class DiagnosticResolution {
...
 
+	private final ITextModification<Diagnostic, EObject, Document, List<TextEdit>> textModification;
+
 	private Factory factory;
 
 	private int relevance;
 
 	private Options options;
 
+	public DiagnosticResolution(final String label, final Factory modificationContextFactory, final ITextModification<Diagnostic, EObject, Document, List<TextEdit>> textModification) {
```
tivervac, line 52: No need for final in method parameters

tivervac: Would also need an `@since` I suppose

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/ITextModification.java
```diff
@@ -0,0 +1,19 @@
...
+package org.eclipse.xtext.ide.editor.quickfix;
+
+/**
+ * @author Rub�n Porras Campo - Initial contribution and API
+ *
+ * @since 2.27
+ */
+@FunctionalInterface
+public interface ITextModification<A, B, C, R> {
+  R apply(A a, B b, C c);
```
tivervac, line 18: Don't think any code uses two space as indent

tivervac: What are a, b and c? Might be useful to give them better names or document them

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/TextEditAcceptor.java
```diff
@@ -0,0 +1,76 @@
...
+import org.eclipse.lsp4j.WorkspaceEditCapabilities;
+import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.eclipse.xtext.ide.server.Document;
+import org.eclipse.xtext.ide.server.ILanguageServerAccess;
+
+public class TextEditAcceptor {
+	private final boolean useDocumentChanges;
+	private final WorkspaceEdit edit;
+
+	public TextEditAcceptor(final WorkspaceEdit edit, final ILanguageServerAccess access) {
```
tivervac, line 36: No need for final parameters

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
tivervac: You probably want a second constructor such that the `relevance` can be set

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/ITextModification.java
szarnekow: Also I don't feel like the current ITextModification looks like an ITextModification at all. More like a Function3.

Judging from the usages, the parameterizations are the same for all. Maybe remove the unnecessary type parameters.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
rubenporras: done.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/ITextModification.java
rubenporras: Done, i have removed the type parameters

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/TextEditAcceptor.java
rubenporras: done, we have different policy according to tabs and spaces and usages of final at Avaloq, and these slipped through when coding.

rubenporras:
@cdietrich, @tivervac, from my point of view I am like the change as it is now. I have cleaned up the warnings and added one test. 

Would you mind taking a look? 

I would be happy if we could merge it before M2 is out so that I can upgrade and use the new API.

Thanks

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolutionAcceptor.java
```diff
@@ -35,12 +35,16 @@ public DiagnosticResolutionAcceptor(DiagnosticModificationContext.Factory modifi
 	}
 
 	public void accept(String label, IModification<EObject> modification) {
-		issues.add(new DiagnosticResolution(label, modificationContextFactory, modification));
+		resolutions.add(new DiagnosticResolution(label, modificationContextFactory, modification));
+	}
+
+	public void accept(final String label, final ITextModification modification) {
```
szarnekow, line 41: Nit: final parameters are not adding value here. Not actionable.

You may want to use Oomph to setup your Xtext dev environment, to simplify the review process for you and the reviewer though :)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
```diff
@@ -33,10 +33,10 @@ public interface IQuickFixProvider {
 	 *
 	 * @param options
 	 *            Contextual action options
-	 * @param issue
-	 *            EMF diagnostic issue
+	 * @param diagnostic
+	 *            the diagnostic
```
szarnekow, line 37: Thank you for the attention to detail.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/ITextModification.java
```diff
@@ -0,0 +1,26 @@
...
+import org.eclipse.xtext.ide.server.Document;
+
+/**
+ * @author Rub�n Porras Campo - Initial contribution and API
+ *
+ * @since 2.27
+ */
+@FunctionalInterface
+public interface ITextModification {
+	List<TextEdit> apply(Diagnostic diagnostic, EObject object, Document document);
```
szarnekow, line 25: Please add a comment that describes which `EObject object` one would expect here.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -61,28 +85,31 @@ public String getLabel() {
...
+			WorkspaceEdit edit = new WorkspaceEdit();
+			if (modification != null) {
+				DiagnosticModificationContext modificationContext = factory.createModificationContext();
+				ChangeConverter2 changeConverter = modificationContext.getConverterFactory().create(edit, access);
+				IChangeSerializer serializer = modificationContext.getSerializer();
+				serializer.addModification(obj, modification);
+				serializer.applyModifications(changeConverter);
+			} else {
+				TextEditAcceptor textEditAcceptor = new TextEditAcceptor(edit, access);
+				Diagnostic diagnostic = params.getContext().getDiagnostics().get(0);
```
szarnekow, line 110: I'm a little scared about the (`.get(0)`) here. This is probably incorrect.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/TextEditAcceptor.java
```diff
@@ -0,0 +1,74 @@
...
+		}
+		this.useDocumentChanges = documentChanges == Boolean.TRUE;
+		if (this.useDocumentChanges) {
+			this.edit.setDocumentChanges(new ArrayList<>());
+		} else {
+			this.edit.setChanges(new LinkedHashMap<>());
+		}
+	}
+
+	public Object accept(String theUri, Document document, List<TextEdit> textEdits) {
```
szarnekow, line 61: This should be a void method as most of the Acceptor-style APIs in Xtext are?

Currently it either return `true` or a random old textEdits or null, none of these do make sense to me.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter2.java
```diff
@@ -163,17 +130,7 @@ protected void _handleReplacements(ITextDocumentChange change) {
 	}
 
 	protected Object addTextEdit(String theUri, Document document, TextEdit... textEdits) {
```
szarnekow, line 132: Please refactor to void method, too. Probably fallout from some Xtend to Java rewrite in the past. line 124 can always return null.

szarnekow (commented):
Thank you Ruben. Please let me know if you disagree with some of the comments.

Do you see - by chance - a path towards having a common APi for Eclipse Quickfixes and LSP Quickfixes based on the API that is introduced here?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter2.java
cdietrich: i assume the problem is that the doRead calls want a return value. what do you propose?

szarnekow: I'd let this particular doRead either return null or make it an IUnitOfWork.Void

cdietrich: i guess an explict return null would do the latter

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolutionAcceptor.java
rubenporras: Sorry for that. I did that this morning already to write the test, it was just too error-prone to do that in GitHub editor :) It will be part of the next push.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
rubenporras: The code in [org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixCodeActionService.java](https://github.com/eclipse/xtext-core/blob/942def122c1b49feac1713de2edeae514b0b1b7c/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixCodeActionService.java#L69-71) makes sure that we have a singleton list, with exactly one element, so as the code currently is, I do not think this is a mistake, as it relies on an invariant, but I also dislike it. I left the code as it is because I am not sure what should be better than this.
Should we add a precondition check that tests that the list has indeed one and only one argument? Would you prefer that?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter2.java
rubenporras: this would be a breaking change if someone relies on it, but if it is fine for you, it is for me as well, so I have done the change. Thanks

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/TextEditAcceptor.java
rubenporras: done

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter2.java
```diff
@@ -120,7 +87,8 @@ protected void _handleReplacements(IEmfResourceChange change) {
 				Document document = context.getDocument();
 				Range range = new Range(document.getPosition(0), document.getPosition(document.getContents().length()));
 				TextEdit textEdit = new TextEdit(range, newContent);
-				return addTextEdit(uri, document, textEdit);
+				addTextEdit(uri, document, textEdit);
+				return null;
```
cdietrich, line 91: :+1:

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
szarnekow: Yes, please use something like com.google.common.collect.Iterables.getOnlyElement(Iterable<T>) to document and assert that invariant in code.

rubenporras: Nice, I was not aware of that Guava API. I have opted for using what they document as the variant for Java 8 using streams.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
```diff
@@ -200,14 +200,14 @@ public ResourceSet newLiveScopeResourceSet(URI uri) {
...
-
-		for (int i = 0; i < actualIssueResolutions.size(); i++) {
-			DiagnosticResolution actualIssueResolution = actualIssueResolutions.get(i);
-			QuickfixExpectation expectedIssueResolution = expectedSorted.get(i);
+		for (QuickfixExpectation expectedIssueResolution: expectedSorted) {
+			List<DiagnosticResolution> actualIssueResolutions = 
+					quickFixProvider.getResolutions(options, issue).stream()
+					.filter(r -> r.getLabel().equals(expectedIssueResolution.getLabel()))
+					.collect(Collectors.toList());
+			assertEquals("More than one quickfixes available!", 1, actualIssueResolutions.size());
```
tivervac, line 208: s/quickfixes/quickfix/

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/quickfix/TestLanguageQuickFixProvider.java
```diff
@@ -17,12 +20,26 @@
...
+		return StringExtensions.toFirstUpper(declaration.getName());
+	}
+
 	@QuickFix(TestLanguageValidator.INVALID_NAME)
 	public void fixLowerCaseName(DiagnosticResolutionAcceptor acceptor) {
-		acceptor.accept("Change element name to first upper", obj -> {
-			final TypeDeclaration element = (TypeDeclaration) obj;
-			element.setName(StringExtensions.toFirstUpper(element.getName()));
+		acceptor.accept(EMF_QF_LABEL, obj -> {
+			final TypeDeclaration declaration = (TypeDeclaration) obj;
```
tivervac, line 33: Probably also from x2j, no need for the final

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/ITextModification.java
```diff
@@ -0,0 +1,39 @@
...
+@FunctionalInterface
+public interface ITextModification {
+	/**
+	 *
+	 * Returns a list of {@link TextEdit} given a diagnostic, object and document.
+	 *
+	 * @param diagnostic
+	 *            the {@link Diagnostic}
+	 * @param object
+	 *            the {@link EObject} next to the offset at the start of the range of the diagnostic
```
tivervac, line 32: This is a bit hard to read, how about 

`the {@link EObject} after {@code diagnostic.getRange().getStart()}`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/ITextModification.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @author Rub�n Porras Campo - Initial contribution and API
+ *
+ * @since 2.27
+ */
+@FunctionalInterface
+public interface ITextModification {
+	/**
+	 *
+	 * Returns a list of {@link TextEdit} given a diagnostic, object and document.
```
tivervac, line 27: Missing `s` after `{@link TextEdit}`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixCodeActionService.java
```diff
@@ -44,26 +46,28 @@ public class QuickFixCodeActionService implements ICodeActionService2 {
...
-					result.add(Either.forRight(createFix(resolution, error)));
-				}
+		if (!handleQuickfixes) {
+			return Collections.emptyList();
+		}
+
+		List<Either<Command, CodeAction>> result = new ArrayList<>();
+		for (Diagnostic diagnostic : options.getCodeActionParams().getContext().getDiagnostics()) {
+			List<DiagnosticResolution> resolutions = quickfixes.getResolutions(options, diagnostic).stream()
+					.sorted(Comparator.comparing(DiagnosticResolution::getLabel)).collect(Collectors.toList());
```
tivervac, line 60: Not sure this comparator will work. I had a quick look at the uses of `getLabel` and I see some places null checking it. The datastructure `DiagnosticResolution` does not make sure that the label is not null and as far as I can see, `Comparator.comparing` can't handle keyExtractors that can produce nulls

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter2.java
```diff
@@ -120,7 +87,8 @@ protected void _handleReplacements(IEmfResourceChange change) {
...
 			}).get();
 		} catch (InterruptedException | ExecutionException | IOException e) {
 			throw Exceptions.sneakyThrow(e);
```
tivervac, line 94: Not your code, but `sneakyThrow` is evil. It will disguise checked Exceptions as unchecked exceptions and throw them without mentioning the checked exceptions in a function's signature.

Since you have to edit the signature to correctly handle this, not sure if there's a good solution here

tivervac (changes requested):

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/editor/quickfix/AbstractIdeQuickfixTest.java
rubenporras: done

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/quickfix/TestLanguageQuickFixProvider.java
rubenporras: done

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/ITextModification.java
rubenporras: done

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter2.java
rubenporras: Thanks. I would be happy changing it, but since it is unrelated to my changes, I would leave it be unless someone wants me to do the change.

tivervac (commented):
LGTM now

tivervac (approved):

cdietrich:
@szarnekow ok to go?

szarnekow (approved):
Thank you @rubenporras

szarnekow:
> @szarnekow , at the moment I do not see a path towards having a common API [..]

It does make sense, yes. Thank you for sharing your thoughts on this.

cdietrich:
thx @rubenporras

PR #1883, opened 2022-04-26
Fix typo in class name DocumentSymbolConfiguration

rubenporras:
Signed-off-by: Rubén Porras Campo <pcr@avaloq.com>


miklossy:
I am not sure if this change is considered as an API break.
@cdietrich : What do you think?

cdietrich:
i dont care in this area. lsp4j breaks all the time

rubenporras:
I am also not sure how you handle the test API. If you consider it a breaking change, I can submit an updated version that leaves the old class in place with a deprecated annotation.

miklossy (approved):

miklossy:
Thx @rubenporras !

PR #1885, opened 2022-04-28
Extend the Quickfix API to support semantic QFs that use the data

rubenporras:
contained in the diagnostic.

Also, add one convenience method to create textual QFs which just
replace the region of the diagnostic with a new text.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
@@ -17,6 +17,7 @@ import java.util.stream.Collectors;
 
 import org.apache.log4j.Logger;
 import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.xtext.ide.server.codeActions.ICodeActionService2.Options;
 
 import com.google.common.annotations.Beta;
@@ -83,4 +84,18 @@ public class AbstractDeclarativeIdeQuickfixProvider implements IQuickFixProvider
 		return getResolutions(options, getFixMethods(diagnostic));
 	}
 
+	/**
+	 * Creates a singleton list with only one {@link TextEdit} that replaces the region of the diagnostic with the given
+	 * text
+	 * 
+	 * @param diagnostic
+	 *            the {@link Diagnostic}
+	 * @param text
+	 *            the text
+	 * @return a singleton list with only one {@link TextEdit}
+	 * @since 2.27
+	 */
+	protected List<TextEdit> createTextEdit(Diagnostic diagnostic, String text) {
+		return Collections.singletonList(new TextEdit(diagnostic.getRange(), text));
+	}
 }
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
@@ -37,7 +37,7 @@ public class DiagnosticResolution {
 
 	private String label;
 
-	private IModification<EObject> modification;
+	private ISemanticModification modification;
 
 	private final ITextModification textModification;
 
@@ -46,6 +46,16 @@ public class DiagnosticResolution {
 	private int relevance;
 
 	private Options options;
+
+	@Deprecated//(forRemoval=true,since="2.27")
+	public DiagnosticResolution(String label, Factory modificationContextFactory,  IModification<EObject> modification) {
+		this(label, modificationContextFactory, (diagnostic, object) -> modification, 0);
+	}
+
+	@Deprecated//(forRemoval=true,since="2.27")
+	public DiagnosticResolution(String label, Factory modificationContextFactory,  IModification<EObject> modification, int relevance) {
+		this(label, modificationContextFactory, (diagnostic, object) -> modification, relevance);
+	}
 
 	/**
 	 * @since 2.27
@@ -65,11 +75,17 @@ public class DiagnosticResolution {
 		this.relevance = relevance;
 	}
 
-	public DiagnosticResolution(String label, Factory modificationContextFactory, IModification<EObject> modification) {
+	/**
+	 * @since 2.27
+	 */
+	public DiagnosticResolution(String label, Factory modificationContextFactory, ISemanticModification modification) {
 		this(label, modificationContextFactory, modification, 0);
 	}
 
-	public DiagnosticResolution(String label, Factory modificationContext, IModification<EObject> modification,
+	/**
+	 * @since 2.27
+	 */
+	public DiagnosticResolution(String label, Factory modificationContext, ISemanticModification modification,
 			int relevance) {
 		this.label = label;
 		this.factory = modificationContext;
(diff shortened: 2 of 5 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -65,11 +64,11 @@ public DiagnosticResolution(String label, Factory modificationContextFactory, IT
 		this.relevance = relevance;
 	}
 
-	public DiagnosticResolution(String label, Factory modificationContextFactory, IModification<EObject> modification) {
```
rubenporras, line 68: these are both breaking changes, if someone uses this API. Shall I create additional methods to maintain backwards compatibility?

cdietrich: maybe you can keep one,
and mark it with
`@Deprecated//(forRemoval=true,since="2.27") `

rubenporras: Thanks, I did that (without commenting the properties of the annotation, since we require Java 11 I think it is not needed)

cdietrich: we are still on java 8

rubenporras:
Hi @cdietrich, I tried to put it as part of the commit message, but maybe it was not enaugh with the small summary. The background is, as commented on https://github.com/eclipse/xtext-core/pull/1881, on the QF API based on EMF modification, the Diagnostic is not available, as the Diagnostic is needed for accessing the data which can be added by a validation. Without it, validations which pre-compute data for the quickfix, cannot be used.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
rubenporras: thx, fixed

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
```diff
@@ -83,4 +84,18 @@ public List<DiagnosticResolution> getResolutions(Options options, Diagnostic dia
...
+	 * text
+	 * 
+	 * @param diagnostic
+	 *            the {@link Diagnostic}
+	 * @param text
+	 *            the text
+	 * @return a singleton list with only one {@link TextEdit}
+	 * @since 2.27
+	 */
+	protected List<TextEdit> createTextEdit(Diagnostic diagnostic, String text) {
```
szarnekow, line 98: Gut feeling:
Given that the list is effectively not able to grow (cannot be used as a starting point for accumulating more fixes), I'd prefer to have a plain `TextEdit` as return type here.

Why did you choose a `List<TextEdit>` ?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -65,11 +75,17 @@ public DiagnosticResolution(String label, Factory modificationContextFactory, IT
...
+	 */
+	public DiagnosticResolution(String label, Factory modificationContextFactory, ISemanticModification modification) {
 		this(label, modificationContextFactory, modification, 0);
 	}
 
-	public DiagnosticResolution(String label, Factory modificationContext, IModification<EObject> modification,
+	/**
+	 * @since 2.27
+	 */
+	public DiagnosticResolution(String label, Factory modificationContext, ISemanticModification modification,
```
szarnekow, line 88: Curiosity: What is the meaning of the `relevance`? Mind adding som docs to the field and getter since you're here?

szarnekow (approved):

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -47,6 +47,16 @@ public class DiagnosticResolution {
 
 	private Options options;
 
+	@Deprecated//(forRemoval=true,since="2.27")
```
tivervac, line 50: Hmm, why is the `forRemoval` in comments?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/ISemanticModification.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @author Rub�n Porras Campo - Initial contribution and API
+ *
+ * @since 2.27
+ */
+@FunctionalInterface
+public interface ISemanticModification {
+	/**
+	 *
+	 * Returns a {@link IModification} given a diagnostic and an object.
```
tivervac, line 24: s/a/an/

tivervac (approved):

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
cdietrich: Cause Java 8

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/ISemanticModification.java
rubenporras: done

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
rubenporras: I also do not know for sure. @tivervac suggested I should add it to the API in https://github.com/eclipse/xtext-core/pull/1881#discussion_r852956806 and I did it (and continue now), as previous methods had such a parameter.
I could guess that it could be used to sort the list of QFs which are proposed, but as far as I see, the Xtext LS does not do anything useful with it. @tivervac , do you know what is it for?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
rubenporras: It returns a list because the generic interface ITextModification expects a list, and not a TextEdit. The method is meant to ease writing simple QFs. The result is the simplified QF that one can write in org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/quickfix/TestLanguageQuickFixProvider.java in this PR.

I tried offering an overloaded accept method in DiagnosticResolutionAcceptor, as in other cases, but for this case, java could not infer the types of the anonymous lambda on the caller side. That is why I when for a factory method like this one.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
tivervac: I do not. Sounds like some kind of priority metric to me

cdietrich: looks like a copy from org.eclipse.xtext.ui.editor.quickfix.IssueResolution.relevance

cdietrich:
is this ready?

rubenporras:
Hi @cdietrich , yes, it is.
Regards

tivervac (approved):

cdietrich:
@szarnekow ok now?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/AbstractDeclarativeIdeQuickfixProvider.java
szarnekow: Understandable.

szarnekow (approved):

PR #1889, opened 2022-05-05
Avoid the creation of a HashSet just to filter elements

rubenporras:
as we still need to iterate over the remaining set, we can be more
efficient by combining the two filters in the same iteration.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
@@ -194,23 +194,22 @@ public class Indexer {
 		}
 		Set<IResourceDescription.Delta> allDeltas = new HashSet<>(deltas);
 		allDeltas.addAll(request.getExternalDeltas());
-		Set<URI> remainingURIs = FluentIterable.from(previousIndex.getAllResourceDescriptions())
-				.transform(IResourceDescription::getURI).copyInto(new HashSet<>());
-		remainingURIs.removeAll(FluentIterable.from(deltas).transform(Delta::getUri).toSet());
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
```diff
@@ -124,7 +123,7 @@ protected static class ResolvedResourceDescription extends AbstractResourceDescr
 
 		private URI uri;
 
-		private ImmutableList<IEObjectDescription> exported;
+		private List<IEObjectDescription> exported;
```
cdietrich, line 126: am not sure if i like this.

cdietrich:
@rubenporras would be nice if you can also create issues. 
this makes tracking more easy

rubenporras:
I will create an issue then, no problem.

org.eclipse.xtext/src/org/eclipse/xtext/build/Indexer.java
szarnekow: If some client code starts calling modifying methods on the `exported` list, it'd be flagged with a deprecation warning iff the type is ImmutableList.

szarnekow: To be clear: This is plus. Please keep this as an immutable list.

rubenporras: I would understand that if `exported` wouldn't be private, but since it is private, no other code that this class can call it. Anyway I thought it would be an improvement, but since you do not think so, I have reverted the change.

cdietrich:
@szarnekow ok now?

szarnekow (approved):
I like performance improvements. Thank you! Approved.

PR #1898, opened 2022-05-17
Only return resolved cross references

rubenporras:
Otherwise the client will think some elements point to definitions which do not exist, and will for example jump to the beginnig of a comment or string when the user press Control-Click somewhere in one of these, which is very confusing.

### DIFF
--- a/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/rename/TestLanguageRenameService.java
+++ b/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/rename/TestLanguageRenameService.java
@@ -9,6 +9,7 @@
 package org.eclipse.xtext.ide.tests.testlanguage.rename;
 
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.xtext.Keyword;
 import org.eclipse.xtext.ide.server.rename.RenameService2;
 import org.eclipse.xtext.naming.IQualifiedNameProvider;
 import org.eclipse.xtext.naming.QualifiedName;
@@ -35,8 +36,8 @@ public class TestLanguageRenameService extends RenameService2 {
 					ICompositeNode rootNode = parseResult.getRootNode();
 					if (rootNode != null) {
 						ILeafNode leaf = NodeModelUtils.findLeafNodeAtOffset(rootNode, offset);
-						if (leaf != null && isIdentifier(leaf)) {
-							EObject element = eObjectAtOffsetHelper.resolveElementAt(xtextResource, offset);
+						if (leaf != null) {
+							EObject element = eObjectAtOffsetHelper.getElementWithNameAt(xtextResource, offset);
 							if (element != null) {
 								IQualifiedNameProvider nameProvider = xtextResource.getResourceServiceProvider()
 										.get(IQualifiedNameProvider.class);
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.java
@@ -44,7 +44,6 @@ import org.eclipse.xtext.nodemodel.ICompositeNode;
 import org.eclipse.xtext.nodemodel.ILeafNode;
 import org.eclipse.xtext.nodemodel.util.NodeModelUtils;
 import org.eclipse.xtext.parser.IParseResult;
-import org.eclipse.xtext.parsetree.reconstr.impl.TokenUtil;
 import org.eclipse.xtext.resource.EObjectAtOffsetHelper;
 import org.eclipse.xtext.resource.IResourceServiceProvider;
 import org.eclipse.xtext.resource.XtextResource;
@@ -77,15 +76,12 @@ public class RenameService2 implements IRenameService2 {
 	private IResourceServiceProvider.Registry serviceProviderRegistry;
 
 	@Inject
-	private TokenUtil tokenUtil;
-
-	@Inject
 	private IValueConverterService valueConverterService;
 
 	@Inject
 	private LinkingHelper linkingHelper;
 
-	private Function<EObject, String> attributeResolver = SimpleAttributeResolver.newResolver(String.class, "name");
+	private Function<EObject, String> attributeResolver = SimpleAttributeResolver.NAME_RESOLVER;
 
 	@Override
 	public WorkspaceEdit rename(IRenameService2.Options options) {
@@ -190,27 +186,7 @@ public class RenameService2 implements IRenameService2 {
 	}
 
 	protected EObject getElementWithIdentifierAt(XtextResource xtextResource, int offset) {
-		if (offset >= 0) {
-			if (xtextResource != null) {
-				IParseResult parseResult = xtextResource.getParseResult();
-				if (parseResult != null) {
-					ICompositeNode rootNode = parseResult.getRootNode();
-					if (rootNode != null) {
-						ILeafNode leaf = NodeModelUtils.findLeafNodeAtOffset(rootNode, offset);
-						if (leaf != null && isIdentifier(leaf)) {
-							return eObjectAtOffsetHelper.resolveElementAt(xtextResource, offset);
-						}
-					}
-				}
-			}
-		}
-		return null;
-	}
-
-	protected boolean isIdentifier(ILeafNode leafNode) {
-		return (leafNode.getGrammarElement() instanceof RuleCall
-				|| leafNode.getGrammarElement() instanceof CrossReference)
-				&& !tokenUtil.isWhitespaceOrCommentNode(leafNode);
+		return eObjectAtOffsetHelper.getElementWithNameAt(xtextResource, offset);
 	}
 
 	@Override
@@ -270,7 +246,7 @@ public class RenameService2 implements IRenameService2 {
 					element = getElementWithIdentifierAt(xtextResource, candidateOffset);
 					if (element != null && !element.eIsProxy()) {
 						ILeafNode leaf = NodeModelUtils.findLeafNodeAtOffset(rootNode, candidateOffset);
-						if (leaf != null && isIdentifier(leaf)) {
+						if (leaf != null) {
 							String convertedNameValue = getConvertedValue(leaf.getGrammarElement(), leaf);
 							String elementName = getElementName(element);
 							if (!Strings.isEmpty(convertedNameValue) && !Strings.isEmpty(elementName)
@@ -368,10 +344,6 @@ public class RenameService2 implements IRenameService2 {
 
 	protected IResourceServiceProvider.Registry getServiceProviderRegistry() {
 		return serviceProviderRegistry;
-	}
-
-	protected TokenUtil getTokenUtil() {
-		return tokenUtil;
 	}
 
 	protected IValueConverterService getValueConverterService() {
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2016, 2020 TypeFox GmbH (http://www.typefox.io) and others.
+ * Copyright (c) 2016, 2022 TypeFox GmbH (http://www.typefox.io) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -25,6 +25,9 @@ import org.eclipse.lsp4j.ReferenceParams;
 import org.eclipse.lsp4j.SymbolInformation;
 import org.eclipse.lsp4j.SymbolKind;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.eclipse.xtext.CrossReference;
+import org.eclipse.xtext.Keyword;
+import org.eclipse.xtext.RuleCall;
 import org.eclipse.xtext.findReferences.IReferenceFinder;
 import org.eclipse.xtext.findReferences.ReferenceAcceptor;
 import org.eclipse.xtext.findReferences.TargetURICollector;
@@ -56,6 +59,10 @@ import com.google.inject.Singleton;
 
 /**
  * @author kosyakov - Initial contribution and API
+ *
+ * Contributors: 
+ *   Rubén Porras Campo (Avaloq Evolution AG) - Do not return definitions for elements without identifiers.
+ *
  * @since 2.11
  */
 @Singleton
@@ -101,10 +108,11 @@ public class DocumentSymbolService implements IDocumentSymbolService {
 
 	public List<? extends Location> getDefinitions(XtextResource resource, int offset,
 			IReferenceFinder.IResourceAccess resourceAccess, CancelIndicator cancelIndicator) {
-		EObject element = eObjectAtOffsetHelper.resolveElementAt(resource, offset);
+		EObject element = eObjectAtOffsetHelper.getElementWithNameAt(resource, offset);
 		if (element == null) {
 			return Collections.emptyList();
 		}
+		
 		List<Location> locations = new ArrayList<>();
 		TargetURIs targetURIs = collectTargetURIs(element);
 		for (URI targetURI : targetURIs) {
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/EObjectAtOffsetHelper.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/EObjectAtOffsetHelper.java
@@ -12,6 +12,7 @@ import java.util.List;
 
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EReference;
+import org.eclipse.emf.ecore.EStructuralFeature;
 import org.eclipse.xtext.AbstractRule;
 import org.eclipse.xtext.CrossReference;
 import org.eclipse.xtext.EcoreUtil2;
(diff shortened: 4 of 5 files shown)


rubenporras:
Just let me know if you would like me to create a ticket for this PR. I am happy to do it, but I do not know if you agree with the PR or not, so I prefer to do that once  I have some feedback.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
```diff
@@ -119,6 +119,13 @@ public List<? extends Location> getDefinitions(XtextResource resource, int offse
 		return locations;
 	}
 
+	@Deprecated
+	public List<? extends Location> getDefinitions(XtextResource resource, int offset,
```
szarnekow, line 123: This is no longer invoked, is it?

rubenporras: That is right, it is no longer invoked. Since it is a public method, I set it as deprecated instead of removing it. It can also add //forRemoval or I could just delete it, if you prefer.

cdietrich: what i dont understand. why cant the old method simply be changed?

rubenporras: I did not change it because it is a public method and I do not know if someone relies on the old behaviour. That is why I leave the method in place with a deprecation, just in case. Maybe I am just to cautious :), if we do not need to be that careful, I would just remove the method or change it so that it does what we need, as you prefer.

cdietrich: but if it is not called subclasses might break anyway ?!?

rubenporras: I have the feeling we are not thinking about the same kind of breakage.
`DocumentSymbolService.getDefinitions(XtextResource, int, IResourceAccess, CancelIndicator)` is public (not protected), so it is not only about subclasses. One could call that method from somewhere else. I have decided to keep this method as is because of that.

Regarding subclasses, `getReferences` and `getDefinitions(Document, XtextResource, DefinitionParams, IResourceAccess, CancelIndicator)` are the methods called from `LanguageServerImpl`. The first one will behave as before, the second one will newly return links where there are cross-references. That is a change in behaviour but I believe it is the correct behaviour for the request `textDocument/definitions`. Since both methods used before `DocumentSymbolService.getDefinitions(XtextResource, int, IResourceAccess, CancelIndicator)` and that will return too much, I do not know how to avoid some kind of breakage here.

I feel it is best to remove usage of the conflicting method and deprecate it, but we could do something different. What would you if you where to correct `getDefinitions(Document, XtextResource, DefinitionParams, IResourceAccess, CancelIndicator)` to return only definitions where the symbol is actually a cross-refs?

cdietrich: i rather completely remove the old stuff
@szarnekow what do you think?

szarnekow: The API is question is clearly designed to be called from framework code instead of user-code. Please don't keep methods around as deprecated that are no longer working.

rubenporras: Thanks, I have done that and added a test case.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
```diff
@@ -96,12 +100,12 @@ public class DocumentSymbolService implements IDocumentSymbolService {
 	public List<? extends Location> getDefinitions(Document document, XtextResource resource, DefinitionParams params,
 			IReferenceFinder.IResourceAccess resourceAccess, CancelIndicator cancelIndicator) {
 		int offset = document.getOffSet(params.getPosition());
-		return getDefinitions(resource, offset, resourceAccess, cancelIndicator);
+		EObject element = eObjectAtOffsetHelper.resolveCrossReferencedElementAt(resource, offset);
+		return getDefinitions(resource, element, resourceAccess, cancelIndicator);
 	}
 
-	public List<? extends Location> getDefinitions(XtextResource resource, int offset,
+	private List<? extends Location> getDefinitions(XtextResource resource, EObject element,
```
szarnekow, line 107: Thank you for adding a test.

The more I look at this, the more I wonder how it should be if designed from scratch.

With this patch, we are using resolvedCrossReferenced when asking for definitions, but line 146 still uses the contained object at the cursor position. So asking for definitions will yield different results than asking for references when includeDefinition is set to true. I've doubts about it.

Also there appears to be quite some duplication at first or second glance.

I'd think it would make sense to extract the part that computes the targetURIs at a given location (offset) into an own method. Subsequently `definitions` and `references` would work with that collection. How does that sound to you?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
```diff
@@ -99,12 +113,43 @@ public List<? extends Location> getDefinitions(Document document, XtextResource
...
+			}
+		}
+		return null;
+	}
+
+	/**
+	 * @since 2.27
+	 */
+	protected boolean isIdentifier(ILeafNode leafNode) {
+		return (leafNode.getGrammarElement() instanceof RuleCall
```
cdietrich, line 141: will this work with name="keyword"?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
cdietrich: is there a list of usecases/examples for these methods?

rubenporras: I do not know about the methods, I just copied them. The examples of what I would like to achieve are added to the test cases, for example, get definition on a comment (test case 3) or a keyword (test case 4) does not return a definition.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
rubenporras: I am not sure what you have in mind. It works for the test case 4, where the cursor is at the beginning of "type", which is a keyword.

cdietrich: if i have in grammar something like

Variable: "var" name=XXX;
or 
VariableInSpecialContext returns Variable: "var" name="thereyougo";
will it work?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
cdietrich: how does this differ from what HyperlinkHelper does in ui?

rubenporras: HyperlinkHelper in ui only create links for cross-references, as in the version of this PR which submit first. Since Sebastian was skeptical about this I looked for some other condition which still meets the requirements.

cdietrich: @szarnekow how shall we continze here?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
rubenporras: I am not sure about what grammar element would that produce, but I think it would because 
a) it works in the test I added for the type
TypeDeclaration:
    'type' name=ID ...
;
and I think that is the same structure as the variable you sketched.
b) if it would not, we would have the same problem for the existing rename service (I guess we do not because no one complained so far)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
rubenporras: Hi @cdietrich , @szarnekow , can I help to make this PR move forward?

cdietrich: i am waiting for feedback from sebastian. we need to be a bit patient. anyway you should rebase as we have bumped the version to 2.28. this also needs changes to the sinces

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
rubenporras: I have found some time to move the code to where I would refactor it (EObjectAtOffsetHelper and a new class called NodeHelper). I still need to work on the other points.
Regards

rubenporras: Hi @szarnekow and @cdietrich,

I have now a version that will work as well with
`VariableInSpecialContext` returns Variable: "var" `name="thereyougo";`
but will still not let you rename it.

I am afraid I did not get your question about the offset being part of a data type rule, could you please explain that point?

Thanks

org.eclipse.xtext/src/org/eclipse/xtext/resource/EObjectAtOffsetHelper.java
```diff
@@ -157,4 +166,45 @@ protected EObject resolveCrossReferencedElement(INode node) {
 		return null;
 	}
 
+	private String getUnquotedToxenText(INode node) {
+		String text = NodeModelUtils.getTokenText(node);
+		if (text != null && text.startsWith("^")) {
+			return text.substring(1);
+		}
```
rubenporras, line 173: is there a better way to do this unquoting?

cdietrich: This is normally done by value converters

org.eclipse.xtext/src/org/eclipse/xtext/resource/EObjectAtOffsetHelper.java
```diff
@@ -157,4 +166,45 @@ protected EObject resolveCrossReferencedElement(INode node) {
...
+		if (offset >= 0 && xtextResource != null) {
+			IParseResult parseResult = xtextResource.getParseResult();
+			if (parseResult != null) {
+				ICompositeNode rootNode = parseResult.getRootNode();
+				if (rootNode != null) {
+					EObject object = resolveElementAt(xtextResource, offset);
+					String name = SimpleAttributeResolver.NAME_RESOLVER.apply(object);
+					if (name != null && !name.isEmpty()) {
+						ILeafNode leaf = NodeModelUtils.findLeafNodeAtOffset(rootNode, offset);
+						if (leaf != null && name.equals(getUnquotedToxenText(leaf)) && !tokenUtil.isWhitespaceOrCommentNode(leaf) && condition.test(leaf.getGrammarElement())) {
```
rubenporras, line 201: name.equals(getUnquotedToxenText(leaf)) is not really accurate, since it will macht any node with text equal to the name of the EObject. Actually I would like a condition that tests that the leaf node "leaf" has been used to set the attribute with name "name" ob the object, but I do not know how to code it. Thus I went for the approximation.

Any help with the condition to make it test that the leaf node has been used to assign the value of a given feature?

cdietrich: shoulnt ecoreutil2.findNodesForFeature or something the like help?

rubenporras: I have found that method in `NodeModelUtils` (Not ECoreUtil2), I will see if that can be used. Thanks.

rubenporras: @cdietrich , thanks, that worked.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
rubenporras: @szarnekow , I have checked by modifying the TestLanguage.xtext to include a data type rule and tested the change. The code I have now in the PR would work for data types rules.
I think I have addressed now all the points. 
Regards

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/rename/TestLanguageRenameService.java
```diff
@@ -56,4 +62,11 @@ protected EObject getElementWithIdentifierAt(XtextResource xtextResource, int of
 		}
 		return null;
 	}
+
+	protected boolean isIdentifier(ILeafNode leafNode) {
```
szarnekow, line 66: This is no longer used, is it?

org.eclipse.xtext/src/org/eclipse/xtext/resource/EObjectAtOffsetHelper.java
```diff
@@ -157,4 +165,43 @@ protected EObject resolveCrossReferencedElement(INode node) {
...
+	 * @param resource
+	 *            the {@link XtextResource} May be <code>null</code>.
+	 * @param offset
+	 *            the offset where the leaf node is to be searched.
+	 * @param condition
+	 *            a condition that the grammar element of the corresponding leaf node must met
+	 * @return the leaf node with an identifier at the given offset or <code>null</code>.
+	 * @since 2.27
+	 */
+	public EObject getElementWithNameAt(XtextResource resource, int offset, Predicate<EObject> condition) {
```
szarnekow, line 182: Please make the signature `Predicate<? super EObject>`

rubenporras: done

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/rename/TestLanguageRenameService.java
rubenporras: It was still used but I have removed the usage now, as we can reuse the implementation of `EObjectAtOffsetHelper#getElementWithNameAt`

org.eclipse.xtext/src/org/eclipse/xtext/resource/EObjectAtOffsetHelper.java
```diff
@@ -157,4 +165,43 @@ protected EObject resolveCrossReferencedElement(INode node) {
...
+			IParseResult parseResult = resource.getParseResult();
+			if (parseResult != null) {
+				ILeafNode leaf = NodeModelUtils.findLeafNodeAtOffset(parseResult.getRootNode(), offset);
+				if (!tokenUtil.isWhitespaceOrCommentNode(leaf) && condition.test(leaf.getGrammarElement())) {						
+					EObject crossRef = resolveCrossReferencedElementAt(resource, offset);
+					if (crossRef != null) {
+						return crossRef;
+					}
+					EObject definition = resolveElementAt(resource, offset);
+					final EStructuralFeature structuralFeature = definition.eClass().getEStructuralFeature("name");
```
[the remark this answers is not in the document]
rubenporras: Good catch. I knew about the API but forgot about it when doing the coding of this particular bit.

I did not follow what you meant with

> Is it possible that clients who want to filter for by name-region can use the locationInFileProvider or findNodesForFeature with the already known name attribute

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.java
```diff
@@ -190,27 +188,8 @@ protected EObject getElementAtOffset(XtextResource xtextResource, Document docum
...
-			}
-		}
-		return null;
-	}
-
-	protected boolean isIdentifier(ILeafNode leafNode) {
-		return (leafNode.getGrammarElement() instanceof RuleCall
-				|| leafNode.getGrammarElement() instanceof CrossReference)
-				&& !tokenUtil.isWhitespaceOrCommentNode(leafNode);
+		Predicate<EObject> renameable = r -> !(r instanceof Keyword);
```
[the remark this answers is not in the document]
rubenporras: I think renaming the element itself would work, as you say, so I have removed the predicate.

The idea behind was that the IDE should not allow renaming keywords (I usually take the JDT as the example to follow, and in the JDT, you do not have a rename option if your cursor is on the "public" keyword, for example), but probably this is not the right place to do that kind of filter and in addition, there are probably keywords which one might want to rename and others not.

rubenporras:
@szarnekow , would this PR be ready to go? I think I have adressed all comments.
Regards

rubenporras:
@cdietrich , shall we merge or do you want me to rebase and sqash the commit?

cdietrich:
@rubenporras would be nice if you can rebase and squash both commits into one

rubenporras:
@cdietrich , I have rebased and squashed the PR. Cheers.

cdietrich:
thx @rubenporras

PR #1910, opened 2022-06-15
[eclipse/xtext#2077] workaround for broken metadata in eclipse platform

cdietrich:
[eclipse/xtext#2077] workaround for broken metadata in eclipse platform


LorenzoBettini:
@cdietrich If I understand the problem correctly, this exclusion could be put in `xtext.generator` POM (actually gradle file, from which the POM is derived), similarly to what we did lately to pin platform dependency versions.

LorenzoBettini:
@cdietrich I meant to put the exclusion itself in the POM of `xtext.generator` (not pinning a version; I was talking about pinning to recall what we did in the past). If we know which dependency X of `xtext.generator` tries to pull the `org.osgi.service.prefs` we can exclude that in correspondence of the dependency X, just like you did in the workaround.

Something similar to what I did here https://github.com/LorenzoBettini/edelta/blob/master/edelta.parent/edelta.maven.plugin/pom.xml#L33

LorenzoBettini:
@cdietrich I'm expecting the exclusion to work at the BOM level: the idea of BOM and `<dependencyManagement>` is exactly to configure a dependency (and its exclusions) in one place only.

cdietrich:
Yes but gradle can’t do it as it seems

LorenzoBettini:
Another very good reason to ditch gradle for good.. ;)

cdietrich:
obsolete by removal of the bad artifact from maven central

PR #1911, opened 2022-06-15
[eclipse/xtext#2077] exclude bad non existing artifact pulled by platform

cdietrich:
[eclipse/xtext#2077] exclude bad non existing artifact pulled by platform


LorenzoBettini:
@cdietrich it depends on how the POM is generated but I guess that's what I meant.
However, yesterday they fixed the problem, so I guess you can simply revert this one and the other workarounds.

cdietrich:
No the new version does not help 
Next step platform tries it to get bad artifacts pulled

cdietrich:
obsolete by removal of the bad artifact from maven central

PR #1914, opened 2022-06-27
Improve code actions performance by loading the document on demand

rubenporras:
only if there are fix methods for the given diagnostic. This might saves
the resource load operation, which can be expensive.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
@@ -1,13 +1,14 @@
 /*******************************************************************************
- * Copyright (c) 2020 TypeFox GmbH (http://www.typefox.io) and others.
+ * Copyright (c) 2020, 2022 TypeFox GmbH (http://www.typefox.io) and others.
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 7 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
```diff
@@ -33,10 +37,22 @@ public interface IQuickFixProvider {
...
 	 * @param options
 	 *            Contextual action options
-	 * @param diagnostic
-	 *            the diagnostic
+	 * @param fixMethods
+	 *            a list of methods that create the resolution
 	 * @return 0..n resolutions for the given issue
 	 */
-	List<DiagnosticResolution> getResolutions(Options options, Diagnostic diagnostic);
+	List<DiagnosticResolution> getResolutions(Options options, List<Method> fixMethods);
```
szarnekow, line 44: The fact that this quickFixProvider is implemented based on reflection and annotated methods, should not become part of the interface. I'm afraid we'll need to find another solution to avoid the expensive part of the document parsing.

rubenporras:
Hi @szarnekow , I have posted a different PR that solves the same problem in a different way: https://github.com/eclipse/xtext-core/pull/1915, would you agree on that direction? if so I would cleanup the PR.
The new approach sets `ICodeActionService2.Options#document` and `ICodeActionService2.Options#resource` on demand

PR #1915, opened 2022-06-27
Improve code actions performance by loading the document on demand

rubenporras:
only if there are fix methods for the given diagnostic. This might saves
the resource load operation, which can be expensive.

### DIFF
--- a/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.java
+++ b/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.java
@@ -46,17 +46,20 @@ public class CodeActionService extends QuickFixCodeActionService {
 	private IChangeSerializer serializer;
 
 	@Override
-	public List<Either<Command, CodeAction>> getCodeActions(ICodeActionService2.Options options) {
-		List<Either<Command, CodeAction>> actions = super.getCodeActions(options);
-		for (Diagnostic d : options.getCodeActionParams().getContext().getDiagnostics()) {
-			Object code = d.getCode().get();
-			if (TestLanguageValidator.INVALID_NAME.equals(code)) {
-				actions.add(Either.forLeft(fixInvalidName(d, options)));
-			} else if (TestLanguageValidator.UNSORTED_MEMBERS.equals(code)) {
-				actions.add(Either.forRight(fixUnsortedMembers(d, options)));
-			}
+	protected List<Either<Command, CodeAction>> getCodeActions(Options options, Diagnostic diagnostic) {
+		Object code = diagnostic.getCode().get();
+		List<Either<Command, CodeAction>> codeActions = super.getCodeActions(options, diagnostic);
+		if (TestLanguageValidator.INVALID_NAME.equals(code)) {
+			codeActions.add(Either.forLeft(fixInvalidName(diagnostic, options)));
+		} else if (TestLanguageValidator.UNSORTED_MEMBERS.equals(code)) {
+			codeActions.add(Either.forRight(fixUnsortedMembers(diagnostic, options)));
 		}
-		return actions;
+		return codeActions;
+	}
+
+	@Override
+	protected boolean handlesDiagnostic(Diagnostic diagnostic) {
+		return true;
 	}
 
 	private Command fixInvalidName(Diagnostic d, ICodeActionService2.Options options) {
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
@@ -1,9 +1,9 @@
 /*******************************************************************************
- * Copyright (c) 2020 TypeFox GmbH (http://www.typefox.io) and others.
+ * Copyright (c) 2020, 2022 TypeFox GmbH (http://www.typefox.io) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
 package org.eclipse.xtext.ide.editor.quickfix;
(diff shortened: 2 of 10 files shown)


org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.java
```diff
@@ -50,10 +53,22 @@ public List<Either<Command, CodeAction>> getCodeActions(ICodeActionService2.Opti
 		List<Either<Command, CodeAction>> actions = super.getCodeActions(options);
 		for (Diagnostic d : options.getCodeActionParams().getContext().getDiagnostics()) {
 			Object code = d.getCode().get();
-			if (TestLanguageValidator.INVALID_NAME.equals(code)) {
-				actions.add(Either.forLeft(fixInvalidName(d, options)));
-			} else if (TestLanguageValidator.UNSORTED_MEMBERS.equals(code)) {
-				actions.add(Either.forRight(fixUnsortedMembers(d, options)));
+			try {
```
szarnekow, line 56: Can we leave this test-lang untouched to see whether the former way to write fixes still works?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
```diff
@@ -33,10 +36,22 @@ public interface IQuickFixProvider {
...
+	 * @param fixMethods
+	 *            a list of methods that create the resolution
 	 * @return 0..n resolutions for the given issue
 	 */
 	List<DiagnosticResolution> getResolutions(Options options, Diagnostic diagnostic);
 
+	
+	/**
+	 *
+	 * Collects all of methods that create the resolution for the given diagnostic
```
szarnekow, line 48: The JavaDoc doesn't match the implementation.

Is it possible to keep the original interface or make this a default method. Should be fine to always return true by default, shouldn't it?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
```diff
@@ -33,10 +36,22 @@ public interface IQuickFixProvider {
 	 *
 	 * @param options
 	 *            Contextual action options
-	 * @param diagnostic
-	 *            the diagnostic
+	 * @param fixMethods
```
szarnekow, line 39: JavaDoc mismatch.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ILanguageServerAccess.java
```diff
@@ -33,15 +33,15 @@
  */
 public interface ILanguageServerAccess {
 	class Context {
-		private final Resource resource;
+		private final XtextResource resource;
```
szarnekow, line 36: Why did you change this?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixCodeActionService.java
```diff
@@ -56,11 +58,23 @@ public List<Either<Command, CodeAction>> getCodeActions(Options options) {
...
+								options.setResource(context.getResource());
+
+								List<Either<Command, CodeAction>> codeActions = new ArrayList<>();
+								quickfixes.getResolutions(options, diagnostic).stream()
+										.sorted(Comparator
+												.nullsLast(Comparator.comparing(DiagnosticResolution::getLabel)))
+										.forEach(r -> codeActions.add(Either.forRight(createFix(r, diagnostic))));
+								return codeActions;
+							}).get());
+				} catch (InterruptedException | ExecutionException e) {
```
szarnekow, line 75: I understand why it was done like this, but having the blocking call here looks wrong to me.
The approach might need some more thought.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
rubenporras: done.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ILanguageServerAccess.java
rubenporras: because org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java had a variable declared with XtextResource, I have downgraded it to a Resource in the last version so that this is not needed.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/IQuickFixProvider.java
rubenporras: yes, it should be possible. done.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixCodeActionService.java
rubenporras: we could avoid the blocking call if we can expose the workspace manage in the ILanguageServerAccess. What about that?

rubenporras (commented):
Thanks for the review.

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/ide/server/CodeActionService.java
rubenporras: not quite because someone needs to create the Resource and the Document. I have submited a newer version where this is abstracted, but I still needed to change the signature of the method a little bit.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixCodeActionService.java
rubenporras: I have added a new method doSyncRead which fixes the problem, I hope you like the solution.

rubenporras:
@szarnekow , what do you think of the change now? there is no longer a blocking call and the test looks almost as before (one just need to override one method to return true indicating that any QF is handled)

szarnekow (approved):

rubenporras:
thanks, I will rebase and squash.

PR #1916, opened 2022-06-29
Storing a binary resource removes its URI from the SourceLevelURIs set

rubenporras:
by doing that we increase the performance of the
BuildContext#executeClustered because it will be possible to use binary
models from resources which has already been built and stored.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/build/BuildRequest.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/build/BuildRequest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2016 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2015, 2022 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -10,7 +10,9 @@ package org.eclipse.xtext.build;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.URI;
@@ -73,6 +75,8 @@ public class BuildRequest {
 	private List<URI> deletedFiles = new ArrayList<>();
 
 	private List<IResourceDescription.Delta> externalDeltas = new ArrayList<>();
+	
+	private Set<URI> sourceLevelURIs = new HashSet<>();
 
 	/**
 	 * Callback after validation, return <code>false</code> will stop the build.
@@ -195,5 +199,22 @@ public class BuildRequest {
 
 	public void setCancelIndicator(CancelIndicator cancelIndicator) {
 		this.cancelIndicator = cancelIndicator;
+	}
+
+	/**
+	 * @since 2.28
+	 */
+	public Set<URI> getSourceLevelUris() {
+		return sourceLevelURIs;		
+	}
+
+	/**
+	 * Sets a reference to the given set of source level URIs. Subsequent modifications of its content will be
+	 * observable here and on the call-site.
+	 * 
+	 * @since 2.28
+	 */
+	public void setSourceLevelUris(Set<URI> sourceLevelURIs) {
+		this.sourceLevelURIs = sourceLevelURIs ;		
 	}
 }
--- a/org.eclipse.xtext/src/org/eclipse/xtext/build/ClusteringStorageAwareResourceLoader.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/build/ClusteringStorageAwareResourceLoader.java
@@ -9,18 +9,14 @@
 package org.eclipse.xtext.build;
 
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Set;
 
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.xtext.resource.IResourceServiceProvider;
 import org.eclipse.xtext.resource.IResourceServiceProviderExtension;
 import org.eclipse.xtext.resource.XtextResourceSet;
-import org.eclipse.xtext.resource.persistence.SourceLevelURIsAdapter;
-import org.eclipse.xtext.resource.persistence.StorageAwareResource;
 import org.eclipse.xtext.xbase.lib.Functions.Function1;
 
 import com.google.common.collect.FluentIterable;
@@ -40,7 +36,6 @@ public class ClusteringStorageAwareResourceLoader {
 	 */
 	public <T> Iterable<T> executeClustered(Iterable<URI> uris, Function1<? super Resource, ? extends T> operation) {
 		int loadedURIsCount = 0;
-		Set<URI> sourceLevelURIs = new HashSet<>();
 		List<Resource> resources = new ArrayList<>();
 		List<T> result = new ArrayList<>();
 		Iterator<URI> iter = uris.iterator();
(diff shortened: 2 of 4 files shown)


rubenporras:
@szarnekow , I have incorporated the code refactoring. For the saving of the binary models in background thread, some complex changes are needed, so I will do that in a different PR. Thanks

org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
```diff
@@ -199,9 +205,16 @@ public URIBasedFileSystemAccess newFileSystemAccess(Resource resource, BuildRequ
 		private OperationCanceledManager operationCanceledManager;
 
 		protected void unloadResource(URI uri) {
+			unloadResource(uri, x -> true);
+		}
+
+		/**
+		 * @since 2.28
```
szarnekow, line 212: Suggested JavaDoc:

Unload the resource with the given uri if it satisfies the given condition.

org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
```diff
@@ -358,6 +370,76 @@ protected void generate(Resource resource, BuildRequest request, Source2Generate
...
+		/**
+		 * Deletes the binary models for the given set of URIs. Might be customized if
+		 * IResourceStorage supports deletion
+		 *
+		 * @param toBeDeleted
+		 *            set of URIs, must not be {@code null}
+		 * 
+		 * @since 2.28
+		 */
+		protected void deleteBinaryResources(final List<URI> toBeDeleted) {
```
szarnekow, line 415: Wouldn't this need a fileSystemAccess?

org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
```diff
@@ -358,6 +370,76 @@ protected void generate(Resource resource, BuildRequest request, Source2Generate
...
+		protected void storeBinaryResource(final Resource resource, IFileSystemAccessExtension3 fileSystemAccess) {
+			if (resource instanceof StorageAwareResource) {
+				IResourceStorageFacade resourceStorageFacade = ((StorageAwareResource) resource)
+						.getResourceStorageFacade();
+				if (resourceStorageFacade != null) {
+					resourceStorageFacade.saveResource((StorageAwareResource) resource, fileSystemAccess);
+					SourceLevelURIsAdapter adapter = SourceLevelURIsAdapter
+							.findInstalledAdapter(resource.getResourceSet());
+					if (adapter != null) {
+						adapter.getSourceLevelURIs().remove(resource.getURI());
```
szarnekow, line 437: As far as I can tell, adapter.getSourceLevelURIs returns an unmodifiable set. How can this work?

org.eclipse.xtext/src/org/eclipse/xtext/build/ClusteringStorageAwareResourceLoader.java
```diff
@@ -54,33 +47,14 @@ public <T> Iterable<T> executeClustered(Iterable<URI> uris, Function1<? super Re
...
 			resources.add(resourceSet.getResource(uri, true));
 		}
 		FluentIterable.from(resources).transform(operation::apply).copyInto(result);
 		return result;
 	}
 
 	/**
-	 * Return true if the given uri must be loaded from source.
-	 */
-	protected boolean isSource(URI uri) {
```
szarnekow, line 76: I wonder if we should keep this method here for backwards compatiblity reasons and delegate to it from the newly introduced isSource on the IncrementalBuilder

org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
rubenporras: Right, I did not see that. What do you think of lifting that restriction? I think it makes, so that the builder does not need to pass the original set around.
I have seen, that we have created our own adapter that does not have the restriction, exactly to adress this point.

org.eclipse.xtext/src/org/eclipse/xtext/build/ClusteringStorageAwareResourceLoader.java
rubenporras: I was also wondering as well if we should keep it. I have left the method as it was, just a @Deprecated tag. 
It would be possible to depend on the IncrementalBuilder, but one would need to create a new instance of it, call setContext on the new instance before one could call the method. It looks a bit too much for my taste, but I could do that if you prefer.

org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
rubenporras: done

org.eclipse.xtext/src/org/eclipse/xtext/build/IncrementalBuilder.java
rubenporras: Yes, that is right, one would need a fileSystemAccess. I have assumed that customizations could create one, I did not do it in your codebase because at the moment one would need to call deleteBinaryResources there is no such fileSystemAccess and creating one if the method does not do anything, would not be necessary.
For the time being, I have removed the method, I will come back to it later.

rubenporras:
@szarnekow , thanks for the review, I think I have adressed all the comments.

org.eclipse.xtext/src/org/eclipse/xtext/build/BuildRequest.java
```diff
@@ -196,4 +200,12 @@ public CancelIndicator getCancelIndicator() {
 	public void setCancelIndicator(CancelIndicator cancelIndicator) {
 		this.cancelIndicator = cancelIndicator;
 	}
+
+	public Set<URI> getSourceLevelUris() {
```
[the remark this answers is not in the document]
rubenporras: thanks, done.

org.eclipse.xtext/src/org/eclipse/xtext/resource/persistence/SourceLevelURIsAdapter.java
```diff
@@ -53,7 +53,7 @@ protected static SourceLevelURIsAdapter findOrCreateAdapter(ResourceSet resource
 	 */
 	public static void setSourceLevelUrisWithoutCopy(ResourceSet resourceSet, Set<URI> uris) {
```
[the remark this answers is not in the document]
rubenporras: I think so because UnmodifiableSet does not copy the data, it creates a view over it. The idea behing the change is to create the view once instead of each time the data is asked, a small performance optimization.

rubenporras:
@szarnekow , what do you think about this one? Would it be ready to merge? If you or Christian are unsure about the changes I did to SourceLevelURIsAdapter.java, I am happy to push them as a separate PR to make the process easier.

org.eclipse.xtext/src/org/eclipse/xtext/build/BuildRequest.java
```diff
@@ -196,4 +200,15 @@ public CancelIndicator getCancelIndicator() {
...
 	}
+
+	/**
+	 * @since 2.28
+	 */
+	public Set<URI> getSourceLevelUris() {
+		return sourceLevelURIs;		
+	}
+
+	public void setSourceLevelUris(Set<URI> sourceLevelURIs) {
```
szarnekow, line 211: Please add a comment that this will keep a reference to the given set and subsequent modifications of its content will be observable here and on the call-site.

szarnekow (approved):

org.eclipse.xtext/src/org/eclipse/xtext/build/BuildRequest.java
rubenporras: @szarnekow , done

org.eclipse.xtext/src/org/eclipse/xtext/build/ClusteringStorageAwareResourceLoader.java
szarnekow: I'm mainly concerned about code that *might* have overwritten this method assuming that it would be called. @cdietrich what's your take? Remove it completely to be sure that we *break* the client rather than silently offering an internal API that's no longer in use?

szarnekow (approved):

cdietrich:
do we already know if this will break github.com/xtext/xtext-gradle-plugin

rubenporras:
@cdietrich , I did not find any reference to ClusteringStorageAwareResourceLoader or to a method isSource, so I would think it should not. Is there something else that I could check in advance or can one merge and see if nightly test still work?

cdietrich:
i am not sure how the plugin populates its classpath. we will see

org.eclipse.xtext/src/org/eclipse/xtext/build/ClusteringStorageAwareResourceLoader.java
cdietrich: we can remove it with the next release

PR #1922, opened 2022-07-06
Catch stack overflows when loading from storage

rubenporras:
That is, catch and log StackOverflowErrors when loading resources from
storage, and fall back to recreating the resource from source.
StackOverflowErrors can happen if the AST is deeply nested, since the
algorithm to load from storage uses recursion.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/resource/persistence/StorageAwareResource.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/resource/persistence/StorageAwareResource.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014, 2020 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2014, 2022 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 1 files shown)


rubenporras:
This is an adapted version of https://github.com/dsldevkit/dsl-devkit/pull/613 for xtext-core.

cdietrich:
@rubenporras do you have an example for the stackoverflow

cdietrich:
so its DirectLinkingResourceStorageLoadable that is recursive?

rubenporras:
Hi @cdietrich,

The recursion is in org.eclipse.emf.ecore.resource.impl.BinaryResourceImpl.EObjectInputStream.loadResource(Resource), called through org.eclipse.xtext.resource.persistence.ResourceStorageLoadable.readContents(...).new EObjectInputStream() {...}.(InputStream, Map<?, ?>).

You can see that org.eclipse.emf.ecore.resource.impl.BinaryResourceImpl.EObjectInputStream.loadEObject() is called recursively, for example through loadEObject -> loadFeatureValue -> loadEObject or loadEObject -> loadFeatureMap -> loadFeatureMapEntry -> loadEObject.

If the user has a file with nested ASTs, and if they are persisted, one will get out of memory when recovering the model.

cdietrich:
hmmm but wont save have the same problem then?

rubenporras:
Hi @cdietrich ,

I have looked at the save implementation, and I think it has also a similar problem. I do not know why we never had problems saving the models, only recovering them. I guess that the loading happens at points where the stack is deeper than at the points where we save them.

Fixing the root problem would require avoiding this type of recursion in the algorithm, or to have a counter that aborts when recursion reaches some level, then one could set Xss accordingly. This PR does not attempt such fix, just a quick-win to make the system as resilient as possible and avoid crashing, given the current algorithm.

If you prefer not to have the safety net and wait for someone to fix it in a better way, I would understand it and just close the PR; we would then keep our "patched" version.

cdietrich:
lets wait what Sebastian thinks

org.eclipse.xtext/src/org/eclipse/xtext/resource/persistence/StorageAwareResource.java
```diff
@@ -59,6 +82,17 @@ public void load(Map<?, ?> options) throws IOException {
 					eAdapters.clear();
 				}
 				unload();
+			} catch (StackOverflowError e) {
+			      trimStackOverflowErrorStackTrace(e);
+			      LOG.warn("Failed to load " + uri + " from storage", e); //$NON-NLS-1$//$NON-NLS-2$
+
+			      if (contents != null) {
```
szarnekow, line 89: Nit: please extract a protected method `unload(Throwable t)`?  that's used from both catch blocks.

szarnekow: Maybe call it `loadFromStorageFailed(Throwable)`

rubenporras: @szarnekow , done

szarnekow (approved):

rubenporras:
@cdietrich , could we merge? Thanks.

PR #1928, opened 2022-07-13
Add protected method for subclasses

rubenporras:
to be able to access the openDocuments map and so be able to implement
its own didOpen/didClose methods.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.java
@@ -82,7 +82,7 @@ public class WorkspaceManager {
 
 	private final Map<String, ResourceDescriptionsData> fullIndex = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


rubenporras:
This is needed for our own implementation of https://github.com/eclipse/xtext/issues/2443

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.java
```diff
@@ -182,6 +182,15 @@ protected List<WorkspaceFolder> getWorkspaceFolders() {
 		return workspaceFolders;
 	}
 	
+	/**
+	 * @return the open document map, never {@code null}
+	 * 
+	 * @since 2.28
+	 */
+	protected Map<URI, Document> getOpenDocuments() {
```
szarnekow, line 190: You may want to extract a method `createOpenDocuments` that's used in the field initializer such that clients can provide a ConcurrentHashMap or a some other kind of map implementation according to their liking.

szarnekow (approved):

PR #1933, opened 2022-07-20
Fix document extensions

rubenporras:
Use the correct language specific DocumentExtentions depending on the URI of the
particular symbol.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
@@ -58,7 +58,9 @@ import com.google.inject.Singleton;
  * @author kosyakov - Initial contribution and API
  *
  * Contributors: 
- *   Rubén Porras Campo (Avaloq Evolution AG) - Do not return definitions for elements without identifiers.
+ *   Rubén Porras Campo (Avaloq Evolution AG) 
+ *     - Do not return definitions for elements without identifiers.
+ *     - Use the language specific DocumentExtentions depending on the URI of the particular symbol.
  *
  * @since 2.11
  */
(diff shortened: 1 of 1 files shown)


rubenporras:
For https://github.com/eclipse/xtext-core/issues/1934

rubenporras:
@szarnekow , any opinion on this one?

szarnekow:
> @szarnekow , any opinion on this one?

I've an opinion about many things :)

The idea looks correct to me. The overhead appears to be acceptable. Should we have a GlobalDocumentExtensions that encapsulates the dispatching to the language specific implementation?

rubenporras:
@szarnekow , I have checked the codebase of the LS, and I think this is the only class that uses `DocumentExtensions#newLocation` for locations which might be in a different resource. So I think that creating a new class is not needed.
Regards

szarnekow:
It would be great to have a test that exercises this code-path and relies on the fact that the document extensions are language specific. I know it's quite an effort to do that, but the infrastructure for that test might be useful in other cases, too.

rubenporras:
@szarnekow , how would you recommend to do such a test? I have been looking at it and I can imagine something like

1) Extend the testlanguage to have cross-refs to another language
2) In that language install a custom DocumentExtension that returns some special location for some special URI
3) Check for that special location given some model that triggers the condition.

That looks indeed like a loot of work. Or do you have other ideas?

I am not sure I would be able to do that before going on summer holidays (that is in three days), could we merge the fix and leave the issue open for the test? 

If not, in case you follow https://wiki.eclipse.org/Category:SimRel-2022-09 we would release Xtext 2.28 M1 without the fix, and that would be a pitty.

Regards

rubenporras:
@cdietrich , do you know how can I rerun the test? I logged in Jenkins but I have no such option available. Thanks

cdietrich:
i did rerun build

cdietrich:
@rubenporras   the eca validation seen your gmail adress somewhere in this commit. can you check?
@szarnekow how to continue here?

cdietrich:
ok rerunning the check helped @rubenporras

szarnekow (approved):
Approved though the concern about the lack of test coverage for multi-language LSP stuff remains. I do understand the motivation to have this fix available, though.

cdietrich:
created https://github.com/eclipse/xtext/issues/2421 as follow up

PR #1948, opened 2022-08-09
[eclipse/xtext#2087] switch from requireBundle to importPackage for junit5

cdietrich:
[eclipse/xtext#2087] switch from requireBundle to importPackage for junit5


org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TestProjectDescriptor.xtend
```diff
@@ -59,7 +59,7 @@ abstract class TestProjectDescriptor extends ProjectDescriptor {
 		}
 		if (config.junitVersion == JUnitVersion.JUNIT_5) {
 			deps += new ExternalDependency()=>[
-				p2.bundleId = "org.junit.jupiter.api"
+				p2.packages = #{"org.junit.jupiter.api"}
```
cdietrich, line 62: do we want to restrict version?

cdietrich: this also does not seem to work in tycho

PR #1952, opened 2022-08-17
Return a WorkspaceEdit even if an exception is thrown

rubenporras:
The error is logged but the method still returns a (null) value. This allows the server to return a CodeAction to the client, that this can inspect and discard (because no editAction is set). Otherwise the client will wait for the server and show it like that on the UI, which does not create a good user experience.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
@@ -8,8 +8,8 @@
  *******************************************************************************/
 package org.eclipse.xtext.ide.editor.quickfix;
 
+import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.URI;
-import org.eclipse.emf.common.util.WrappedException;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.resource.ResourceSet;
@@ -47,6 +47,9 @@ public class DiagnosticResolution {
 	private int relevance;
 
 	private Options options;
+
+	private static Logger log = Logger.getLogger(DiagnosticResolution.class);
+
 
 	@Deprecated//(forRemoval=true,since="2.27")
 	public DiagnosticResolution(String label, Factory modificationContextFactory,  IModification<EObject> modification) {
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -131,7 +134,8 @@ public WorkspaceEdit apply() {
 			}
 			return edit;
 		} catch (Exception exc) {
-			throw new WrappedException(exc);
+			log.error("Creation of WorkspaceEdit failed.", exc);
```
cdietrich, line 137: @heinrichWeichert are you ok with this change?

heinrichWeichert: Could you we return a null-Edit instead?

rubenporras: Yes, we could return `new WorkspaceEdit()`, LSP defines the workspaceEdit as a nullable field. I checked LSP4E and it always checks for the field to be != null before accessing it, I also checked what it would do with an empty edit, and it will have the same result, as it will contain no changes. I think returning null is simpler, but if you think returning a `new WorkspaceEdit()` is better, I would do that, I do not have a strong opinion. May I ask why you think that returning an empty Edit is preferable?

cdietrich: @szarnekow  what do you think?

heinrichWeichert: @rubenporras OK, i didn't check that. If it is indeed nullsafe an expected it's fine with me.

szarnekow: I've a preference towards returning null, too. An empty edit suggests that there is something the client *could* apply which might mess with its undo queue or whatsoever.

cdietrich (approved):

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
cdietrich: i am not sure if and how tracing works in standalone mode.

heinrichWeichert: all fine for me

cdietrich:
thx @rubenporras

Issue #1968, opened 2022-09-19
UriExtensions decodes URI twice


rubenporras:
If there are files in the LS Client with URIs like

_schema:/.../abc%23xyz.foobar_

the client will encode such URIs before sending the request to

_schema:/.../abc%2523xyz.foobar_

%25 is how % encoded.

The current version of UriExtensions will decode the URI twice as part in the `toEmfUri ` method, first to the original _schema:/.../abc%23xyz.foobar_ and then again to _schema:/.../abc#xyz.foobar_. That would already result in the wrong URI, but the URI might be handled to org.eclipse.xtext.ide.server.WorkspaceManager.read(URI), the `trimFragment` is called, stripping all characters beginning on the last '#'. As a consequence exceptions like the following are thrown:

```
java.lang.RuntimeException: Cannot create a resource for 'schema:/.../abc'; a registered resource factory is needed
	at org.eclipse.xtext.resource.XtextResourceSet.getResource(XtextResourceSet.java:263)
```

cdietrich:
can you give detail where the decoding is happening twice?

rubenporras:
Hi @cdietrich , see the PR I have just created. I would like to do a test for the PR but I will not get to do that until tomorrow.

cdietrich:
@szarnekow @rubenporras  is this fixed now?

rubenporras:
I think so, closing the issue

PR #1969, opened 2022-09-19
use ignoreEscaped=true to avoid decoding escaped characters twice

rubenporras:


rubenporras:
for #1968

rubenporras:
Hi @cdietrich, I have added a test. You can revert the change to uriExtensions and check that without, uriExtensions is does not return the uri that was encoded by URLEncoder, which I believe it must do.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.java
```diff
@@ -117,6 +119,13 @@ public void test_toUri_14() throws Exception {
 				uriExtensions.toUri("file:///c%3A/Users/dietrich/%D0%A4%D1%83%20%D0%91%D0%B0%D1%80.mydsl"));
 	}
 
+	@Test
+	public void test_toUri_15() throws Exception {
+		String uri = "something:/path/to/res%23ource";
+		String endocedUri = URLEncoder.encode(uri, StandardCharsets.UTF_8.toString());
```
szarnekow, line 125: typo in local variable name

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
```diff
@@ -34,7 +34,7 @@ public URI toUri(String stringUri) {
 	 */
 	public URI toEmfUri(java.net.URI netUri) {
 		String decoded = toDecodedString(netUri);
```
szarnekow, line 36: Is the method name `toDecodedString` a shameless lie?

szarnekow: No, it's not :)

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
```diff
@@ -34,7 +34,7 @@ public URI toUri(String stringUri) {
...
-		URI uri = URI.createURI(decoded, false);
+		URI uri = URI.createURI(decoded, true);
```
szarnekow, line 37: Can you help me understand the use case?
We receive a valid, encoded java.net.URI.
We decode it to something (which might now contain a % sign). Whether it contains it or not, is not relevant though - the string is supposed to be _decoded_.
Now we make an EMF URI from it: The decoded string must in any case be encoded. So `ignoreEscaped=true` appears to be wrong to me.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.java
```diff
@@ -117,6 +119,13 @@ public void test_toUri_14() throws Exception {
...
+		assertEquals(URI.createURI(uri), uriExtensions.toUri(endocedUri));
```
szarnekow, line 126: URI.createURI(uri) as used here does explicitly state 

> This method also does not perform encoding of invalid characters, so it should only be used when the URI string is known to have already been encoded, so as to avoid double encoding.

So if I mentally rename the local `uri` to `encodedURI` it becomes obvious that the test in fact tries to assert some double-encoding trickery. It escapes me why this would be a reasonable test case.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.java
rubenporras: fixed

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
rubenporras: Agree :)

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
rubenporras: @szarnekow: The use case is that you create a file on the client with something that contains as part of the name something that looks like an scaped character, for example "WyEx%23mpleFile.whatever".
When the client encodes it, it will transform it to "WyEx%2523mpleFile.whatever".
Unless you apply my fix, the uri on the server side, ends up being "WyEx#mpleFile.whatever" and not the original one, because we decode twice and not once (as I think we should).

I discover this problem because we have files with '#' in the name, and the decoding would think #... was the fragment part of an EMF URI. One of my first tries was to replace '#' with '%23' in the files names, and was surprised to still get '#' as a fragment. Looking at the code, it was due to the double decoding.

After looking at it later in more detail, I have discovered there is a another bug,  because in LSP '#' is not a fragment separator, just a normal character, so as part of this PR I have fixed both problems and created two different tests for them.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
```diff
@@ -34,7 +34,7 @@ public URI toUri(String stringUri) {
 	 */
```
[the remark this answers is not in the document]
rubenporras: I think it is supposed to be encoded, shall I add that to the method javadoc?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
```diff
@@ -34,7 +34,7 @@ public URI toUri(String stringUri) {
...
 	public URI toEmfUri(java.net.URI netUri) {
 		String decoded = toDecodedString(netUri);
-		URI uri = URI.createURI(decoded, false);
+		URI uri = URI.createURI(decoded, true, URI.FRAGMENT_NONE);
```
rubenporras, line 37: I am not sure if URI.FRAGMENT_NONE is correct for all uses cases, so maybe this override must be done in org.eclipse.xtext.ide.server.UriExtensions and not in this class. What do you think, @szarnekow ?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
szarnekow: Can you by chance add an integration test with a filename that contains a `#` char`

rubenporras: I have added one.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
szarnekow: The assumption that there's no fragment is probably not possible to guarantee. I'd rather expect a fragment to be allowed.

What I'm thinking: The netUri is already correctly encoded. Why do we decode and re-encode it at all?

rubenporras: I do not know either, but from testing, it looks like encoding of java.net.URI is different than encoding of an EMF URI.
netUri.toString() returns something like "something%3A%2Fpath%2Fto%2Fres%23ource" while uri.toString() returns "something:/path/to/res%23ource".

rubenporras: @szarnekow , regarding "The assumption that there's no fragment is probably not possible to guarantee", I think we can do that in `org.eclipse.xtext.ide.server.UriExtensions`, but not in `org.eclipse.xtext.util.UriExtensions`. So I have done that bit only in the `ide.server` package class. I think that is guaranteed because `#` has no special meaning in LSP, so no LSP client is going to send EMF fragments. Why should it?

I have updated the PR to reflect that.

rubenporras: I run the test test_toUri_16 and put a breakpoint in the debugger, then called toString() on both variables.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.java
```diff
@@ -117,6 +119,20 @@ public void test_toUri_14() throws Exception {
 				uriExtensions.toUri("file:///c%3A/Users/dietrich/%D0%A4%D1%83%20%D0%91%D0%B0%D1%80.mydsl"));
 	}
 
+	@Test
+	public void test_toUri_15() throws Exception {
+		String uri = "something:/path/to/res%20ource";
+		String encodedUri = URLEncoder.encode(uri, StandardCharsets.UTF_8.toString());
```
szarnekow, line 125: Sorry for the ping-pong here.

What is the *intent* of this test?
URLEncoder.encode(..) will encode the string as if it was a single segment of a URI.
Instead I'd assume that the intention was to have a URI with the *scheme* being `something` and three segments, right?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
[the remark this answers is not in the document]
szarnekow: Yes, that would be greatly appreciated.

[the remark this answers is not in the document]
szarnekow: Would this be accurate?
```java
	/**
	 * Converts the given absolute, encoded stringUri to an EMF URI.
	 * If the URI is a {@link URI#isFile()}, it is guaranteed that its {@link URI#authority()} is
	 * not null.
	 */
```

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
szarnekow: The tests seem to assert some weird contract.
The already encoded string is apparently supposed to be parsed to a URI that has *some* escape-sequences removed where *allowed* while other escape sequences are supposed to be retained where *necessary*.

szarnekow:
Having looked deeper into the existing test cases and the thereby defined semantics, I implemented a different approach:

https://github.com/eclipse/xtext-core/pull/1975

Newly added tests also revealed more weirdness that was fixed in that merge request.

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.java
rubenporras: The intent of the test is to show, that if there is resource where the name contains a sequence of characters that looks like an escaped sequence (the `%20`), it is handled properly, that is, the `%20` is not "unescaped". The current code has a bug here because it double-decodes. I could test this with LSP4E by creating a file with name "res%20ource" and seeing who Xtext current code takes this as "res ource" (the `%20` is replaced by ` `).

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/UriExtensionsTest.java
```diff
@@ -117,6 +119,20 @@ public void test_toUri_14() throws Exception {
...
+		assertEquals(URI.createURI(uri), uriExtensions.toUri(encodedUri));
+	}
+
+	@Test
+	public void test_toUri_16() throws Exception {
+		String uri = "something:/path/to/res#ource";
+		String encodedUri = URLEncoder.encode(uri, StandardCharsets.UTF_8.toString());
+		assertEquals(URI.createURI(uri, true, URI.FRAGMENT_NONE), uriExtensions.toUri(encodedUri));
```
rubenporras, line 133: the intent of this is to test that a file with `#`, when converted to the EMF URI on the server, does not end up being `res` with fragment `ource`.

cdietrich:
@rubenporras i assume this one can be closed as replaced by other PR correct?

rubenporras:
right.

Issue #1964, opened 2022-09-15
java.lang.IllegalArgumentException thrown when applying quickfix 


agonzalezgaliana:
Currently using Xtext version 2.28.0.

The language server shows a` java.lang.IllegalArgumentException` when executing a quickfix on a context with more than 1 error (more than one Diagnostic present). By taking a look on the code I think I've narrowed the problem to the `DiagnosticResolution `class where the following code seems to be incorrect:

`Diagnostic diagnostic = params.getContext().getDiagnostics().stream().collect(MoreCollectors.onlyElement());`

Apparently the current implementation only expects one diagnostic in the stream (`MoreCollectors.onlyElement()`) however the LSP standard allows a list of `Diagnostic` as part of the `CodeActionContext` object in the `CodeActionParams` ([see documentation ](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionContext))

Kind regards.

cdietrich:
cc @rubenporras

rubenporras:
I can take a look next week. As far as I recall, the pattern of just dealing with one diagnostic was present before my changes, but it might have ignored the rest instead of throwing the exception (I was not aware that `MoreCollectors.onlyElement()` would throw the exception).

cdietrich:
fixed in 2.29. thx for the fix @rubenporras

PR #1970, opened 2022-09-20
IllegalArgumentException thrown when applying quickfix

rubenporras:

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
@@ -13,7 +13,6 @@ import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.resource.ResourceSet;
-import org.eclipse.lsp4j.CodeActionParams;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.WorkspaceEdit;
 import org.eclipse.xtext.ide.editor.quickfix.DiagnosticModificationContext.Factory;
@@ -25,9 +24,8 @@ import org.eclipse.xtext.ide.server.TextEditAcceptor;
 import org.eclipse.xtext.ide.server.codeActions.ICodeActionService2.Options;
 import org.eclipse.xtext.ide.server.rename.ChangeConverter2;
 import org.eclipse.xtext.resource.EObjectAtOffsetHelper;
+import org.eclipse.xtext.resource.IResourceServiceProvider;
 import org.eclipse.xtext.resource.XtextResource;
-
-import com.google.common.collect.MoreCollectors;
 
 /**
  * @author Heinrich Weichert
@@ -46,7 +44,13 @@ public class DiagnosticResolution {
 
 	private int relevance;
 
-	private Options options;
+	private URI uri;
+	
+	private Document document;
+	
+	private ILanguageServerAccess access;
+
+	private Diagnostic diagnostic;
 
 	private static Logger log = Logger.getLogger(DiagnosticResolution.class);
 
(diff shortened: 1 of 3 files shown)


rubenporras:
for https://github.com/eclipse/xtext-core/issues/1964.
I will fix the line endings. I do not know why my eclipse installation keeps breaking every time and again...

cdietrich:
@rubenporras can you check eca

rubenporras:
@cdietrich , I do not know why github shows so many whitespace changes, the file looks good in my workspace. If that looks fine for you as well, the fix would be ready to merge.

cdietrich:
@rubenporras the code you have PRed seems to have windows endings and not linux.
i assume this is not what we want.

rubenporras:
By looking at other files in your project, it looks like you use windows delimiters, so maybe this was a file that was using wrongly the unix ones.
In any case, I have changed them for the diff to be clean.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -121,7 +119,7 @@ public WorkspaceEdit apply() {
 			EObject obj = helper.resolveContainedElementAt(tmpResource, offset);
 
 			WorkspaceEdit edit = new WorkspaceEdit();
-			Diagnostic diagnostic = params.getContext().getDiagnostics().stream().collect(MoreCollectors.onlyElement());
+			Diagnostic diagnostic = params.getContext().getDiagnostics().stream().findFirst().get();
```
cdietrich, line 122: wont this throw if there is no diagnostic? should we catch?

cdietrich:
we use in general unix ones, normally the git checkout/push process on windows might do back and forth conversion.
have no idea why this is not workiong for you

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
rubenporras: Yes, it will (same as MoreCollectors.onlyElement()). It fill be cached by the general handler some lines below. If you like I can handle it in a nicer way, even though I am not sure if it is possible to get that far without diagnostics.

cdietrich: ok

tivervac: I would indeed handle it a bit nicer. I would also place it a bit higher in the method for an early exit

```java
try {
List<Diagnostic> diagnostics = params.getContext().getDiagnostics();
if (diagnostics.isEmpty()) {
    log.warn("No diagnostics present for code action")
    return null;
}

Diagnostic diagnostic = Iterables.getFirst(params.getContext().getDiagnostics())
```

cdietrich: other question: do we need to pick / filter the "correct" diagnostic. am not sure how the diagostics are sorted and we the quickfix can handle a incorrect one.

rubenporras: We should not need to pick the "correct" diagnostic, that happened already before. It looks to me we just need to pick the diagnostic we chose already. 

I have pushed a new version of the PR to to that and have reorganized the DiagnosticResolution.apply() method a little bit to declare variables in the order they are used.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -139,7 +134,10 @@ public WorkspaceEdit apply() {
 		}
 	}
 
-	void configure(Options options) {
-		this.options = options;
+	void configure(Options options, Diagnostic diagnostic) {
+		this.resource = options.getResource();
```
szarnekow, line 138: Why did you decide to inline the options' properties here?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
```diff
@@ -108,20 +111,12 @@ public String getLabel() {
...
-			Document document = options.getDocument();
-			int offset = document.getOffSet(params.getRange().getStart());
-			EObjectAtOffsetHelper helper = new EObjectAtOffsetHelper();
-			ILanguageServerAccess access = options.getLanguageServerAccess();
 			ResourceSet resourceSet = access.newLiveScopeResourceSet(uri);
 			XtextResource tmpResource = (XtextResource) resourceSet.getResource(uri, true);
-			EObject obj = helper.resolveContainedElementAt(tmpResource, offset);
-
+			int offset = document.getOffSet(diagnostic.getRange().getStart());
+			EObject obj = new EObjectAtOffsetHelper().resolveContainedElementAt(tmpResource, offset);
```
szarnekow, line 118: I know this code was present before, but `new EObjectAtOffsetHelper` might cause trouble since language specific customizing wouldn't be supported. Can we read the helper from the tmpResource's services?

szarnekow (approved):

szarnekow:
I think the changes to the public methods are fine.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
rubenporras: Sure, I will update the PS tomorrow, or if this one is already merged, create a new PR for only this.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
rubenporras: I think passing the Options as part of the API in the configure is not good, it has some information which cannot be used as part of the diagnostic resolution (like the list of diagnostics that we where using before). I did not want to change the public API if not needed, so I decided to inline the parameters that are used, hopefully next time something must be inlined, some thinking is done about if it is correct to access the data at this place or not.

szarnekow: Agreed. Makes sense.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java
rubenporras: @szarnekow , I have done the change you suggested.

szarnekow (approved):

rubenporras:
@cdietrich , can we merge? Thanks

tivervac (approved):

cdietrich:
is there anything left?

tivervac:
> is there anything left?

I think we're fine to merge

rubenporras:
thanks for the reviews.

PR #1975, opened 2022-09-28
[#1968] Improved handling of encoded URIs in UriExtensions 

szarnekow:
The given encoded URI string was wrongly decoded / re-encoded in some
cases.

### DIFF
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
@@ -34,7 +34,7 @@ public class UriExtensions {
 	 */
 	public URI toEmfUri(java.net.URI netUri) {
 		String decoded = toDecodedString(netUri);
-		URI uri = URI.createURI(decoded, false);
+		URI uri = URI.createURI(decoded, true);
 		URI result = withEmptyAuthority(uri);
 		return result;
 	}
--- a/org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
+++ b/org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2017, 2019 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2017, 2022 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
```diff
@@ -8,35 +8,103 @@
...
+	 * is a {@link URI#isFile() file URI}, it is guaranteed that its
+	 * {@link URI#authority() authority} is not null.
 	 */
 	public URI toUri(String stringUri) {
-		try {
-			java.net.URI netUri = new java.net.URI(stringUri);
-			return toEmfUri(netUri);
-		} catch (URISyntaxException e) {
-			throw new RuntimeException(e);
+		URI result = URI.createURI(stringUri);
```
szarnekow, line 29: Turns out this approach also gives a nice performance boost for URIs that do in fact contain escape sequences.

Simple performance tests show 2 secs -> 0.4 secs = ~80% improvement

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
```diff
@@ -8,35 +8,103 @@
...
+	}
+
+	private URI cleanupEncodedCharacters(URI encoded) {
+		// Decompose the URI
+		String authority = encoded.authority();
+		String[] segments = encoded.segments();
+		String device = encoded.device();
+		String opaque = encoded.opaquePart();
+		String query = encoded.query();
+		String fragment = encoded.fragment();
```
cdietrich, line 44: do these include all parts of an uri like userinfo?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
```diff
@@ -8,35 +8,103 @@
...
+			opaque = URI.encodeOpaquePart(URI.decode(opaque), false);
+		}
+		if (query != null) {
+			query = URI.encodeQuery(URI.decode(query), false);
+		}
+		if (fragment != null) {
+			query = URI.encodeFragment(URI.decode(fragment), false);
+		}
+		// Depending on the shape of the URI recreate it from the decomposed parts which are now
+		// correctly encoded in a minimal (only the strictly necessary but no optional escape sequences) way
```
rubenporras, line 90: Is this the reason we re-encode the whole URI but the schema? Shall this comment rather be put as javadoc on the method?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
rubenporras: This is indeed nice.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
```diff
@@ -8,35 +8,103 @@
...
 	public URI toUri(String stringUri) {
-		try {
-			java.net.URI netUri = new java.net.URI(stringUri);
-			return toEmfUri(netUri);
-		} catch (URISyntaxException e) {
-			throw new RuntimeException(e);
+		URI result = URI.createURI(stringUri);
+		// Short-circuit the good case where the URI doesn't contain any unusual escape sequences
+		if (stringUri.contains("%")) {
+			return cleanupEncodedCharacters(result);
```
rubenporras, line 32: It would even be faster if we do not need to reencode the string :)

I still does not quite get why it is so. Has it anything to do with Cyrillic and other characters as such? I tested with URIs like the ones we have (e.g.` ctx:/src/AAA/MBA/MBA$MASS%23STEX.SCRIPT%20PACKAGE`) and the recoding does not do anything visible. It is exactly the same afterwards.

Maybe it is overdoing it, but since we only allow ASCII characters on our files URIs, could we put maybe `stringUri.contains("%")` in a protected method that we could override to just return false?

rubenporras (approved):
My tests are passing, but still I have tested this version locally, and it looks to fix the problems I wanted to address.

I still do not understand why the reencoding, and maybe it would be nice to be able to prevent it from happening for better performance (see my comments). In any case, since it works, for me this change would work out -> so approved from my side :)

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
szarnekow: I was under the impression, yes. I'll double-check.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
szarnekow: I'll make it a JavaDoc.

Yes, the cyrillic tests and the escape sequences in the device `c:` are the reason.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
szarnekow: I'll extract a protected method.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
szarnekow: User-Info is part of the authority. Should be fine.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/UriExtensions.java
szarnekow: Done.

cdietrich (approved):

rubenporras:
Thanks

Issue #1973, opened 2022-09-27
NPE following usage of  FoldingRangeService


rubenporras:
ep 27, 2022 3:04:49 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
SEVERE: Internal error: java.lang.NullPointerException
java.util.concurrent.CompletionException: java.lang.NullPointerException
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:346)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:704)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
	at org.eclipse.xtext.ide.server.concurrent.AbstractRequest.logAndCompleteExceptionally(AbstractRequest.java:73)
	at org.eclipse.xtext.ide.server.concurrent.ReadRequest.lambda$doRun$0(ReadRequest.java:69)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.NullPointerException
	at org.eclipse.xtext.ide.editor.folding.DefaultFoldingRangeProvider.computeObjectFolding(DefaultFoldingRangeProvider.java:67)
	at org.eclipse.xtext.ide.editor.folding.DefaultFoldingRangeProvider.getFoldingRanges(DefaultFoldingRangeProvider.java:60)
	at org.eclipse.xtext.ide.server.folding.FoldingRangeService.createFoldingRanges(FoldingRangeService.java:37)
	at org.eclipse.xtext.ide.server.LanguageServerImpl.lambda$foldingRange$50(LanguageServerImpl.java:1045)
	at org.eclipse.xtext.ide.server.WorkspaceManager.doRead(WorkspaceManager.java:460)
[...]

PR #1984, opened 2022-10-05
Fix NPE if document of resource is null

rubenporras:
For #1973 


szarnekow (approved):
Does it make sense to check earlier, e.g. in the doRead method itself?

rubenporras:
We already check in the doRead method itself, and call apply.work(null, null). Thus all implementations must check for null values with the current API. We could change the method doRead to accept two lambdas, one for the case where a resource/document is not null, and one for the case where those are null. Then the API would force checking and acting accordingly. I am happy to do that and refactor consumers if you prefer that, but it would be a breaking change unless I created a new method and leave the old one as deprecated.
Just tell me which version of the fix you prefer and I will work on it.

szarnekow (approved):
Reasonable. Thank you, @rubenporras

PR #1985, opened 2022-10-05
Build of persisted state to run in parallel with work on dirty state.

andrewL-avlq:
Enable a build of the persisted state to run in parallel to work performed on the dirty state by splitting the index, the resource set, the content provider, and the request manager into persisted and dirty components.

Introduces a limited scope dirty build so that validations are performed for dirty editors and the effect of changes in one dirty editor can be seen in other dirty editors.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
@@ -17,11 +17,13 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
 import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.common.util.WrappedException;
 import org.eclipse.emf.ecore.resource.ResourceSet;
 import org.eclipse.lsp4j.ClientCapabilities;
 import org.eclipse.lsp4j.CodeAction;
@@ -159,7 +161,10 @@ public class LanguageServerImpl implements LanguageServer, WorkspaceService, Tex
 	private static final Logger LOG = Logger.getLogger(LanguageServerImpl.class);
 
 	@Inject
-	private RequestManager requestManager;
+	private RequestManager persistedStateRequestManager;
+
+	@Inject
+	private RequestManager dirtyStateRequestManager;
 
 	@Inject
 	private WorkspaceSymbolService workspaceSymbolService;
@@ -223,16 +228,27 @@ public class LanguageServerImpl implements LanguageServer, WorkspaceService, Tex
 
 		result.setCapabilities(createServerCapabilities(params));
 		access.addBuildListener(this);
-		return requestManager.runWrite(() -> {
-			if (workspaceManager.isSupportsWorkspaceFolders()) {
-				List<WorkspaceFolder> workspaceFolders = params.getWorkspaceFolders();
-				if (workspaceFolders == null)
-					workspaceFolders = Collections.emptyList();
-				workspaceManager.initialize(workspaceFolders, this::publishDiagnostics, CancelIndicator.NullImpl);
-			} else {
-				workspaceManager.initialize(baseDir, this::publishDiagnostics, CancelIndicator.NullImpl);
+		// put initialisation job on both request queues so that all subsequent requests wait.
+		return dirtyStateRequestManager.runWrite(() -> {
+			try {
+				return persistedStateRequestManager.runWrite(() -> {
+					if (workspaceManager.isSupportsWorkspaceFolders()) {
+						List<WorkspaceFolder> workspaceFolders = params.getWorkspaceFolders();
+						if (workspaceFolders == null)
+							workspaceFolders = Collections.emptyList();
+						workspaceManager.initialize(workspaceFolders, this::publishDiagnostics,
+								CancelIndicator.NullImpl);
+					} else {
+						workspaceManager.initialize(baseDir, this::publishDiagnostics, CancelIndicator.NullImpl);
+					}
+					return result;
+				}, (cancelIndicator, it) -> it).get();
+			} catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+				throw new WrappedException("Interrupted while initialising the LanguageServer", e); //$NON-NLS-1$
+			} catch (ExecutionException e) {
+				throw new WrappedException("Could not initialise the LanguageServer", e); //$NON-NLS-1$
 			}
-			return result;
 		}, (cancelIndicator, it) -> it).thenApply(it -> initializeResult = it);
 	}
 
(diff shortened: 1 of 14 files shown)


szarnekow:
Hi @andrewL-avlq 

this is quite an impactful change that you prepared here. I'm afraid without a comprehensive set of tests, we'll have a hard time moving this forward.
From an initial reading of the code, I'd like to see tests that exercice the following sequence of events:

- Prepare a file with a validation issue
- Open that file in a document
- Change the content of the document such that the initial validations are fixed and new one are being introduced
- Close the document without saving
- Assert that the initial validation result is seen by the client again

In that area are multiple imaginable interactions, especially with multiple files that refer to each other and only a subset of these files is being opened in the editor. File changes and disk / in the document must be tested in various orders and permutations.

It'd be also interesting to make this a change that does not necessarily impact existing language implementations. Can we do it in such a way that all new functionality is only in new classes. Maybe a second default Guice module would allow to enable / disable the behavior.

What do you think?

andrewL-avlq:
Hi @szarnekow 
I'm happy to put together some tests for this. Are there existing tests with similar setups that I can use as a starting point?

Regarding putting the new functionality in separate classes: this is of course possible, but introduces duplication that would need to be managed somehow. Possible options include:

1. A complete split starting at a second dirty state aware LanguageServerImpl with no relation between implementations. This option obviously has the most duplication, but the least coupling between implementations.
2. The dirty state aware implementations derive from the existing implementations. This doesn't entirely eliminate duplication as many of the classes have not been written to be extended this way. Also, you end up with asymmetrically named methods (should the existing method be the dirty or the persisted version) and existing implementations deriving from the existing classes will not automatically be overriding the correct methods.
3. Abstract base classes with two implementation classes. This keeps duplication to a minimum, but you now have three classes where you used to have one, and the existing implementation has not been left untouched.

I am sure there are many other options - do you have a preference?

szarnekow:
Before following the path towards two modes (dirty-state yes/no), I think we should see how the tests evolve and how much we can rely on the feature to be (at least as) correct (as the current implementation).

szarnekow:
No, I'm not aware of test infra for that. I'd probably start from the tests for the validation and index reads.

cdietrich:
@andrewL-avlq  can you please rebase on origin/master so that build runs properly

andrewL-avlq:
Hi @szarnekow ,
I have added some basic tests that show opening, editing, saving, and closing in various orders.
Are there other scenarios you would like to see exercised?

andrewL-avlq:
Hi @szarnekow ,
do you have any further feedback for this PR?

cdietrich:
@szarnekow @andrewL-avlq  my time is running out. i need to start M3 latest today noon.

gerdleon:
Hi @szarnekow ,
@andrewL-avlq works in my team and I wonder if there is a possibility to move ahead with this change - or whether we should rather not pursue it further. I believe it is a general improvement worth sharing.

szarnekow:
Hi @gerdleon 

thank you for following up on this. Though I agree that proposed change is a general improvement, it does also impose significant effort on the reviewers side since it is far from a trivial change.

Let alone the simple removal of `@Singleton`: This is a change for the contract that cannot be done lightheartedly. It will break existing clients which rely on the fact that the RequestManager imposes a strict order on the submitted read and write tasks. I'm taking this as a random example from the proposed change set.

Please understand that the review for correctness alone is not something that I can do easily in my spare-time.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
```diff
@@ -159,7 +161,10 @@ public class LanguageServerImpl implements LanguageServer, WorkspaceService, Tex
 	private static final Logger LOG = Logger.getLogger(LanguageServerImpl.class);
 
 	@Inject
-	private RequestManager requestManager;
+	private RequestManager persistedStateRequestManager;
```
szarnekow, line 164: Please note: There must be exactly one request manager. It is basically the synchronization primitive as used by the Xtext LSP server. Having two breaks the entire approach taken towards read / write access

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ProjectManager.java
```diff
@@ -59,32 +60,49 @@ public class ProjectManager {
...
+	private Provider<Map<String, ResourceDescriptionsData>> persistedIndexProvider;
+
+	private Provider<Map<String, ResourceDescriptionsData>> dirtyIndexProvider;
 
 	private IExternalContentSupport.IExternalContentProvider openedDocumentsContentProvider;
 
-	private XtextResourceSet resourceSet;
+	private IExternalContentSupport.IExternalContentProvider dirtyDocumentsContentProvider;
+
+	private XtextResourceSet persistedResourceSet;
```
szarnekow, line 79: Two resource-sets: Does this mean that the entire project will be in memory twice?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ProjectManager.java
```diff
@@ -59,32 +60,49 @@ public class ProjectManager {
...
 
-	private Provider<Map<String, ResourceDescriptionsData>> indexProvider;
+	private Provider<Map<String, ResourceDescriptionsData>> persistedIndexProvider;
+
+	private Provider<Map<String, ResourceDescriptionsData>> dirtyIndexProvider;
 
 	private IExternalContentSupport.IExternalContentProvider openedDocumentsContentProvider;
 
-	private XtextResourceSet resourceSet;
+	private IExternalContentSupport.IExternalContentProvider dirtyDocumentsContentProvider;
```
szarnekow, line 77: What is the openDocument content vs the dirtyDocumentContent ?

gerdleon:
> Hi @gerdleon
> 
> thank you for following up on this. Though I agree that proposed change is a general improvement, it does also impose significant effort on the reviewers side since it is far from a trivial change.
> 
> Let alone the simple removal of `@Singleton`: This is a change for the contract that cannot be done lightheartedly. It will break existing clients which rely on the fact that the RequestManager imposes a strict order on the submitted read and write tasks. I'm taking this as a random example from the proposed change set.
> 
> Please understand that the review for correctness alone is not something that I can do easily in my spare-time.

Hi @szarnekow ,
I understand and thank you for your efforts!

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
andrewL-avlq: The read/write semantics implemented by the request manager really only makes sense for a single index. Once you introduce a second index, then you do not want reads and writes to the first to be blocked by writes to the second (for example), but the desired behavior for requests to one of the indexes is the same as the existing request manager. There are other ways the requests could be managed (one manager with multiple queues for instance), but the simplest approach was just to have two request managers. I am happy to refactor if you think this should be handled differently.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ProjectManager.java
andrewL-avlq: It means that the resources loaded for dirty editors are loaded into a different resource set than the one for building the main index. So only as much of the project as you have loaded to support your dirty editors are in memory twice. However, worst case that could be the entire project, yes.

cdietrich:

PR #1989, opened 2022-10-10
[eclipse/xtext#1982] support for java 17

cdietrich:
[eclipse/xtext#1982] support for java 17


org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/Test.java
```diff
@@ -0,0 +1,11 @@
+package org.eclipse.xtext.testing;
+
+public class Test {
```
szarnekow, line 3: Who are you?

szarnekow (approved):

PR #1990, opened 2022-10-13
Simplify the DefaultFoldingRangeProvider

rubenporras:
Simplify the DefaultFoldingRangeProvider, since LSP does not support the properties initiallyFolded and visualPlaceholderRegion there is no need to calculate them.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeAcceptor.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/folding/DefaultFoldingRangeAcceptor.java
@@ -11,8 +11,6 @@ package org.eclipse.xtext.ide.editor.folding;
 
 import java.util.Collection;
 
-import org.eclipse.xtext.util.ITextRegion;
-
 /**
  * @author Michael Clay - Initial contribution and API
  * @author Sebastian Zarnekow - Introduced FoldedPosition
@@ -21,7 +19,7 @@ import org.eclipse.xtext.util.ITextRegion;
  * @since 2.26
  */
 public class DefaultFoldingRangeAcceptor implements IFoldingRangeAcceptor {
-	private Collection<FoldingRange> result;
+	private final Collection<FoldingRange> result;
 
 	public DefaultFoldingRangeAcceptor(Collection<FoldingRange> result) {
 		this.result = result;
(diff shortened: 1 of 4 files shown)


rubenporras:
@szarnekow , I realize this is a breaking change because it removes existing APIs, but since I do not understand why we have APIs that we do not use (I could not find any usage) and that produce data that is not specified in LSP, yet this computations consume resources on each request. Thus I decided to create a PR to remove the overhead and simplify the code and ask your opinion about it.
Regards

cdietrich:
@msujew would that be an issue for you?

msujew (approved):
Initially, I had the idea to adapt the already existing folding provider from `xtext.ui` to also use this new folding range provider. However, I never had the opportunity to get back to it.

If we ignore this idea, I'm also in favor of simplifying the LSP folding provider. The changes look good to me, so this gets a thumbs-up from me 👍

szarnekow:
To me this would be a step backwards. The API was designed such that we can single source the folding computation for LSP and Eclipse by exposing the superset of "UI" features. Modifying this here will require clients to have two implementations again if they want to provide a more suitable experience in Eclipse.

rubenporras:
I understand one would like to have common code for LSP and Eclipse UI, so I left the abstractions over LSP4J in place. 

However, I do not think it is necessary to have on the IDE plugin APIs which are specific for the Eclipse UI, and it is actually not nice to have computations for data that will never be used (like the visual region). I think if the API and the computation is needed for the Eclipse UI, the interface as well as the default implementation could be extended in the Eclipse UI plugin, is this not how the layers are designed?

cdietrich:
@rubenporras  yes but this was never contributed.

rubenporras:
Hi @cdietrich , I am not sure I understand your comment. Do you mean that I could go on with the refactor in this PR if I add the deleted functionality to the Xtext UI plugin in another PR, so that both can be merge at the same time?

cdietrich:
no, i just answered your question about "why is the ide class not used in eclipse"

cdietrich:
@szarnekow @rubenporras  my time is running out. i need to start M3 latest today noon.

rubenporras:
I think @szarnekow does not agree at the time being that this PR goes into the right direction, for the time being, I have converted it into a draft. If you like me, I can close it.

szarnekow:
I do like the goal of this PR: faster conversion to position. However I'm not sure its the best idea to perform these optimizations on all services one by one. @cdietrich idea to move the API into the document doesn't sound too bad to me.

cdietrich:

PR #1996, opened 2022-10-31
Replace quadratic algorithm in FoldingRangeService by one that scales up

rubenporras:
nicely. That reduces the time to compute foldings for big documents with lots of folding regions from 3 minutes to less than a second in some of our manual tests.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
@@ -8,12 +8,20 @@
  *******************************************************************************/
 package org.eclipse.xtext.ide.server.folding;
 
+import java.io.IOException;
+import java.io.LineNumberReader;
+import java.io.StringReader;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.SortedSet;
 import java.util.stream.Collectors;
 
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.emf.common.util.WrappedException;
 import org.eclipse.lsp4j.FoldingRange;
-import org.eclipse.lsp4j.Position;
 import org.eclipse.xtext.ide.editor.folding.IFoldingRangeProvider;
 import org.eclipse.xtext.ide.server.Document;
 import org.eclipse.xtext.resource.XtextResource;
@@ -24,6 +32,7 @@ import com.google.inject.Singleton;
 
 /**
  * @author Mark Sujew - Initial contribution and API
+ * 		   Rubén Porras Campo - Improve complexity of offset mapping algorithm
  * 
  * @since 2.26
  */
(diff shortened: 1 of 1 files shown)


rubenporras:
@szarnekow , I am aware I am breaking the API of this class but I do not think we can fix the algorithm if we keep the existing API.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
```diff
@@ -38,18 +47,56 @@ public List<FoldingRange> createFoldingRanges(Document document, XtextResource r
 		if (resource == null || document == null) {
 			return Collections.emptyList();
 		}
-		return foldingRangeProvider.getFoldingRanges(resource, cancelIndicator).stream()
-				.map(range -> toFoldingRange(document, range)).filter(range -> isValidRange(range))
-				.collect(Collectors.toList());
+		SortedSet<org.eclipse.xtext.ide.editor.folding.FoldingRange> foldingRanges = foldingRangeProvider
```
szarnekow, line 50: About backwards compatiblity:
Assuming the optimization would not return a Map<Integer, Integer> but wrap the FoldingRange with a FoldingRangeAndLineInformation. This could leave the APIs as they are and still perform way better than the old implementation. @cdietrich worth the dance?

cdietrich: i still dont get the problem in the first place. maybe too early in the morning.
where is the quadratic and where is the api change?

cdietrich: the offset calculation?

cdietrich: so an alternative would be to have a list of all newlines and used that one to calculate.
dont we do this somewhere already

cdietrich: maybe we can add an api for this to document.

cdietrich: unfortunately i cannot find this code i have the dejavu about right now.

szarnekow: The line-endings can be obtained from the node model and NodeModelUtils.getLineAndColum(..) will usually be quite fast.

cdietrich:
@szarnekow @rubenporras  my time is running out. i need to start M3 latest today noon.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/folding/FoldingRangeService.java
rubenporras: @cdietrich , yes, the offset calculation is the problem. The current code iterates through the characters from the beginning of the document to the point in the document where the offset is to be translated. If you have hundreds of offsets and long documents, then starting again and again from scratch leads to very slow responses.

That is not very good, but the biggest problem is that because the cancel indicator is not used to break the iteration while translating offsets, when a write requests comes and ask for all existing read requests to be cancelled, it will need to wait for this very long request to finish. Only then the write request can start. This is how we noticed it, because we show that when editing, markers which should appear took minutes to appear (basically the time that the build was waiting for the folding range to be computed).

Since the folding ranges are always computed, the problem makes the LS unusable for long files with many folding regions.

rubenporras: Yes, maybe we can code the document in such a way that calling getPosition will try to reuse the previous offset if it was smaller than the previous one. Now that we have this PositionReader we have merge, that should be easy. I will give it a try now and update the PR if doable.

rubenporras: @szarnekow , I am not convinced we should use `NodeModelUtils.getLineAndColum(..)`, that will be slower than the current proposal. One would need to fetch the nearest EObject for the begin and end offset of the folding ranges, and from there go to the node model again and ask for the line and column. The code should also handle the case where the folding range does not start at the beginning of such EObject, but on some property of it.

It also looks more complex than the proposed fix, that is just sorting the offsets and then doing one pass on the document to translate them.

In any case, I see M3 is out, so we have now some days again to look into the topic ;)

szarnekow: >  will be slower than the current proposal.

Yes, it will. Probably not measurable though.

> One would need to fetch the nearest EObject for the begin and end offset of the folding ranges, and from there go to the node model again and ask for the line and column. The code should also handle the case where the folding range does not start at the beginning of such EObject, but on some property of it.

I don't follow. getLineAndColumn takes an offset and returns the line and column, independently from which node you pass. No need to fuzz with EObjects as far as I can tell.

rubenporras: I understand now, I have not realized that `NodeModelUtils.getLineAndColum(..)` can take any node, so we can always past the root node. Still, I would not know how to do that in `org.eclipse.xtext.ide.server.folding.FoldingRangeService.toFoldingRange(Document, FoldingRange)` we do not have the `XtextResource`, only the `Document`, and I think we cannot get the `XtextResource` from a `Document`, can we?

Assuming that we cannot, to keep the current API, we could wrap the FoldingRange with a FoldingRangeAndLineInformation, as you suggested. We could also follow on https://github.com/eclipse/xtext-core/pull/2012, I think that we create a new API in `Document` that allows us to reset the `PositionReader` to the state of the offset of the start of the FoldingRange, that should also work nicely, and having a method `Document#getPosition()` that works nicely outside of the box for successive calls on increasing offsets is nice to have.

I do not mind any of the options as long as they perform nicely, so which one do you prefer?

rubenporras: @szarnekow , any answer to question above?

rubenporras: Okay, I will look into that.

cdietrich:

PR #1998, opened 2022-11-03
Set completionList only as incomplete if the request has been cancell…

rubenporras:
…ed or if the proposal acceptor could not accept all the proposals because the limit of proposals to accept has been reached.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/IIdeContentProposalAcceptor.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/IIdeContentProposalAcceptor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2020 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2015, 2022 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -19,5 +19,16 @@ public interface IIdeContentProposalAcceptor {
 	void accept(ContentAssistEntry entry, int priority);
 	
 	boolean canAcceptMoreProposals();
+	
+	/**
+	 * If the acceptor accepted all proposals.
+	 * 
+	 * @since 2.29
+	 */
+	default boolean didAcceptAllProposals() {
+		// this default implementation will be accurate for all cases but the one
+		// where the acceptor accepted all proposals so far but should reject another one
+		return canAcceptMoreProposals();
+	}
 	
 }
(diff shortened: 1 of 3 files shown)


rubenporras:
for https://github.com/eclipse/xtext-core/issues/1999

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/IdeContentProposalProvider.java
```diff
@@ -61,16 +61,18 @@ public class IdeContentProposalProvider {
 
 	/**
 	 * Create content assist proposals and pass them to the given acceptor.
+	 * @return true if the acceptor accepted all the proposals, false otherwise.
 	 */
-	public void createProposals(Collection<ContentAssistContext> contexts, IIdeContentProposalAcceptor acceptor) {
```
szarnekow, line 65: This is not binary compatible. Do you strictly need the boolean return value or would it be possible to communicate this differently?

rubenporras: I did not think about binary compatibility. 

One alternative I can think about would be to add a method `didAcceptedAllProposals` to `IIdeContentProposalAcceptor`. That would not break binary compatibility as far as I understand. If we also add a default implementation, it should not break source compatibility either, the problem three is that I would not know if we should return false or true.

So to keep both probably the cleanest would be to add a new interface `IIdeContentProposalAcceptor2` that inherit from `IIdeContentProposalAcceptor` and adds a `didAcceptedAllProposals`. Should I go for that one?

rubenporras:
@szarnekow , any more comments on this one? I think it would be ready to merge,

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/IIdeContentProposalAcceptor.java
```diff
@@ -20,4 +20,15 @@ public interface IIdeContentProposalAcceptor {
 	
 	boolean canAcceptMoreProposals();
 	
+	/**
+	 * If the acceptor accepted all proposals.
+	 * 
+	 * @since 2.29
+	 */
+	default boolean couldAcceptAllProposals() {
+		// this default implentation will be accurate for all cases but the one
```
szarnekow, line 29: Typo in `implentation`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/IIdeContentProposalAcceptor.java
```diff
@@ -20,4 +20,15 @@ public interface IIdeContentProposalAcceptor {
 	
 	boolean canAcceptMoreProposals();
 	
+	/**
+	 * If the acceptor accepted all proposals.
+	 * 
+	 * @since 2.29
+	 */
+	default boolean couldAcceptAllProposals() {
```
szarnekow, line 28: Maybe call it `didAcceptAllProposals`

szarnekow (approved):
Two small things that I'd like to ask you to fix.

szarnekow (approved):

PR #2002, opened 2022-11-09
Rename private variable to follow public method name

rubenporras:


rubenporras:
I am sorry, I forgot to rename the private field accordingly. It is not very important since it is private and can always be refactored, but I think it is better to do it right away

rubenporras:
this PR would fix the build as well

rubenporras:
@cdietrich , could we merge this to fix the build?

PR #2012, opened 2022-11-14
Replace quadratic algorithm in FoldingRangeService by one that scales up

rubenporras:
more nicely.

That reduces the time to compute foldings for big documents with lots of folding regions from 3 minutes to less than a second in some of our manual tests.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.java
@@ -8,11 +8,15 @@
  */
 package org.eclipse.xtext.ide.server;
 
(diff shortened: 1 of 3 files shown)


rubenporras:
@cdietrich , this is an alternative for https://github.com/eclipse/xtext-core/pull/1996 which I hope you like more than the original PR. I think that it is nicer that my first attempt.

rubenporras:
I have noticed that even though this will work better than current code, it is still not as good as https://github.com/eclipse/xtext-core/pull/1996, because folding ranges usually are nested, thus it is not possible to just move forward on the input stream while translating folding ranges, the start offset of a nested folding range is smaller than the end offset of the parent folding range.

So closing it, it is not good enough.

PR #2015, opened 2022-11-16
Improve SemanticTokensFullConfiguration to avoid text comparison

rubenporras:
also, include a small fix to IndexOnlyProjectTest to avoid a "possible NPE" marker given by the JDT Null Pointer Analysis.


szarnekow:
> @szarnekow something that can or should still go into 2.29?

I think so, yes

PR #2021, opened 2022-11-18
Pin also emf.ecore

LorenzoBettini:
See eclipse/xtext#2397
This is the same but for org.eclipse.xtext

Signed-off-by: Lorenzo Bettini lorenzo.bettini@gmail.com


cdietrich (approved):

cdietrich:
Thx

Issue #2029, opened 2022-11-23
Wizard/Generator old testing plugins


cdietrich:
Wizard/Generator should not create dependencies to xtext.junit4 / xbase.junit

cdietrich:
done

PR #2030, opened 2022-11-23
[#2029] remove old test plugins from wizard, add new ones

cdietrich:
[#2029] remove old test plugins from wizard, add new ones


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/JUnitFragment.xtend
```diff
@@ -72,7 +73,9 @@ class JUnitFragment extends AbstractStubGeneratingFragment {
 	protected def String getXbaseUiTestingPackage() {
```
cdietrich, line 73: do dont see any way how to make this transition nicely. the fragment is very wird in the test plugin regards

Issue #2052, opened 2023-01-13
Switch to Maven/Tycho (EXPERIMENT)


LorenzoBettini:
As said here https://github.com/eclipse/xtext/issues/2133 I'd like to try to switch to Maven/Tycho

LorenzoBettini:
I started working on that in my fork, this branch https://github.com/LorenzoBettini/xtext-core/tree/lb_2052_maven_tycho

For the moment org.eclipse.xtext.util and org.eclipse.xtext are compiling!

I also set up a GitHub Actions workflow since I'll deal with Jenkins later.
In case, later, we can remove the GitHub Actions workflow.

LorenzoBettini:
This is the build log https://github.com/LorenzoBettini/xtext-core/actions

LorenzoBettini:
@cdietrich in xtext.ide the MANIFEST wants org.eclipse.lsp4j with at least 0.19. I can't find a p2 repo providing that. Do you know where I can find it?

cdietrich:
Please check Xtext.setup it contains the url

LorenzoBettini:
Right! Thanks! I'm currently not using Xtext.setup and haven't thought about that.

LorenzoBettini:
`org.eclipse.xtext.testing` builds and tests run!

LorenzoBettini:
@cdietrich should the Xtend files be processed during the build? If so, I guess I should use the latest released version of the `xtend-maven-plugin`, because it's not in this repo, right

LorenzoBettini:
`org.eclipse.xtext.xtext.generator` and `wizard` build!

cdietrich:
regarding xtend: yes we do the same with the gradle build.
in xtext-eclipse and xtext-xtend we had a https://github.com/eclipse/xtext-xtend/blob/8521139f5652be55444114de850af23aee37ab47/releng/org.eclipse.xtend.tycho.parent/pom.xml#L13 property

LorenzoBettini:
`xtext.ide.tests` run and pass in the Maven build! :)

LorenzoBettini:
Build and tests are performed also on macOS and Windows

LorenzoBettini:
In the current maven build script I see

```
-Dmaven.repo.local=.m2/repository
```

I guess it's useless since the Jenkins workspace is ephemeral anyway, so I removed that.

It's installing in

```
/home/jenkins/.m2/repository/org/eclipse/xtext/
```

[...]

PR #2053, opened 2023-01-14
[eclipse/xtext#2134] remove dependency to obsolte jdt.compiler.(tool|apt)

cdietrich:
[eclipse/xtext#2134] remove dependency to obsolte jdt.compiler.(tool|apt)


cdietrich:
@LorenzoBettini can i merge this now to have a green build again?

LorenzoBettini:
@cdietrich I still haven't got time to test it personally, but I think we can merge it. The fragments were needed because there were no other ways to force them in the TP (when jdt.core was also present). Now that the fragments are gone, we should be safe without them (as long as they are also removed from the dependencies of the maven plugin `xtext-maven-plugin`).

cdietrich:
@LorenzoBettini the point is that they are not, cause we are still on old platform from a maven perspective
thus i wanted you to have a look

LorenzoBettini:
@cdietrich OK, I see. Then I'll try to do that ASAP. I'm actively working on https://github.com/eclipse/xtext-core/issues/2052 for the moment.

I guess it won't be a problem anyway, because we won't have a mixture of apt and tool from the TP, because from the TP point of view those fragments are not there anymore.

I guess that, from the xtext-maven-plugin point of view:
- if jdt.core is taken from the more recent TP and apt from the old Maven dep, apt will not simply be used
- if jdt.core is taken from the old Maven dep it will also find the old apt, so it should be fine

cdietrich:
gonna merge for now, if more adjustment needed we need a second pr anyway

PR #2054, opened 2023-01-16
Move to Maven/Tycho for projects and build infrastructure

LorenzoBettini:
See also #2052 

Of course, it's still WIP. However, it builds and tests are green (locally and in the CI)

Gradle is not used at all. Just Maven/Tycho. All projects are Eclipse projects (besides dependencies specifications in the POM for those that have to be deployed as Maven artifacts, besides p2 artifacts).

I still haven't updated the Oomph setup, so for the moment, we have to use a manual procedure to use Eclipse with the converted Maven projects.

- Download a distribution Eclipse DSL "Eclipse IDE for Java and DSL Developers"
- Add the Xtext milestones update site (https://download.eclipse.org/modeling/tmf/xtext/updates/milestones/) and update Eclipse
- MAKE SURE Buildship is NOT installed (in case, uninstall it): I still haven't removed any Gradle build files (though I removed the Gradle nature from all the projects; however, it's better not to have Gradle to avoid possible interferences)
- Disable Maven -> "Automatically update Maven projects configuration" (due to a few limitations/issues of mwe2, it's better not to let m2e update our projects upon POM changes, since it modifies .classpath files in an unwanted way); this should be set in Oomph
- import the Git repository "Import from Git" (clone it from scratch so that it does not interfere with existing ones), git@github.com:eclipse/xtext-core.git, specifying the branch "lb_2052_maven_tycho", make sure "Search for nested projects" is selected and import all projects
- open /xtext-core/xtext-core.target and set it as the target platform
- once the target platform has been resolved and set, and the build completes, everything should be compiled without dirty Git files.
- Running tests from Eclipse should work

More details on the relevant changes I made and what's left to do will follow shortly.


xtext-core.target
```diff
@@ -0,0 +1,55 @@
...
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/modeling/emf/emf/builds/release/2.29"/>
+			<unit id="org.eclipse.emf.sdk.feature.group" version="0.0.0"/>
+		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/modeling/emft/mwe/updates/releases/2.14.0/"/>
+			<unit id="org.eclipse.emf.mwe2.language.sdk.feature.group" version="0.0.0"/>
+		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/releases/2022-12/"/>
```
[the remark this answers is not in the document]
LorenzoBettini: Yes, it's on the TODO list I was about to document.
I had to switch to 2022-12 because with 2022-03 I couldn't run plug-in tests from Eclipse (Eclipse wasn't starting at all, though the CPU was doing something; I haven't investigated further). Actually, there's no need to run plug-in tests in this Git repository, but just in case.

org.eclipse.xtext.util/pom.xml
```diff
@@ -15,7 +15,10 @@
 		<dependency>
 			<groupId>org.eclipse.xtend</groupId>
 			<artifactId>org.eclipse.xtend.lib</artifactId>
-			<!-- TODO: update it -->
+			<!--
+			TODO: update it
+			The dev BOM should be in the same Git repository of this one
+			-->
 			<version>${xtext-current-release-version}</version>
```
cdietrich, line 22: shouldnt this be project.version?

LorenzoBettini: I'm writing the notes for this PR (including this part), but since you asked: this is feasible (easily) only if the xtext-lib projects are in the same Git repository (which I'd like to do as well); otherwise, we could still do that now if we enable the snapshot repository pointing to the latest Jenkins build. (note that currently the Maven repository in the POM are disabled).

cdietrich: @LorenzoBettini we still can do it,
as we do here
https://github.com/eclipse/xtext-maven/blob/be49c0cf901f9110e57ff3342806b6ef0878d8bd/org.eclipse.xtext.maven.parent/pom.xml#L22
https://github.com/eclipse/xtext-maven/blob/be49c0cf901f9110e57ff3342806b6ef0878d8bd/org.eclipse.xtext.maven.parent/pom.xml#L285

or here
https://github.com/eclipse/xtext-core/blob/62b0ee64f3590befaabc1066fca185e0bec32dba/gradle/upstream-repositories.gradle#L30

LorenzoBettini:
These are the additional notes on the relevant changes:

NOTES:

- Must use in TP https://download.eclipse.org/modeling/tmf/xtext/updates/milestones/ because xtext.wizard requires xbase.lib 2.30.0 (is that intentional?)
- org.eclipse.xtext.testing: saved launch configuration for JUnit5 and removed the now useless org.eclipse.xtext.testing.tests.junit5.RunJUnit5Suite
- org.eclipse.xtext.tests: as above
- in POMs I left this comment concerning the dev-bom and xtend.lib
```xml
<!--
TODO: update it
The dev BOM should be in the same Git repository of this one
-->
```
when everything is normalized (one single repo?) we should use ${project.version} instead of a hardcoded version in the POMs.
The only point where we rely on the currently released version or milestone is for the xtend-maven-plugin
- Removed from target platform configuration in the POM `<pomDependencies>consider</pomDependencies>`, which is useless in our context
- currently there's a new parent POM for this project that corresponds to the `xtext-core` project
- the old `releng` subdirectory has been removed, since now there's the root parent POM (see above)
- a new project `org.eclipse.xtext.core.p2repository` has been introduced that replaces the old one in `releng` subdirectory, at the same level of the other projects; the `categoy.xml` has also cleaned removing explicit versions, which are useless
- for creating the POMs, I started from the ones available on Maven Central, which I seem to understand are now created by Gradle; I used them only for the Maven dependencies. In that respect, I removed these exclusions that were attached to all the dependencies:
```xml
      <scope>compile</scope>
      <exclusions>
        <exclusion>
          <artifactId>org.osgi.foundation</artifactId>
          <groupId>org.apache.felix</groupId>
        </exclusion>
        <exclusion>
          <artifactId>stringtemplate</artifactId>
          <groupId>org.antlr</groupId>
        </exclusion>
        <exclusion>
          <artifactId>cglib</artifactId>
          <groupId>*</groupId>
        </exclusion>
      </exclusions>
```
And I also removed the redundant default `<scope>compile</scope>`. The only place where I left the `stringtemplate` exclusion is when there's a dependency to `antlr-runtime` in `org.eclipse.xtext` POM:
```xml
	<dependency>
		<groupId>org.antlr</groupId>
		<artifactId>antlr-runtime</artifactId>
		<exclusions>
			<exclusion>
				<artifactId>stringtemplate</artifactId>
				<groupId>org.antlr</groupId>
			</exclusion>
		</exclusions>
	</dependency>
```
If I'm not missing something, the other exclusions are useless.
- Added in `xtext.testing` (to make it compile)
```
Import-Package:...
 org.junit.platform.runner;version="[1.0.0,2.0.0)"
 org.junit.platform.suite.api;version="[1.0.0,2.0.0)"
```
- To make JUnit 5 tests run from Eclipse (when there are also JUnit4 tests):
in `.classpath`
```xml
	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5"/>
```
in MANIFEST (imported packages)
```
 org.junit.jupiter.api;version="[5.1.0,6.0.0)",
 org.junit.jupiter.api.extension;version="[5.1.0,6.0.0)",
 org.junit.jupiter.api.function;version="[5.1.0,6.0.0)",
 org.junit.jupiter.api.io;version="[5.1.0,6.0.0)",
 org.junit.jupiter.params;version="[5.1.0,6.0.0)",
 org.junit.jupiter.params.provider;version="[5.1.0,6.0.0)",
```
In TP
```xml
<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
	<repository location="https://download.eclipse.org/tools/orbit/downloads/2023-03"/>
	<unit id="org.apache.log4j" version="1.2.24.v20221221-2012"/>
	<unit id="org.junit" version="4.13.2.v20211018-1956"/>
	<unit id="com.google.guava" version="30.1.0.v20221112-0806"/>
	<unit id="org.apiguardian" version="0.0.0"/>
	<unit id="org.junit.jupiter.api" version="0.0.0"/>
	<unit id="org.junit.jupiter.engine" version="0.0.0"/>
	<unit id="org.junit.jupiter.migrationsupport" version="0.0.0"/>
	<unit id="org.junit.jupiter.params" version="0.0.0"/>
	<unit id="org.junit.platform.commons" version="0.0.0"/>
	<unit id="org.junit.platform.engine" version="0.0.0"/>
	<unit id="org.junit.platform.launcher" version="0.0.0"/>
	<unit id="org.junit.platform.runner" version="0.0.0"/>
	<unit id="org.junit.platform.suite.api" version="0.0.0"/>
	<unit id="org.junit.platform.suite.commons" version="0.0.0"/>
	<unit id="org.junit.vintage.engine" version="0.0.0"/>
</location>
```
The last one is to avoid signing problems during the execution
- In `.xtext.wizards` added imported package `org.apache.log4j` and required bundle `org.kohsuke.args4j;bundle-version="2.33.0"` 
taken from Orbit. Add this dependency also in POM (which wasn't there, and that looked suspicious because in the gradle build file there's `api 'args4j:args4j'`).
- `xtext.xtext.bootstrap`: Missing MANIFEST.MF (FIXED), not a PDE project (FIXED), Added a POM. No configuration for running the workflow during the Maven build but I guess that's not needed: This can now be run from the Eclipse workspace (without egg-and-chicken problem, I guess). Currently, it requires `xtext-eclipse` to be cloned as well and present in a specific folder. (with repo merge this should not be a problem anymore). not meant to be deployed to Maven Central I guess, so there are no dependencies in the POM.
- `xtext.testlanguages` is not deployed on Maven Central, so there's no need to specify Maven deps in the POM (it's just a plugin project). Currently there's no Maven mechanism to generate the languages (but can be added in a profile), but since the generated sources must then be committed, I guess the generation is meant to be run from Eclipse.
- `.xtext.ide.tests`: Added `org.eclipse.emf.mwe2.lib` to required bundle otherwise `GenerateTestLanguage.mwe2` cannot find
`org.eclipse.emf.mwe2.ecore.EcoreGenerator`. Added `org.aopalliance;bundle-version="1.0.0"` to avoid ClassNotFoundError during the tests. Similarly, Added `com.google.gson;bundle-version="2.9.1"`. Removed `xtext.ide.tests (xtend) launch` because there are no more Xtend tests in this project.
- `.xtext.tests` Added `org.eclipse.emf.mwe2.lib;bundle-version="2.14.0"` for the same reason as above.
Added required bundle `org.aopalliance;bundle-version="1.0.0"` To avoid `java.lang.NoClassDefFoundError: [Lorg/aopalliance/intercept/MethodInterceptor;` and imported package `org.apache.commons.logging;version="1.2.0"` to avoid `java.lang.NoClassDefFoundError: org/apache/commons/logging/LogFactory`.
`/org.eclipse.xtext.tests/lib/simple.jar` is added explicity as a JAR to the launch configuration, otherwise
org.eclipse.xtext.resource.ClassloaderClasspathUriResolverTest.testClasspathUriForFileInJar() and
org.eclipse.xtext.resource.ClassloaderClasspathUriResolverTest.testClasspathUriForFileInJarWithFragment() fail with a NPE. From the Maven point of view, nothing else is required.
- Removed `org.eclipse.xtext.testing.tests.junit5.RunJUnit5Suite` and `org.eclipse.xtext.junit5.RunSuiteJUnit5` that were used for running JUnit5 tests with JUnit4 (they were also based on deprecated code); now tests are executed through JUnit5 that is capable of running also JUnit4.
- `.xtext.xtext.generator` exports also `org.eclipse.xtext.xtext.generator.index` otherwise in `.xtext.tests` `org.eclipse.xtext.xtext.generator.index.ResourceDescriptionStrategyFragment` has restriction errors.
- In general, renamed a few tests to adhere the Maven surefire convention (e.g., end with Test or Tests or TestCase)
- In this Git repository, all tests are executed as JUnit tests, so, from Maven, they must be executed through `maven-surefire-plugin`, NOT `tycho-surefire-plugin`. Tycho supports that, as long as the package type is `eclipse-plugin` (NOT `eclipse-test-plugin`). In the parent POM I configured in pluginManagement this:
```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <version>3.0.0-M8</version>
    <dependencies>
        <dependency>
            <groupId>org.apache.maven.surefire</groupId>
            <artifactId>surefire-junit-platform</artifactId>
            <version>3.0.0-M8</version>
        </dependency>
    </dependencies>
    <configuration>
        <!--
        The Surefire Plugin excludes nested classes by default: http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#excludes
        By removing all default exclusions the nested classes get executed too, JUnit 4 and JUnit 5.
        -->
        <excludes>
            <exclude />
        </excludes>
    </configuration>
</plugin>
```
this allows to run our tests as JUnit tests; it forces the JUnit5 runner so that we're sure we can run both JUnit4 and JUnit5

org.eclipse.xtext.util/pom.xml
LorenzoBettini: OK, I'll try to re-enable that.

LorenzoBettini: This https://github.com/eclipse/xtext-core/pull/2054/commits/65f209b8838c9f56d30e02a14701cf706a96f1a6 should implement it; for the moment, it's only useful for xtext-lib; xtend-maven-plugin is still taken from the latest release or milestone. Of course, it could be adapted as well.

cdietrich: for xtend maven plugin we eplicitely want to use a bootstraped older version

LorenzoBettini: OK, so that's fine as it is now, isn't it? Or do you prefer a released version (2.29.0)?

cdietrich: no we want to bootstrap against M1

LorenzoBettini:
> created lb_2052_maven_tycho branches in all downstream repositories to test the results of the build

I seem to understand that they are succeeding (though there were a few failures because the p2 repository was being replaced by a new Jenkins build). This one https://ci.eclipse.org/xtext/job/xtext-xtend/job/lb_2052_maven_tycho/ was interrupted; was that intentional?

LorenzoBettini:
Things that are still left to do, but could be done also later if the PR is accepted:

TODO:

- all the gradle files are still there, though all the gradle natures from projects have been removed. They should be harmless (though, as I said, I explicitly uninstalled buildship from my eclipse), but to keep things clean we should remove them.
- remove xtend-gen contents from the Git repository (they are processed during the build anyway): I guess that was needed to make buildship happy when creating configurations during the Oomph setup? This is not needed anymore since the Java sources will be automatically regenerated when projects are imported in Eclipse (and now the projects are proper Eclipse plug-in projects). The only thing that I guess is still needed is that the folders xtend-gen must be there when the projects are imported, so we should add a file .keepme or .gitkeep (and exclude that from the maven clean)
- sign Maven artifacts (for Maven central deployment) and p2 artifacts. I already know how to do that (typically in a separate profile), but maybe it's too early now?
- deploy Maven artifacts to Central after removing qualifiers; typically I prefer to publish p2 repositories and Maven artifacts separately: the released Maven artifacts typically don't have a release qualifier; so Maven artifacts are deployed separately after removing all the -SNAPSHOT,.qualifier. I already know how to do that with Maven and Tycho plugins. This should be executed before the deployment.
- Javadoc: Description expected after @deprecated in org.eclipse.xtext.impl.XtextFactoryImpl.getPackage() for Javadoc @deprecated; I disabled that check in the jdt preferences (also considered by Tycho) to avoid the error.
- `xtext.ide` has many more deps in the MANIFEST than in the POM (watching the one currently deployed to Maven central)! Same for `xtext.testing`. Compare with the build.gradle (some are optional, maybe they should be in the POM?).

cdietrich:
yeds i pushed all branches at the same time
they now should resolve as i started the root job with trigger downstream manually

cdietrich:
the xtend-gen files are checked in by intention, not to satisfy gradle

LorenzoBettini:
@cdietrich do you think I could start working on `xtext-lib` as well?

LorenzoBettini:
> the xtend-gen files are checked in by intention, not to satisfy gradle

For any specific reason? (just to understand)

cdietrich:
basically to see changes and diffs locally with new xtend versions

cdietrich:
xtend-lib and extras would be interesting too.
i assume we have to skip the gwt stuff in the former

cdietrich:
@LorenzoBettini yes the optional things should be in the pom. see this debt https://github.com/eclipse/xtext/issues/1847

org.eclipse.xtext/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+		<version>2.30.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>org.eclipse.xtext</artifactId>
+	<packaging>eclipse-plugin</packaging>
+
+	<name>Xtext Runtime Bundle</name>
+	<description>Core component of Xtext.</description>
+
+	<dependencies>
+		<dependency>
```
cdietrich, line 15: here optional stuff is missing too

cdietrich:
see also https://github.com/eclipse/xtext/issues/2135

org.eclipse.xtext.ide.tests/META-INF/MANIFEST.MF
```diff
@@ -17,6 +17,9 @@ Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.eclipse.lsp4j;bundle-version="[0.19.0,0.20.0)";resolution:=optional,
  org.eclipse.lsp4j.jsonrpc;bundle-version="[0.19.0,0.20.0)";resolution:=optional,
  org.eclipse.xtext.testlanguages,
- org.eclipse.xtext.testlanguages.ide
+ org.eclipse.xtext.testlanguages.ide,
+ org.eclipse.emf.mwe2.lib;bundle-version="2.14.0",
+ org.aopalliance;bundle-version="1.0.0",
+ com.google.gson;bundle-version="2.9.1"
```
cdietrich, line 23: isnt this a transitive dep?

LorenzoBettini: Unless they are re-exported in the MANIFEST by other projects I guess they're not (the project did not compile)

LorenzoBettini:
@cdietrich I handled optional dependencies in the last two commits for .xtext and .xtext.testing (the other ones are OK or they are not meant to be deployed to Maven central).

However, in `.xtext.testing` I ignored testImplementation (they should not be needed, if I understand correctly?)

org.eclipse.xtext.ide.tests/META-INF/MANIFEST.MF
cdietrich: might be that gradle behaves differently.
do you know what the error message was?
something along required by class file?

cdietrich:
@LorenzoBettini  i dont know if the tests there should be moved to a proper test bundle
currently we have runtime and test code there
and these are maven scope test dependencies

org.eclipse.xtext.ide.tests/META-INF/MANIFEST.MF
LorenzoBettini: The tests were failing at runtime with NoClassFoundError

cdietrich:
@LorenzoBettini i am talking about xtext.testing and not ide.tests

cdietrich:
ran locally with j17 and got

```
INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.117 s - in [INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   CliProjectsCreatorMainTest.testCreateGreetingsGradleProject:47 »  Unresolved compilation problems: 
	CmdLineParser cannot be resolved to a type
	CmdLineParser cannot be resolved to a type
	CmdLineException cannot be resolved to a type
	CmdLineException cannot be resolved to a type
	CmdLineException cannot be resolved to a type

```
is there something wrong with the args4j dependency?

cdietrich:
[org.eclipse.xtext.xtext.wizard.cli.CliProjectsCreatorMainTest.txt](https://github.com/eclipse/xtext-core/files/10437494/org.eclipse.xtext.xtext.wizard.cli.CliProjectsCreatorMainTest.txt)

cdietrich:
```
/home/dietrich/xtext-master/git/xtext-core/.m2/repository/args4j/args4j/2.33/args4j-2.33.jar  /home/dietrich/xtext-master/git/xtext-core/.m2/repository/p2/osgi/bundle/org.kohsuke.args4j/2.33.0.v20160323-2218/org.kohsuke.args4j-2.33.0.v20160323-2218.jar
```
@LorenzoBettini  is it possible stuff is twice on surefire path?
[a.log](https://github.com/eclipse/xtext-core/files/10437690/a.log)

LorenzoBettini:
### 

> ran locally with j17 and got
> 
> ```
> INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.117 s - in [INFO] Results:
> [INFO] 
> [ERROR] Errors: 
> [ERROR]   CliProjectsCreatorMainTest.testCreateGreetingsGradleProject:47 »  Unresolved compilation problems: 
> 	CmdLineParser cannot be resolved to a type
> 	CmdLineParser cannot be resolved to a type
> 	CmdLineException cannot be resolved to a type
> 	CmdLineException cannot be resolved to a type
> 	CmdLineException cannot be resolved to a type
> ```
> 
> is there something wrong with the args4j dependency?

I'm currently on a macOS where I only have JDK 17 and the build runs fine:

```
[INFO] Running org.eclipse.xtext.xtext.wizard.cli.CliProjectsCreatorMainTest
45461 [main] WARN  xtext.wizard.cli.CliProjectsCreator  - The combination of Gradle + flat project layout will no longer be supported in Gradle 8.0: https://docs.gradle.org/7.1.1/userguide/upgrading_version_7.html#deprecated_flat_project_structure
45465 [main] INFO  t.wizard.cli.CliProjectsCreatorMain  - Created projects for language org.xtext.example.mydsl.MyDsl in ...
```

LorenzoBettini:
> ```
> /home/dietrich/xtext-master/git/xtext-core/.m2/repository/args4j/args4j/2.33/args4j-2.33.jar  /home/dietrich/xtext-master/git/xtext-core/.m2/repository/p2/osgi/bundle/org.kohsuke.args4j/2.33.0.v20160323-2218/org.kohsuke.args4j-2.33.0.v20160323-2218.jar
> ```
> 
> @LorenzoBettini is it possible stuff is twice on surefire path? [a.log](https://github.com/eclipse/xtext-core/files/10437690/a.log)

What is the Maven repository in the same folder of the Git repository?
Are you running the build from Eclipse, specifying a different Maven local repository?
From the command line?

In any case, I have them both in my .m2 repository as well:

```
find . -name "*args4j*"
./repository/p2/osgi/bundle/org.kohsuke.args4j
./repository/p2/osgi/bundle/org.kohsuke.args4j/2.33.0.v20160323-2218/org.kohsuke.args4j-2.33.0.v20160323-2218-p2artifacts.xml
./repository/p2/osgi/bundle/org.kohsuke.args4j/2.33.0.v20160323-2218/org.kohsuke.args4j-2.33.0.v20160323-2218.jar
./repository/args4j
./repository/args4j/args4j
./repository/args4j/args4j/2.33/args4j-2.33.jar.sha1
./repository/args4j/args4j/2.33/args4j-2.33.pom
./repository/args4j/args4j/2.33/args4j-2.33.jar
./repository/args4j/args4j/2.33/args4j-2.33.pom.sha1
./repository/args4j/args4j-site
./repository/args4j/args4j-site/2.33/args4j-site-2.33.pom
./repository/args4j/args4j-site/2.33/args4j-site-2.33.pom.sha1
```

cdietrich:
will try on different machines tomorrow

cdietrich:
i am running maven from command line

LorenzoBettini:
Then, it's crucial to know the version of Maven (I'm using 3.8.6, which is also the one embedded in m2e); if you recall, we had a few problems due to a bug in Maven a few versions ago

cdietrich:
i tried 3.8.7 and 3.8.6 and

cdietrich:
i also notied that the jar files contain xtend files.

LorenzoBettini:
> i also notied that the jar files contain xtend files.

which ones? the .xtend files and trace files are excluded by the Tycho compiler plugin the configuration: they must be only in the source jars, not in the binary jars

cdietrich:
javap -v CliProjectsCreatorMain.class also shows the problem
```
  public static void main(java.lang.String[]);
    descriptor: ([Ljava/lang/String;)V
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
      stack=3, locals=1, args_size=1
         0: new           #42                 // class java/lang/Error
         3: dup
         4: ldc           #55                 // String Unresolved compilation problems: \n\tCmdLineParser cannot be resolved to a type\n\tCmdLineParser cannot be resolved to a type\n\tCmdLineException cannot be resolved to a type\n\tCmdLineException cannot be resolved to a type\n\tCmdLineException cannot be resolved to a type\n
         6: invokespecial #46                 // Method java/lang/Error."<init>":(Ljava/lang/String;)V

```
and with
javap -v ./org.eclipse.xtext.xtext.wizard/target/classes/org/eclipse/xtext/xtext/wizard/cli/CliProjectsCreatorMain.class
but why does then the maven-compiler-plugin not fail

cdietrich:
will also do a fresh clone, i also have multiple variants of the main class floating around.
also the timestamp does not seem to get updated.
am not sure which output folder is used or old class reused

cdietrich:
xtend files in jar
```
cd org/eclipse/xtext/xtext/wizard/
[dietrich@dietrich-x250 wizard]$ ls -la
insgesamt 748
drwxrwxr-x 4 dietrich dietrich  4096 Jan 16 17:53  .
drwxrwxr-x 3 dietrich dietrich  4096 Jan 16 17:53  ..
-rw-rw-r-- 1 dietrich dietrich  1300 Jan 16 17:53  AbstractFile.class
-rw-rw-r-- 1 dietrich dietrich   954 Jan 16 17:53  BinaryFile.class
-rw-rw-r-- 1 dietrich dietrich  1415 Jan 16 17:53  BuildSystem.class
drwxrwxr-x 2 dietrich dietrich  4096 Jan 16 17:53  cli
drwxrwxr-x 2 dietrich dietrich  4096 Jan 16 17:53  ecore2xtext
-rw-rw-r-- 1 dietrich dietrich  1416 Jan 16 17:53  Ecore2XtextConfiguration.class
-rw-rw-r-- 1 dietrich dietrich  2771 Jan 16 17:53  EPackageInfo.class
-rw-rw-r-- 1 dietrich dietrich  2158 Jan 16 17:53 'ExternalDependency$MavenCoordinates.class'
-rw-rw-r-- 1 dietrich dietrich  1542 Jan 16 17:53 'ExternalDependency$P2Coordinates.class'
-rw-rw-r-- 1 dietrich dietrich  2939 Jan 16 17:53  ExternalDependency.class
-rw-rw-r-- 1 dietrich dietrich  7444 Jan 17 18:15  GradleBuildFile.class
-rw-rw-r-- 1 dietrich dietrich  4149 Jan 16 17:41  .GradleBuildFile.java._trace
-rw-rw-r-- 1 dietrich dietrich  1668 Jul 20 20:05  GradleBuildFile.xtend
-rw-rw-r-- 1 dietrich dietrich  5021 Jan 16 17:41  .GradleBuildFile.xtendbin

```
will also test this one with a fresh workspace

cdietrich:
the wizardtest problem is gone on a fresh workspace, do is the xtend in jar problem
will try to find out which bad local class files is pulled in the old workspace.

LorenzoBettini:
I checked the produced binary JARS and I don't have .xtend files nor trace files. But if you run the build locally (after using Eclipse), you have to do a clean build, otherwise Maven/Tycho will take the .xtend and trace files that Eclipse copied to the output folder. (this behavior in Eclipse can be configured, but from Eclipse it's harmless and the build is meant to be executed in a clean environment).

Concerning where Tycho generates the output there's a small thing:
- it uses the target/classes for main code
- it uses the target/test-classes for code in test source folder
- this holds for `eclipse-plugin` projects, because tests will be executed by maven-surefire; for `eclipse-test-plugin` it generates everything in `target/classes`
This is OK in a project containing both main sources and test sources and should correspond to the specification in .classpath.
For projects containing only test source folders (e.g., .xtext.ide.tests) we should still configure .classpath to generate to target/classes because otherwise the packaging-plugin will complain that there's no target/classes; this means that during the build Tycho will still generate class files in target/test-classes. This might result in the same class files in both output folders (in target/classes generated during the standard compile phase, in target/test-classes generated during testCompile phase). In rare cases, e.g., when you delete a test Java class from Eclipse, JUnit will still find the old .class file in the other output folder.
I can tweak that if we want.

In general, however, the maven/tycho build, if executed from Eclipse, should better be always a clean build.

LorenzoBettini:
> xtend files in jar
> 
> ```
> cd org/eclipse/xtext/xtext/wizard/
> [dietrich@dietrich-x250 wizard]$ ls -la
> insgesamt 748
> drwxrwxr-x 4 dietrich dietrich  4096 Jan 16 17:53  .
> drwxrwxr-x 3 dietrich dietrich  4096 Jan 16 17:53  ..
> -rw-rw-r-- 1 dietrich dietrich  1300 Jan 16 17:53  AbstractFile.class
> -rw-rw-r-- 1 dietrich dietrich   954 Jan 16 17:53  BinaryFile.class
> -rw-rw-r-- 1 dietrich dietrich  1415 Jan 16 17:53  BuildSystem.class
> drwxrwxr-x 2 dietrich dietrich  4096 Jan 16 17:53  cli
> drwxrwxr-x 2 dietrich dietrich  4096 Jan 16 17:53  ecore2xtext
> -rw-rw-r-- 1 dietrich dietrich  1416 Jan 16 17:53  Ecore2XtextConfiguration.class
> -rw-rw-r-- 1 dietrich dietrich  2771 Jan 16 17:53  EPackageInfo.class
> -rw-rw-r-- 1 dietrich dietrich  2158 Jan 16 17:53 'ExternalDependency$MavenCoordinates.class'
> -rw-rw-r-- 1 dietrich dietrich  1542 Jan 16 17:53 'ExternalDependency$P2Coordinates.class'
> -rw-rw-r-- 1 dietrich dietrich  2939 Jan 16 17:53  ExternalDependency.class
> -rw-rw-r-- 1 dietrich dietrich  7444 Jan 17 18:15  GradleBuildFile.class
> -rw-rw-r-- 1 dietrich dietrich  4149 Jan 16 17:41  .GradleBuildFile.java._trace
> -rw-rw-r-- 1 dietrich dietrich  1668 Jul 20 20:05  GradleBuildFile.xtend
> -rw-rw-r-- 1 dietrich dietrich  5021 Jan 16 17:41  .GradleBuildFile.xtendbin
> ```
> 
> will also test this one with a fresh workspace

I can confirm you that they are not there if you do a clean build, as I said above.
(on a side note, on the CI, the build is always clean ;)

cdietrich:
```
find . -name "ProjectDescriptor.xtend"
./org.eclipse.xtext.xtext.wizard/target/classes/org/eclipse/xtext/xtext/wizard/ProjectDescriptor.xtend
./org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ProjectDescriptor.xtend
./org.eclipse.xtext.xtext.wizard/bin/main/org/eclipse/xtext/xtext/wizard/ProjectDescriptor.xtend
```

LorenzoBettini:
> ```
> find . -name "ProjectDescriptor.xtend"
> ./org.eclipse.xtext.xtext.wizard/target/classes/org/eclipse/xtext/xtext/wizard/ProjectDescriptor.xtend
> ./org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ProjectDescriptor.xtend
> ./org.eclipse.xtext.xtext.wizard/bin/main/org/eclipse/xtext/xtext/wizard/ProjectDescriptor.xtend
> ```

The `bin/main` tells me that you haven't cloned from scratch nor followed my instructions ;)

Note that neither Eclipse nor Maven can remove output folders they (now) know nothing about, left from previous configurations. Either you remember to do that manually the first time, or you start from scratch. (with git you can also run `git clean -fdx`).

cdietrich:
clean -fdx makes my oomph env ususable cause it will delete files needed by buildship/gradle :(

cdietrich:
=> once we roll out we need to communicate the projects need cleanup.
sry for the noise

LorenzoBettini:
> clean -fdx makes my oomph env ususable cause it will delete files needed by buildship/gradle :(

That's expected, that's why I wrote the instructions that currently are manual, until I update the Oomph file ;)

2-maven-build.sh
```diff
@@ -5,11 +5,12 @@ if [ -z "$JENKINS_URL" ]; then
 fi
 
 mvn \
-  -f releng \
+  deploy \
```
cdietrich, line 8: should we add clean goal here anyway?

LorenzoBettini: It would be useless because in the CI the environment is always clean (it wasn't there before either); of course, I guess it would be harmless (it would basically only clean xtend-gen folders). So if you want we can add it.

cdietrich: i use the 1 2 3 scripts also locally

cdietrich:
@LorenzoBettini did you try to generate the testlanguages and xtext via workflow from eclipse?
check
```
org.eclipse.xtext.ide.tests/build.gradle:task generateTestLanguages(type: XtextGeneratorTask) {
org.eclipse.xtext.testlanguages/build.gradle:task generateTestLanguages(type: XtextGeneratorTask) {
org.eclipse.xtext.testlanguages/build.gradle:compileJava.mustRunAfter generateTestLanguages
org.eclipse.xtext.tests/build.gradle:task generateTestLanguages(type: XtextGeneratorTask) {
```
to find the root workflows
we might have to adjust build.properties

LorenzoBettini:
> @LorenzoBettini did you try to generate the testlanguages and xtext via workflow from eclipse? check
> 
> ```
> org.eclipse.xtext.ide.tests/build.gradle:task generateTestLanguages(type: XtextGeneratorTask) {
> org.eclipse.xtext.testlanguages/build.gradle:task generateTestLanguages(type: XtextGeneratorTask) {
> org.eclipse.xtext.testlanguages/build.gradle:compileJava.mustRunAfter generateTestLanguages
> org.eclipse.xtext.tests/build.gradle:task generateTestLanguages(type: XtextGeneratorTask) {
> ```
> 
> to find the root workflows we might have to adjust build.properties

No, not yet.

The workflows are meant to be run from Eclipse only, aren't they?
Not from Maven, right?

cdietrich:
yes. the ones in xtext-eclipse we run manually too. see https://github.com/eclipse/xtext/blob/master/Builds.md

2-maven-build.sh
LorenzoBettini: I added 'clean' to the Maven script

org.eclipse.xtext/pom.xml
LorenzoBettini: So optional dependencies should now be handled correctly

LorenzoBettini:
> @LorenzoBettini i am talking about xtext.testing and not ide.tests

Probably you're right: I gave for granted that the contents of the tests source folder would not end in the JAR, while they do. I'll first ask Tycho if that's intended.

LorenzoBettini:
> > @LorenzoBettini i am talking about xtext.testing and not ide.tests
> 
> Probably you're right: I gave for granted that the contents of the tests source folder would not end in the JAR, while they do. I'll first ask Tycho if that's intended.

I opened https://github.com/eclipse-tycho/tycho/issues/2000

LorenzoBettini:
> > > @LorenzoBettini i am talking about xtext.testing and not ide.tests
> > 
> > 
> > Probably you're right: I gave for granted that the contents of the tests source folder would not end in the JAR, while they do. I'll first ask Tycho if that's intended.
> 
> I opened [eclipse-tycho/tycho#2000](https://github.com/eclipse-tycho/tycho/issues/2000)

The problem is fixed! It was my fault. We can leave the tests there.

cdietrich:
there is also something that generates xtext itself in bootstrap.
and i would expect 3 workflows for tests in xtext-core

LorenzoBettini:
> there is also something that generates xtext itself in bootstrap. and i would expect 3 workflows for tests in xtext-core

I have to try the bootstrap. As I said above, the other 3 workflows are working.

LorenzoBettini:
@cdietrich I also tried bootstrap Xtext and it works! I had to update the build.properties with 2 additional bundles and the saved launch configuration, which contained strange references and wrong arguments in my opinion. See the last commit above.

So concerning mwe2 launches, everything works (of course, from Eclipse)

LorenzoBettini:
Another change: to reduce to the minimal the target platform of xtext-core (which, hopefully, after the repo merge, will be self-contained concerning our bundles), I modify it like that:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<?pde version="3.8"?>
<target name="xtext-core" sequenceNumber="0">
	<locations>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://ci.eclipse.org/xtext/job/xtext-lib/job/lb_2052_maven_tycho/lastSuccessfulBuild/artifact/build/p2-repository/"/>
			<unit id="org.eclipse.xtext.xbase.lib.feature.group" version="0.0.0"/>
		</location>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/modeling/tmf/xtext/updates/milestones/"/>
			<unit id="org.eclipse.xtext.common.types" version="0.0.0"/>
			<unit id="org.eclipse.xtext.ecore" version="0.0.0"/>
			<unit id="org.eclipse.xtext.logging" version="0.0.0"/>
		</location>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/modeling/m2t/xpand/updates/releases/R201605260315/"/>
			<unit id="org.eclipse.xtend" version="0.0.0"/>
			<unit id="org.eclipse.xtend.typesystem.emf" version="0.0.0"/>
		</location>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/modeling/emf/emf/builds/release/2.29"/>
			<unit id="org.eclipse.emf.sdk.feature.group" version="0.0.0"/>
		</location>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/modeling/emft/mwe/updates/releases/2.14.0/"/>
			<unit id="org.eclipse.emf.mwe2.language.sdk.feature.group" version="0.0.0"/>
		</location>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/releases/2022-12/"/>
			<unit id="org.eclipse.equinox.executable.feature.group" version="0.0.0"/>
			<unit id="org.eclipse.sdk" version="0.0.0"/>
			<unit id="org.eclipse.platform.feature.group" version="0.0.0"/>
			<unit id="org.eclipse.jdt.feature.group" version="0.0.0"/>
			<unit id="org.eclipse.pde.feature.group" version="0.0.0"/>
		</location>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/lsp4j/updates/releases/0.19.0"/>
			<unit id="org.eclipse.lsp4j.sdk.feature.group" version="0.0.0"/>
		</location>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.itemis.com/updates/releases/2.1.1"/>
			<unit id="de.itemis.xtext.antlr.feature.feature.group" version="0.0.0"/>
		</location>
		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
			<repository location="https://download.eclipse.org/tools/orbit/downloads/2023-03"/>
			<unit id="org.apache.log4j" version="1.2.24.v20221221-2012"/>
			<unit id="org.kohsuke.args4j" version="0.0.0"/>
			<unit id="org.junit" version="4.13.2.v20211018-1956"/>
			<unit id="com.google.guava" version="30.1.0.v20221112-0806"/>
			<unit id="org.apiguardian" version="0.0.0"/>
			<unit id="org.junit.jupiter.api" version="0.0.0"/>
			<unit id="org.junit.jupiter.engine" version="0.0.0"/>
			<unit id="org.junit.jupiter.migrationsupport" version="0.0.0"/>
			<unit id="org.junit.jupiter.params" version="0.0.0"/>
			<unit id="org.junit.platform.commons" version="0.0.0"/>
			<unit id="org.junit.platform.engine" version="0.0.0"/>
			<unit id="org.junit.platform.launcher" version="0.0.0"/>
			<unit id="org.junit.platform.runner" version="0.0.0"/>
			<unit id="org.junit.platform.suite.api" version="0.0.0"/>
			<unit id="org.junit.platform.suite.commons" version="0.0.0"/>
			<unit id="org.junit.vintage.engine" version="0.0.0"/>
		</location>
	</locations>
</target>
```

So, xbase lib feature is taken from my Jenkins branch (xtext-lib), old Xtend stuff from the Xpand update site, other bundles of Xtext from the milestones update site.

LorenzoBettini:
@cdietrich @szarnekow since maybe I'm on the right track and in the right mood ;) can I go on with something else? `xtext-extras` might be the next?

cdietrich:
Yes Extras ist the next and last  big thing for the build part itself (leaving Web and deployment assiee)

LorenzoBettini:
I've also started to migrate the Oomph setup.
This is the process to have a workspace with Oomph with the currently ported versions of xtext-core and xtext-lib (xtext-umbrella is not imported, and must not be imported or it will break the projects through buildship). Of course, currently, there are a few manual steps:

- manually clone from scratch the two repositories switching to my branches (**don't use existing repositories**):
```
git clone git@github.com:eclipse/xtext-core.git
cd xtext-core/
git checkout -t origin/lb_2052_maven_tycho
cd ..
git clone git@github.com:eclipse/xtext-lib.git
cd xtext-lib/
git checkout -t origin/lb_2052_maven_tycho
```
- Start Oomph and select "Distro for Committers"
- Use this setup file (adding that to your user)
https://raw.githubusercontent.com/eclipse/xtext-eclipse/lb_2052_maven_tycho/releng/org.eclipse.xtext.contributor/Xtext.setup
- **Select ONLY** "Xtext Core" and "Xbase/Xtend Libraries"
- For the 2 git repositories clones, select "Located in the specified folder location"
And specify the absolute paths to the two above clones repositories (see the screenshot for an example)

Screenshot 2023-01-19 at 18 31 22

- Press finish, after Eclipse restarted, the projects of the two repositories should be imported and after Maven dependencies have been resolved (for the pure Maven projects) the workspace should contain no error and no dirty files.

LorenzoBettini:
I've updated the above instructions for "Xtext Extras" as well.

stbischof:
@laeubi could you please have a look on that PR in general for me? Its not bnd style, what i would prefere, but maybe you can give some helpful hints (annotation support or something)

laeubi:
I can try to take a look but at a first glance it seems just moving from gradle->maven?
If xtext desire PDE/m2e support `bnd-process` goal quite well if they like to manage dependencies in `pom.xml` and want automatic generation of manifests, annotations are also supported by PDE since last release by default with any extra configuration and Tycho can build mixed projects or even create full maven based update-sites.

LorenzoBettini:
@stbischof are you sure this is the PR you want @laeubi to take a look at?
I guess you're referring to the porting of `xtext-web` to a Maven build. This PR has to do with something else.
The projects involved in this PR have already been taken care of (of course they have to be reviewed).

stbischof:
Have a good setup on core is the important part for me. But if you thing this is all that could be done here its okay.

Web is not important for us

LorenzoBettini:
> Have a good setup on core is the important part for me. But if you thing this is all that could be done here its okay.
> 
> Web is not important for us

@stbischof sorry I have misunderstood then. If you want you can try the new currently temporary Oomph setup based completely on Maven/Tycho (remember to import only core and lib)

cdietrich:
with repo merge this is obsolete

PR #2069, opened 2023-02-17
Prevent NPE

rubenporras:


rubenporras:
for #2070

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/util/NodeModelUtils.java
```diff
@@ -70,6 +70,9 @@ public class NodeModelUtils extends InternalNodeModelUtils {
 	public static ILeafNode findLeafNodeAtOffset(/* @NonNull */ INode node, int leafNodeOffset) {
 		INode localNode = node;
 		while(!(localNode instanceof AbstractNode)) {
+			if (localNode == null) {
```
szarnekow, line 73: The parameter is not supposed to be null. I’m skeptical about this change since it might cloak the real problem. Please show steps to reproduce.

rubenporras: Yes, that is right, but `localNode.getParent()` could return null, couldn't it?

For node to be `null`, in `EObjectAtOffsetHelper.getElementWithNameAt(XtextResource, int)` the call to `parseResult.getRootNode()` would have need to return null, but that looks less likely, so my guess is that 
localNode.getParent() returned null.

Unfortunately I saw the stacktrace after a long session of using the server, I do not know exactly what I did to trigger it.

rubenporras: This second version will still NPE if node is `null`, but not if `getParent` returns null. I think it does not hurt, but if you prefer not doing the change, I would understand it, just let me know and I will close the PR.

szarnekow: Do you have a custom implementation of INode or do you use the SyntheticLinkingLeafNode?

rubenporras: Yes, we have a custom implementation. I have not seen this problem for a while, and since I do not have time to actively work on it, I will close the PR for the moment.

